PR #108, opened 2022-05-03
Reafctoring react modules

spaceo:
#### What does this PR do?
- Module names and namespaces change to use dpl instead of ddb
- Composer changed to pull library from dpl-react repository
- All react drupal libraries removed and replaced with hello world library

#### Should this be tested by the reviewer and how?
* Try to build site by running: `task dev:reset`
* Open up the site in the browser at following url: "/react-dpl-demo"
* The hello-world app should be visible

#### Any specific requests for how the PR should be reviewed?
Look at commits

#### What are the relevant tickets?
https://reload.atlassian.net/browse/DDFSOEG-102


kasperg (approved):
:+1: but before this is merged we have to deal with the failing Behat test.

See my suggested solution below.

PR #111, opened 2022-05-05
Wait for page to be loaded before checking Hello World component

kasperg:
On GitHub Actions the text from the Hello World component is not
available when checking for it. [This causes the Behat test to fail](https://github.com/danskernesdigitalebibliotek/dpl-cms/runs/6306798524?check_suite_focus=true).
The logs show that the text is not available in the DOM but the
screenshot actually contain the text. The tests also pass locally.

The error may be caused by timing issues so try to wait a bit to
ensure that the page has loaded before we test.

### DIFF
--- a/tests/behat/context/ReactContext.php
+++ b/tests/behat/context/ReactContext.php
@@ -43,4 +43,14 @@ class ReactContext implements MinkAwareContext {
     }
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


spaceo:
LGTM 👍

spaceo (approved):

PR #112, opened 2022-05-12
Feature/streamline setup

cableman:
#### Link to issue

No current issue.

#### Description

* Ensure that go-task start/stop do the what people expect when working with docker compose.
* Add the PR template used by DDB/DDF CMS

#### Screenshot of the result

No screenshots

#### Checklist

No comments


spaceo (approved):
Look good.
Makes sense to run down instead of stop to get an actual clean slate :)
I could not check the last option in the description, because of failing GH action jobs.
The jobs are related to Lagoon integration and has nothing to do with the changes here so therefor I approve the PR.

PR #113, opened 2022-06-01
Add task for downloading and overwriting dpl-react

spaceo:
Simple way of fetching a specific release of dpl-react.

#### Link to issue

Doing as a part of this task:
https://reload.atlassian.net/browse/DDFSOEG-98

#### Description
As it is now we have quite rigid way of working with builds of dpl-react in dpl-cms.
We rely on builds that only are available from the latest main branch on dpl-react.
With this change: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/23 we will get a release of the HEAD of every branch on a push. That enables us to be able to download recent dev version of whatever branch we want.
The changes in this PR adds a task that downloads a given release and overwrites the dpl-react library with it.

#### Checklist


Taskfile.yml
```diff
@@ -132,6 +132,30 @@ tasks:
...
 
+  dev:dpl-react:get-asset-url:
+    summary: Outputs the url to the build in Github
+    cmds:
+      - ./dev-scripts/get-dpl-react-assets.py {{ .BRANCH }}
+    preconditions:
+    - sh: "[ ! -z {{.BRANCH}} ]"
+      msg: "Env variable BRANCH is not set or empty."
+
+  dev:dpl-react:download:
```
kasperg, line 143:
```suggestion
  dev:dpl-react:overwrite:
```

This stresses the purpose of the task is to overwrite an existing version of the PR - not just download it.

dev-scripts/get-dpl-react-assets.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
```
kasperg, line 1: I am not really happy about including a Python script here. It adds another language dependency that we have to be aware of. 

It creates a dependency for the development environment as we need to ensure that Python3 is available where the task is executed (now: locally or potentially using Docker). It may also create an additional overhead for maintenance.

It is quite a simple script. I suggest that we implement it in PHP or Bash.

A third option might be to create a dependency on the `gh` CLI tool which [supports downloading of assets](https://cli.github.com/manual/gh_release_download).

Taskfile.yml
```diff
@@ -132,6 +132,30 @@ tasks:
...
+    desc: Downloads assets from Github and overwrites existing dpl-react library
+    cmds:
+      - curl {{ .ASSET_URL }} -L -o dist.zip
+      - rm -rf web/libraries/dpl-react
+      - unzip dist.zip -d web/libraries/dpl-react
+      - echo "{{ .BRANCH }}" > web/libraries/dpl-react/RELEASE.txt
+      - echo "[Success] The dpl-react library was overwritten with the one fetched from {{ .BRANCH }}."
+      - defer: rm dist.zip
```
kasperg, line 151: :+1: for remembering to clean up. Also I did not know about defer.

Taskfile.yml
```diff
@@ -132,6 +132,30 @@ tasks:
...
+      msg: "Env variable BRANCH is not set or empty."
+
+  dev:dpl-react:overwrite:
+    desc: Downloads assets from Github and overwrites existing dpl-react library
+    cmds:
+      - curl {{ .ASSET_URL }} -L -o dist.zip
+      - rm -rf web/libraries/dpl-react
+      - unzip dist.zip -d web/libraries/dpl-react
+      - echo "{{ .BRANCH }}" > web/libraries/dpl-react/RELEASE.txt
+      - echo "[Success] The dpl-react library was overwritten with the one fetched from {{ .BRANCH }}."
```
kasperg, line 150: Do we want/need to clear the cache afterwards?

kasperg (changes requested):
I have issues with the use of Python to get the download information.

Perhaps also consider updating the documentation explaining the intended workflow for developers. Perhaps in `documentation/local-development.md`.

dev-scripts/get-dpl-react-assets.py
spaceo: My thought was that the python script is run on the host - so no docker worries.
And another thought was that everyones OS'es runs python.
I knew I would get comments on this. But honestly I don't see the problem about using python.

Taskfile.yml
spaceo: Probably a good idea. I'll add that.

Taskfile.yml
```diff
@@ -132,6 +132,30 @@ tasks:
...
+    - sh: "[ ! -z {{.BRANCH}} ]"
+      msg: "Env variable BRANCH is not set or empty."
+
+  dev:dpl-react:download:
+    desc: Downloads assets from Github and overwrites existing dpl-react library
+    cmds:
+      - curl {{ .ASSET_URL }} -L -o dist.zip
+      - rm -rf web/libraries/dpl-react
+      - unzip dist.zip -d web/libraries/dpl-react
+      - echo "{{ .BRANCH }}" > web/libraries/dpl-react/RELEASE.txt
```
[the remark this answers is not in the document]
spaceo: Found the problem. It has been fixed in the gh action building the dpl-react dist.zip file

dev-scripts/get-dpl-react-assets.py
spaceo: Done

cableman (approved):
Optional: the console command could also download an unpack the zip file.

cableman (approved):

PR #118, opened 2022-07-15
Feature/integrate material app

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-176

#### Description
This PR adds the material app to dpl-cms.

What is done:
* Refactored dpl_search to dpl_react_apps, because then we do not need to have multiple modules. If we need a dpl_search module in the future we will just create one. But this module could be resposible for rendering most of the apps.
* Added material page rendering with pid route argument

#### Screenshot of the result
image

#### Checklist


composer.json
```diff
@@ -17,10 +17,10 @@
             "type": "package",
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-react",
-                "version": "0.2.1",
+                "version": "0.2.0-rc.5",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/release-feature%2Fsearch-result-app/dist.zip",
+                    "url": "https://github.com/reload/dpl-react/releases/download/release-feature%2Fautocomplete-accessibility/dist.zip",
```
kasperg, line 23: Please update this to point to the official builds.

We should also drop ad16c28034b77a78d1038ede96d1d60aee24dc19 as the related fixes are available in builds of `release/3` in the official repo.

tests/behat/features/30-react-components.feature
```diff
@@ -3,17 +3,10 @@ Feature: React components
 
   @api
   Scenario: Tokens are available
-    Given I enable modules "dpl_react, dpl_react_demo"
```
kasperg, line 6: :+1:

web/modules/custom/dpl_react_apps/dpl_react_apps.routing.yml
```diff
@@ -0,0 +1,14 @@
...
+  requirements:
+    _permission: "access content"
+
+dpl_react_apps.material_page:
+  path: "/material/{pid}"
+  defaults:
+    _controller: '\Drupal\dpl_react_apps\Controller\DplReactAppsController::material'
+  requirements:
+    _permission: "access content"
+    pid: "[0-9]+-[a-z]+:[0-9]+"
```
kasperg, line 14: Consider updating this pattern. 

I understand the intention but this may be too strict. Example: [On beta.bibliotek.dk we see work ids starting with `work-of`](https://beta.bibliotek.dk/materiale/harry-potter-og-de-vises-sten_joanne-k-rowling/work-of%3A870970-basis%3A22629344). I expect that we will need to support such pids ourselves at well. They will not work with the current pattern.

config/sync/core.extension.yml
```diff
@@ -7,7 +7,7 @@ module:
   dpl_login: 0
   dpl_mapp: 0
   dpl_react: 0
-  dpl_search: 0
+  dpl_react_apps: 0
```
kasperg, line 10: Consider whether this is the right approach.

This will effectively put all React applications into a single module, `dpl_react_apps` but what will then happen with any Drupal code related to searching e.g. configuration forms, server side code etc. Does this go in another `dpl_search` module or in `dpl_react_apps`?

As I see it we already have one supporting "tech" module - `dpl_react`. I suggest that we from here branch into domain related modules e.g. `dpl_search` which contains all code related to handling this domain.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -0,0 +1,56 @@
...
+    $c = ['context' => 'Material Page'];
+
+    return [
+      'material' => dpl_react_render('material', [
+        'pid' => $pid,
+        'material-header-author-by-text' => $this->t('By', [], $c),
+        'periodikum-select-year-text' => $this->t('Year', [], $c),
+        'periodikum-select-week-text' => $this->t('Week', [], $c),
+        'reserve-book-text' => $this->t('Reserve Book', [], $c),
+        'fine-on-bookshelf-text' => $this->t('Find at Book Shelf', [], $c),
```
kasperg, line 48:
```suggestion
        'reserve-book-text' => $this->t('Reserve book', [], $c),
        'fine-on-bookshelf-text' => $this->t('Find at book Shelf', [], $c),
```

Minor capitalization changes.

web/modules/custom/dpl_react_apps/dpl_react_apps.routing.yml
spaceo: We had a longer talk about this. :) For now we stick to this version and adress future problems if they should arise. There is pro's and con's with every solution. And no seemingly "right" choice.

config/sync/core.extension.yml
spaceo: I leave it like this for now.
Would make sense to continue this discussion with ITK

spaceo:
@cableman You can review this now...

cableman (approved):

PR #120, opened 2022-08-18
Search and material app updates

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-176

#### Description

This PR updates the configuration of the search and material apps inserted into the CMS. This is primarily in the way of props being passed to the apps in regard to UI texts and routes.

The route to works/materials is also renamed to `/work/:workid` because what we show is actually works - not materials.

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_react_apps/dpl_react_apps.module
+++ b/web/modules/custom/dpl_react_apps/dpl_react_apps.module
@@ -19,11 +19,18 @@ use Drupal\Core\Url;
  */
 function dpl_react_apps_preprocess_page(array &$variables): void {
   $search_result_url = Url::fromRoute('dpl_react_apps.search_result')->toString();
+
+  $options = ['context' => 'Search Header'];
+
   $data = [
     'search-header' => [
       'search-header-url' => $search_result_url,
-      'alt-text' => t('Search field', [], ['context' => 'Search Header']),
-      'input-placeholder' => t('Start typing in order to search', [], ['context' => 'Search Header']),
+      'alt-text' => t('Search field', [], $options),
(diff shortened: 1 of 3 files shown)


web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -18,12 +18,18 @@ use Drupal\Core\Url;
...
+  $options = ['context' => 'Search Header'];
+
   $data = [
     'search-header' => [
-      'search-header-url' => $search_result_url,
-      'alt-text' => t('Search field', [], ['context' => 'Search Header']),
-      'input-placeholder' => t('Start typing in order to search', [], ['context' => 'Search Header']),
+      'search-url' => DplReactAppsController::searchResultUrl(),
+      'material-url' => DplReactAppsController::materialUrl(),
+      'alt-text' => t('Search field', [], $options),
```
spaceo, line 27: Should be `$this-t. You can either change it or it will be fixed when merged with this pr: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/122

web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -18,12 +18,18 @@ use Drupal\Core\Url;
...
+      'input-placeholder-text' => t('Start typing in order to search', [],
+        $options),
+      'string-suggestion-author-text' => t('Author', [], $options),
```
spaceo, line 30: We need context in the t functions.

spaceo (changes requested):

spaceo (approved):
LGTM

cableman (approved):

PR #121, opened 2022-08-18
Trying to clean up codeing style in login module

cableman:
#### Link to issue

No issue.

#### Description

Trying to clean up the code style after switch til php 8 (sid effect of reading the code).

#### Screenshot of the result

No screenshot.

#### Checklist

The failed code style errors are from Drupal arrays that PHPStan whats us to define (due the check level is set to max). Not sure this will be possible: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type

### DIFF
--- a/web/modules/custom/dpl_library_token/src/LibraryToken.php
+++ b/web/modules/custom/dpl_library_token/src/LibraryToken.php
@@ -3,6 +3,7 @@
 namespace Drupal\dpl_library_token;
 
 use Drupal\dpl_library_token\Exception\LibraryTokenResponseException;
+use Safe\Exceptions\JsonException;
 use function Safe\json_decode as json_decode;
 
 /**
@@ -32,14 +33,16 @@ class LibraryToken {
    *
    * @return LibraryToken
    *   Token object created based on a json formed response.
+   *
+   * @throws \Drupal\dpl_library_token\Exception\LibraryTokenResponseException
    */
   public static function createFromResponseBody(string $response_body): self {
     // Get token data by decoding json.
     try {
       $token_data = json_decode($response_body, TRUE);
     }
-    catch (\JsonException $e) {
-      throw new LibraryTokenResponseException('Could not decode library token response', $e);
+    catch (JsonException $e) {
+      throw new LibraryTokenResponseException($e->getMessage(), $e->getCode(), $e);
     }
     if (empty($token_data['access_token'])) {
       throw new LibraryTokenResponseException('Access token is missing');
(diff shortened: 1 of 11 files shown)


web/modules/custom/dpl_library_token/src/LibraryToken.php
```diff
@@ -32,14 +33,16 @@ class LibraryToken {
    *
    * @return LibraryToken
    *   Token object created based on a json formed response.
+   *
+   * @throws \Drupal\dpl_library_token\Exception\LibraryTokenResponseException
```
kasperg, line 37: I do not think it is a good idea to add `@throws` annotations everywhere in the project as it stands now

1. We do not have any tooling to support/remind developers to add it in our CI pipeline.
2.  If we add `@throws` we have no way to ensure that it matches the current code. Subsequent can change what exceptions are thrown in a function.
3. The '@throws` section can quickly become littered with exceptions that the developer can do little about except letting them bubble up to the caller. Using [`thecodingmachine/safe`](https://github.com/thecodingmachine/safe) makes this situation worse

It seems like [PHPStan can help us with `@throws` and Exceptions](https://phpstan.org/blog/bring-your-exceptions-under-control) though.

Either I suggest we drop `@throws` entirely ([perhaps even forbid them](https://github.com/slevomat/coding-standard/#slevomatcodingstandardcommentingforbiddenannotations-)) or that we check up on the options provided by PHPStan and form an option on how we want to work with them including checked and unchecked exceptions.

web/modules/custom/dpl_library_token/src/LibraryTokenHandler.php
```diff
@@ -57,34 +53,42 @@ class LibraryTokenHandler {
...
    *   The library token logger channel.
+   *
+   * @throws \Drupal\dpl_library_token\Exception\MissingConfigurationException
+   * @throws \Safe\Exceptions\StringsException
    */
   public function __construct(
     KeyValueExpirableFactoryInterface $keyValueFactory,
     ConfigFactoryInterface $configFactory,
-    ClientInterface $http_client,
+    protected ClientInterface $httpClient,
```
kasperg, line 67: Constructor property promotion is not something we use elsewhere. Personally I am not sure how I feel about it outside of DTOs. In the current form it somewhat bypasses requirements regarding documentation of class properties mentioned above. Perhaps that is in fact a good thing.

If we want it we should consider [requiring it with our coding standard](https://github.com/slevomat/coding-standard#slevomatcodingstandardclassesrequireconstructorpropertypromotion-). If not we should consider preventing it using the same tools.

web/modules/custom/dpl_login/src/Controller/DplLoginController.php
```diff
@@ -67,7 +62,7 @@ public function __construct(
    *
    * @return static
    */
-  public static function create(ContainerInterface $container) {
+  public static function create(ContainerInterface $container): static {
```
kasperg, line 65: Here we should be able to drop the `@return static` in the doc block.

We should consider [enforcing return type hints as close to the language level as possible in our coding standards](https://github.com/slevomat/coding-standard#slevomatcodingstandardtypehintsreturntypehint-).

kasperg (changes requested):
There are a lot of different types of changes here. Some absolutely necessary because they fix logic errors and some which I think requires further consideration.

In the current form I do not think it can be merged.

PR #134, opened 2022-09-19
Add dpl library agency module with sms notification setting

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-209

#### Description

This PR introduces a new module for handling configuration belonging to a library agency.
The first part of the configuration is the possibility for an administartor to control wether it is possible for patrons to get notified by sms about current reservations.

#### Screenshot of the result

image

---

image

#### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
                 "version": "0.3.0-rc6",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/reload/dpl-react/releases/download/release-indloggetudlogget-funktionalitet-DDFSOEG-178/dist.zip",
+                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/release-release%2F4/dist.zip",
                     "type": "zip"
                 },
                 "require": {
@@ -35,7 +35,7 @@
                 "type": "drupal-library",
                 "version": "v1.7.0-rc1",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/release-release%2F3/dist.zip",
+                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/release-release%2F4/dist.zip",
                     "type": "zip"
                 }
             }
@@ -49,7 +49,7 @@
         "amazeeio/drupal_integrations": "0.3.5",
         "composer/installers": "1.11.0",
         "cweagans/composer-patches": "1.7.0",
-        "danskernesdigitalebibliotek/dpl-design-system": "^1.7@rc",
+        "danskernesdigitalebibliotek/dpl-design-system": "^1.7@RC",
         "danskernesdigitalebibliotek/dpl-react": "^0.3.0@RC",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.1",
         "drupal/config_ignore": "^2.3",
--- a/config/sync/config_ignore.settings.yml
+++ b/config/sync/config_ignore.settings.yml
@@ -8,6 +8,7 @@ ignored_config_entities:
   - '~core.entity_view_mode.node.*'
   - ~core.menu.static_menu_link_overrides
   - ~dblog.settings
+  - '~dpl_library_agency.general_settings:reservation_sms_notifications_disabled'
   - '~dpl_mapp.settings:domain'
   - ~field.settings
   - ~field.storage.node.body
--- a/config/sync/dpl_library_agency.general_settings.yml
+++ b/config/sync/dpl_library_agency.general_settings.yml
@@ -1,3 +1,1 @@
-library_agency_name: null
-sms_notifications_for_patrons_enabled: 1
 reservation_sms_notifications_disabled: 0
--- a/config/sync/dpl_library_agency.general_settings.yml
+++ b/config/sync/dpl_library_agency.general_settings.yml
@@ -1,0 +1,3 @@
+library_agency_name: null
+sms_notifications_for_patrons_enabled: 1
+reservation_sms_notifications_disabled: 0
--- a/web/modules/custom/dpl_library_agency/dpl_library_agency.module
+++ b/web/modules/custom/dpl_library_agency/dpl_library_agency.module
@@ -1,0 +1,39 @@
+<?php
+
+/**
+ * @file
+ * Contains dpl_library_agency.module.
+ */
+
+use Drupal\Core\Routing\RouteMatchInterface;
+
+/**
+ * Implements hook_help().
+ */
+function dpl_library_agency_help(string $route_name, RouteMatchInterface $route_match): string {
+  switch ($route_name) {
+    // Help for the dpl_library_agency module.
+    case 'help.page.dpl_library_agency':
+      $output = '';
+      $output .= '<h3>' . t('About') . '</h3>';
+      $output .= '<p>' . t('Module for handling library agency configuration and functionality') . '</p>';
+      return $output;
+
+    default:
+      return '';
+  }
+}
+
+/**
+ * Checks wether sms notifications for reservations are enabled.
+ *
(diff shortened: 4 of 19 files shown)


config/sync/dpl_library_agency.general_settings.yml
```diff
@@ -0,0 +1,3 @@
+library_agency_name: null
+sms_notifications_for_patrons_enabled: 1
+reservation_sms_notifications_disabled: 0
```
kasperg, line 3: Please reexport the configuration or remove the `sms_notifications_for_patrons_enabled` and `library_agency_name` entries.

As far as I can tell the only proper entry is `reservation_sms_notifications_disabled`.

web/modules/custom/dpl_library_agency/dpl_library_agency.module
```diff
@@ -0,0 +1,39 @@
...
+
+use Drupal\Core\Routing\RouteMatchInterface;
+
+/**
+ * Implements hook_help().
+ */
+function dpl_library_agency_help(string $route_name, RouteMatchInterface $route_match): string {
+  switch ($route_name) {
+    // Help for the dpl_library_agency module.
+    case 'help.page.dpl_library_agency':
```
kasperg, line 16: I wonder if you put any consideration into adding a `hook_help` implementation?

It is not something that we have added for other modules in DPL CMS, it is not something that we use in the Ding2 project, and the current contents does not provide much value compared to what is offered elsewhere e.g. in the menu descriptions. On the other hand I guess translations could help out with that.

web/modules/custom/dpl_library_agency/dpl_library_agency.routing.yml
```diff
@@ -0,0 +1,17 @@
...
+    _controller: '\Drupal\system\Controller\SystemController::systemAdminMenuBlockPage'
+    _title: 'Library Agency'
+  requirements:
+    _permission: 'administer library agency configuration'
+
+dpl_library_agency.general_settings:
+  path: '/admin/config/dpl-library-agency/general-settings'
+  defaults:
+    _form: '\Drupal\dpl_library_agency\Form\GeneralSettingsForm'
+    _title: 'Dpl Library Agency General Settings'
```
kasperg, line 13:
```suggestion
    _title: 'General Settings'
```

I do not think we need to mention the module name in site administrator facing texts.

web/modules/custom/dpl_library_agency/dpl_library_agency.permissions.yml
```diff
@@ -0,0 +1,3 @@
+administer library agency configuration:
+  title: 'Administer library agency configuration'
+  description: 'Read and view access for the library agency configuration.'
```
kasperg, line 3: Please rewrite this description. "Read and view access" does not seem very telling for the scope of the module.

web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
```diff
@@ -0,0 +1,64 @@
...
+
+    return parent::buildForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state): void {
+    $this->config('dpl_library_agency.general_settings')
+      ->set('library_agency_name', $form_state->getValue('library_agency_name'))
```
kasperg, line 57: Please remove `library_agency_name`.

There does not seem to be any control for it.

config/sync/dpl_library_agency.general_settings.yml
kasperg: Please remember to add all this configuration to our `config_ignore` configuration so local administrators can manage it freely.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -55,6 +55,8 @@ public function work(string $wid): array {
     return [
       'material' => dpl_react_render('material', [
         'wid' => $wid,
+        'sms-notifications-for-reservations-enabled' =>
+        (int) dpl_library_agency_reservation_sms_notifications_is_enabled(),
```
kasperg, line 59: Consider refactoring library agency configuration into a service.

A static function which accesses state/configuration makes it hard to setup unit tests of the controller. An injected service would make this easier.

Another simple approach would be to just let the `DplReactAppsController` access the configuration directly through the `config()` method but that would make the dependency between the two modules soft. I appreciate that it is an actual method.

web/modules/custom/dpl_react_apps/dpl_react_apps.info.yml
```diff
@@ -3,6 +3,7 @@ description: For integration of dpl_react apps on the site.
 package: DPL
 dependencies:
   - dpl_react:dpl_react
+  - dpl_library_agency:dpl_library_agency
```
kasperg, line 6: :+1: for getting this in

composer.json
```diff
@@ -20,7 +20,7 @@
                 "version": "0.3.0-rc6",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/reload/dpl-react/releases/download/release-indloggetudlogget-funktionalitet-DDFSOEG-178/dist.zip",
+                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/release-release%2F4/dist.zip",
```
kasperg, line 23: :+1: It would be nice if we could have some kind of warning to prevent merges of local builds to main branches.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
kasperg: Also I would appreciate a comment regarding why we need to convert a proper type for a value, a boolean, to an integer.

kasperg (changes requested):
:+1:

I has been a while since I have worked with actual Drupal code. I have quite a few thoughts. The primary blockers here have to do with configuration handling - illegal values and remembering to ignore the configuration.

config/sync/dpl_library_agency.general_settings.yml
spaceo: Yes, had in mind to get that part before I finished the PR

web/modules/custom/dpl_library_agency/dpl_library_agency.module
spaceo: Ok, honestly this and the weird name configuration key mentioned elsewhere are products of the co-pilot extensions. Great examples of that usage of co-pilot can escalate and create unneeded code/functionality.

web/modules/custom/dpl_library_agency/dpl_library_agency.routing.yml
spaceo: 👍

web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
spaceo: Yes 👍

composer.json
spaceo: yes 🎄

config/sync/config_ignore.settings.yml
```diff
@@ -8,6 +8,7 @@ ignored_config_entities:
   - '~core.entity_view_mode.node.*'
   - ~core.menu.static_menu_link_overrides
   - ~dblog.settings
+  - '~dpl_library_agency.general_settings:reservation_sms_notifications_disabled'
```
kasperg, line 11: Consider whether we want to ignore all settings under `dpl_library_agency.general_settings`.

If they are ment to be configured by the individual library we probably do not want to control them from the configuration management system.

kasperg (approved):
:+1: LGTM

PR #135, opened 2022-09-20
Fix CI builds

achton:
#### Description

This change does two things:

1. Fixes a bug with some CI-related composer packages
2. Fixes build errors when running CI tests

#### Checklist


kasperg (approved):
:+1: LGTM.

Excellent level of commit commenting here.

PR #136, opened 2022-09-23
Url proxy functionality

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-235

#### Description

This PR adds a new module called: dpl_url_proxy.
The module does following:
* Exposes a configuration form only adminesterable by admin/local_admin
* Provides an endpoint to be used by clients (in current project: react apps) to translate material urls.
* Sets an url prop on the material react app for the url proxy endpoint.
 
#### Screenshot of the result
image

image

image

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
+++ b/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
@@ -94,6 +94,7 @@ class DplReactAppsController extends ControllerBase {
         'search-url' => self::searchResultUrl(),
         'material-url' => self::materialUrl(),
         'auth-url' => self::authUrl(),
+        'url-proxy-url' => self::urlProxyUrl(),
         'material-header-author-by-text' => $this->t('By', [], $c),
         'periodikum-select-year-text' => $this->t('Year', [], $c),
         'periodikum-select-week-text' => $this->t('Week', [], $c),
@@ -159,6 +160,15 @@ class DplReactAppsController extends ControllerBase {
   public static function authUrl(): string {
     return self::ensureUrlIsString(
       Url::fromRoute('dpl_login.login')->toString()
+    );
+  }
+
+  /**
+   * Builds an url for the react apps to for proxy urls.
+   */
+  public static function urlProxyUrl(): string {
+    return self::ensureUrlIsString(
+      Url::fromRoute('dpl_url_proxy.generate_url')->toString()
     );
   }
 
--- a/web/modules/custom/dpl_url_proxy/dpl_url_proxy.info.yml
+++ b/web/modules/custom/dpl_url_proxy/dpl_url_proxy.info.yml
@@ -1,0 +1,6 @@
+name: 'dpl_url_proxy'
+type: module
+description: 'Module for generating specially formatted external urls'
+core: 8.x
+package: 'DPL'
+core_version_requirement: ^8 | ^9
--- a/web/modules/custom/dpl_url_proxy/dpl_url_proxy.info.yml
+++ b/web/modules/custom/dpl_url_proxy/dpl_url_proxy.info.yml
@@ -1,4 +1,4 @@
-name: 'dpl_url_proxy'
+name: 'Dpl Url Proxy'
 type: module
 description: 'Module for generating specially formatted external urls'
 core: 8.x
--- a/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
+++ b/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
@@ -1,0 +1,9 @@
+dpl_url_proxy.configure:
+  path: 'admin/config/services/dpl-url-proxy'
+  defaults:
+    _form: '\Drupal\dpl_url_proxy\Form\ProxyUrlConfigurationForm'
+    _title: 'Configure proxy urls'
+  requirements:
+    _permission: 'access administration pages'
+  options:
+    _admin_route: TRUE
--- a/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
+++ b/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
@@ -4,7 +4,7 @@ dpl_url_proxy.configure:
     _form: '\Drupal\dpl_url_proxy\Form\ProxyUrlConfigurationForm'
     _title: 'Configure proxy urls'
   requirements:
-    _permission: 'access administration pages'
+    _permission: 'administer library agency configuration'
   options:
     _admin_route: TRUE
 
@@ -13,7 +13,8 @@ dpl_url_proxy.generate_url:
   defaults:
     _controller: '\Drupal\dpl_url_proxy\Controller\DplUrlProxyController::generateUrl'
     _title: 'Generate proxy url'
+  methods:  [POST]
   requirements:
     _permission: 'access content'
   options:
-    no_cache: 'TRUE'
+    no_cache: TRUE
--- a/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
+++ b/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
@@ -13,8 +13,5 @@ dpl_url_proxy.generate_url:
   defaults:
     _controller: '\Drupal\dpl_url_proxy\Controller\DplUrlProxyController::generateUrl'
     _title: 'Generate proxy url'
-  methods:  [POST]
   requirements:
     _permission: 'access content'
-  options:
-    no_cache: TRUE
--- a/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
+++ b/web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
@@ -4,7 +4,7 @@ dpl_url_proxy.configure:
     _form: '\Drupal\dpl_url_proxy\Form\ProxyUrlConfigurationForm'
     _title: 'Configure proxy urls'
   requirements:
-    _permission: 'administer library agency configuration'
+    _permission: 'administer url proxy configuration'
   options:
     _admin_route: TRUE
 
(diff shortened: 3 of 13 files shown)


web/modules/custom/dpl_url_proxy/src/Form/ProxyUrlConfigurationForm.php
```diff
@@ -0,0 +1,303 @@
...
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId(): string {
+    return 'form_api_example_ajax_addmore';
+  }
```
kasperg, line 227: Please correct the form id.

`form_api_example_ajax_addmore` does not seem right.

web/modules/custom/dpl_url_proxy/tests/src/Unit/DplUrlProxyControllerTest.php
```diff
@@ -0,0 +1,200 @@
...
+      $request = Request::create(
+        '/dpl-url-proxy/generate-url',
+        'POST', [], [], [], [], $request_body
+      );
+      $controller->generateUrl($request);
+    }
+  }
+
+  /**
+   * Data provider for testThatUrlIsGenerated.
```
kasperg, line 117: Please fix the DocBlock comment.

This is not a data provider.

web/modules/custom/dpl_url_proxy/tests/src/Unit/DplUrlProxyControllerTest.php
```diff
@@ -0,0 +1,200 @@
...
+use Drupal\Core\DependencyInjection\ContainerBuilder;
+use Drupal\Core\StringTranslation\TranslationManager;
+use Drupal\dpl_url_proxy\Controller\DplUrlProxyController;
+use Drupal\dpl_url_proxy\DplUrlProxyInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use function Safe\json_encode;
+
+/**
+ * Unit tests for the Library Token Handler.
```
kasperg, line 19: :+1: for adding a unit test. Please remember to update the DocBlocks.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -126,6 +127,15 @@ public static function authUrl(): string {
     );
   }
 
+  /**
+   * Builds an url for the react apps to for proxy urls.
+   */
+  public static function urlProxyUrl(): string {
+    return self::ensureUrlIsString(
+      Url::fromRoute('dpl_url_proxy.generate_url')->toString()
```
kasperg, line 135: Consider whether this creates a dependency on `dpl_url_proxy`.

Perhaps it is only a soft dependency if `url-proxy-url` is not required. In the current form it seems like a hard dependency.

web/modules/custom/dpl_url_proxy/dpl_url_proxy.info.yml
```diff
@@ -0,0 +1,6 @@
+name: 'dpl_url_proxy'
```
kasperg, line 1: Please add a proper name for the module.

web/modules/custom/dpl_url_proxy/src/Form/ProxyUrlConfigurationForm.php
```diff
@@ -0,0 +1,303 @@
...
+
+  /**
+   * Create indexes used for the host names add more/delete part.
+   *
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   Drupal form state.
+   * @param mixed[] $values
+   *   Saved configuration.
+   *
+   * @return array<int|string>
```
kasperg, line 57: Consider if we can replace `array<int|string>` with `mixed[]' to fix the problem with invalid return types in the PHPCS sniff. [It seems to be a problem there](https://www.drupal.org/project/coder/issues/3303625).

An alternative solution would be to drop the Drupal return type definition sniff entirely. I for one prefer actual return type definitions or PHPStan ones anyway :smile:

web/modules/custom/dpl_url_proxy/src/DplUrlProxyInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Drupal\dpl_url_proxy;
+
+/**
+ * Interface for dpl_url_proxy constants.
+ */
+interface DplUrlProxyInterface {
```
kasperg, line 8: Consider if we can come up with a better name suffix than `Interface` for interfaces.

It does not provide any value and we do not intend it to be used as a "normal" interface which other classes can implement anyway.

web/modules/custom/dpl_url_proxy/src/Form/ProxyUrlConfigurationForm.php
```diff
@@ -0,0 +1,303 @@
...
+      function ($carry, $item) {
+        if (!empty($item['hostname'])) {
+          unset($item['remove_this']);
+          $carry[] = $item;
+        }
+        return $carry;
+      }, []);
+
+    $this->config('dpl_url_proxy.settings')
+      ->set('values', $values)
```
kasperg, line 297: I considered if we could do better than simply storing arrays of values. With [PHP 8.0 and constructor property promotion](https://stitcher.io/blog/constructor-promotion-in-php-8) (and read-only properties in PHP 8.1) we have a pretty simple way to create value objects. However then we would need some way to serialize/deserialize these objects from the configuration. Then I thought out this would do for now :smile:.

web/modules/custom/dpl_url_proxy/src/Controller/DplUrlProxyController.php
```diff
@@ -0,0 +1,118 @@
...
+    if (!$prefix = $conf['prefix'] ?? NULL) {
+      throw new HttpException(500, 'Could not generate url. Insufficient configuration');
+    }
+
+    // Search host names.
+    foreach ($conf['hostnames'] as $config) {
+      if ($host == $config['hostname']) {
+        // Rewrite/convert url using regex.
+        if (
+            (isset($config['expression']['regex']) && !empty($config['expression']['regex']))
```
kasperg, line 93: This logic of modifying an url based on configuration could be moved to a separate service.

web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
```diff
@@ -0,0 +1,20 @@
...
+    _title: 'Configure proxy urls'
+  requirements:
+    _permission: 'administer library agency configuration'
+  options:
+    _admin_route: TRUE
+
+dpl_url_proxy.generate_url:
+  path: '/dpl-url-proxy/generate-url'
+  defaults:
+    _controller: '\Drupal\dpl_url_proxy\Controller\DplUrlProxyController::generateUrl'
```
kasperg, line 14: This is getting to a state where we are beginning to expose an API which can be consumed by others.

The current use case is very simple but when we get to search campaigns etc. things will get more complex. I suggest that we consider which paradigm we want to work with going forward (REST, JSON:API, GraphQL) so all parties use that for future work.

kasperg (changes requested):
Good stuff. :+1: for unit tests and putting an extra effort into the UI.

There is quite a lot of code here. I have some comments and considerations.

web/modules/custom/dpl_url_proxy/src/Form/ProxyUrlConfigurationForm.php
spaceo: I forsee a future PR with removal of the Drupal return type definition sniff

web/modules/custom/dpl_url_proxy/src/DplUrlProxyInterface.php
spaceo: I get your point about the "mis"use of an interface here.
The thing is I think I started solving the issue about having constants stored somewhere in code like this after I tried doing it the old D7 way with just including them in a module file. But then phpcs was complaining and saying this was the way to do it. 
Another solution could be to create a `final class`.

web/modules/custom/dpl_url_proxy/src/DplUrlProxyInterface.php
```diff
@@ -0,0 +1,15 @@
...
+
+  const CONFIG_NAME = 'dpl_url_proxy.settings';
+  const TRANSLATION_OPTIONS = [
+    'context' => 'dpl_url_proxy',
```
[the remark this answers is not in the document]
spaceo: True

web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
```diff
@@ -0,0 +1,20 @@
...
+    _title: 'Generate proxy url'
+  methods:  [POST]
+  requirements:
+    _permission: 'access content'
+  options:
+    no_cache: TRUE
```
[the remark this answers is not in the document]
spaceo: I agree I've changed it to GET and added cache handling

web/modules/custom/dpl_url_proxy/src/Form/ProxyUrlConfigurationForm.php
spaceo: 👍

web/modules/custom/dpl_url_proxy/src/Controller/DplUrlProxyController.php
spaceo: Yes somewhere in the future ;)

web/modules/custom/dpl_url_proxy/dpl_url_proxy.routing.yml
spaceo: Needs discussion in the architecture forum

cableman (approved):

PR #141, opened 2022-10-05
Url proxy endpoint featuring OpenAPI

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-245

#### Description

The purpose of this PR is twofold:

1. We need to expose an endpoint which supports transforming an "unproxied" url to a url which has been modified to support proxies used by a library agency for access control based on the configuration provided by the library. This endpoint can be used by the React components which render data containing unproxied urls. This is handled by the `UrlProxyResource` class.
2. We want a standardized approach to exposing functionality and data from Drupal to external parties. We have looked into our options and argue that we should use Drupals REST and OpenAPI modules.

This PR does the following:

- Add and patch the required modules to provide a valid OpenAPI specification in `composer.json`
- Replace the url proxy controller and routing with a resource and configuration and update tests accordingly
- Add an initial OpenAPI 2.0 specification in `openapi.json`
- Add a GitHub Actions job with tasks to ensure that our `openapi.json` specification stays valid and does not drift
- Add documentation describing reasoning and future development process for API endpoints

#### Checklist


config/sync/user.role.anonymous.yml
```diff
@@ -5,8 +5,10 @@ dependencies: {  }
 _core:
   default_config_hash: j5zLMOdJBqC0bMvSdth5UebkprJB8g_2FXHqhfpJzow
 id: anonymous
-label: 'Anonymous user'
+label: 'Anonym bruger'
```
cableman, line 8: Why the label changes to danish?

config/sync/user.role.authenticated.yml
```diff
@@ -5,7 +5,7 @@ dependencies: {  }
 _core:
   default_config_hash: dJ0L2DNSj5q6XVZAGsuVDpJTh5UeYkIPwKrUOOpr8YI
 id: authenticated
-label: 'Authenticated user'
+label: 'Godkendt bruger'
```
cableman, line 8: Why the label changes to danish?

cableman (changes requested):

PR #142, opened 2022-10-10
Add correct url prop for the url proxy endpoint

spaceo:
#### Description

Since the proxy handling changed from controller to rest resource we need to refer to another route name.

#### Checklist


kasperg (approved):
:+1: LGTM

cableman (approved):

PR #143, opened 2022-10-10
Create kampagne

spaceo:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-238

#### Description

This PR sets up the campaign content type with referenced campaign rule paragraphs.
What is handled:
- Campaign content type with corresponding fields added
- Paragraph for handling campaign rules added
- Hook that ensures that user 1 is running admin interface in english. A side effect to that is that exported config language is in english too.
- Campaign CRUD permissions for editor and the two admin roles.

#### Checklist

This is the "only" the setup of the content type. A PR with a rest resource endpoint for resolving a campaign from search terms and facets will follow.


config/sync/node.type.campaign.yml
```diff
@@ -0,0 +1,11 @@
+uuid: 37edecd9-3cc1-4ab7-81b9-7a34d0c2c913
+langcode: da
+status: true
+dependencies: {  }
+name: Kampagne
+type: campaign
+description: 'Kampagne der kan vises på søgeresultater med mathende facetter og termer'
```
kasperg, line 7: Please ensure that there configurations are in English by default and translateable to Danish.

config/sync/paragraphs.paragraphs_type.campaign_rule.yml
```diff
@@ -0,0 +1,10 @@
+uuid: d2050f18-57d7-4791-ba90-a4e15399f0fb
+langcode: da
+status: true
+dependencies: {  }
+id: campaign_rule
+label: 'Kampagne regel'
+icon_uuid: null
+icon_default: null
+description: 'En regel for udvælgelse af en given kampagne'
```
kasperg, line 9: As above.

config/sync/core.base_field_override.node.campaign.promote.yml
```diff
@@ -0,0 +1,22 @@
...
+langcode: da
+status: true
+dependencies:
+  config:
+    - node.type.campaign
+id: node.campaign.promote
+field_name: promote
+entity_type: node
+bundle: campaign
+label: 'Forfremmet til forside'
```
kasperg, line 11: Consider whether the new content type should have a promoted property.

I do not think it makes sense in the current context and should be left out if possible. On the other hand a the standard published property would make sense.

kasperg (changes requested):
I fear there is some trouble regarding the language of the exported configuration in this PR.

I wonder how we can fix it though...

config/sync/node.type.campaign.yml
spaceo: I thought that when langcode is "da" the name and description should be represented in the same language.
Like it is the case of the promoted field, which has all been generated by the system:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/143/files/6ef736f396a1ba614baee15f74ca3e6201997f9e#diff-44ed5b104914630c7c07b5eea2a98403cc5024925256512d337c2f3eea57b081

config/sync/core.base_field_override.node.campaign.promote.yml
spaceo: Allthough that this configuration is generated the UI behaves like the functionality has been deactivated and I don't see that I have more options than I already have been configuring:

Configuration of the content type:
image

Configuration of the edit formular:
image

config/sync/node.type.campaign.yml
spaceo: Language export problems should have been solved by this commit:
d0302ef6ce9713e69c322655e82ed0e302e964ec

kasperg (approved):
:+1: LGTM

cableman (approved):

kasperg:
Closed in favour of #154

PR #146, opened 2022-10-11
Set base url of the dpl cms API

spaceo:
#### Description

Set base url of the whole dpl cms API
Instead of only defining the proxy url endpoint.
Orval will take care of the rest.

#### Checklist


kasperg (approved):
:+1: LGTM

PR #131, opened 2022-09-06
Loans list page

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5387

#### Description

N/A

#### Screenshot of the result

N/A

#### Checklist

- [ ] My complies with [our coding guidelines](../docs/code_guidelines.md).
- [ ] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [ ] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

#### Additional comments or questions

N/A

PR #153, opened 2022-10-26
Admin interface for excluding branches

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-248

#### Description

This PR adds an administration interface for handling branches which should be excluded from certain parts of the system and logic for passing this information to the related React apps e.g. https://github.com/danskernesdigitalebibliotek/dpl-react/pull/191

To achieve this the following major changes are introduced:

- A client for FBS is generated based on the specification also used by the React components through OpenAPI Generator
- The client is used to retrieve all branches from FBS
- The admin interface is updated to present all branches and store the ones library staff choose to exclude as configuration
- All branches and excluded branch configuration are passed on to React applications as props

Note that the parts of this PR e.g. Drupal services and wrapper code relating to FBS may be better placed in a more generic module. There may be work on this in #131. Consequently this PR keeps all this logic in the local module. It can be moved to a `dpl_fbs` module in the future.

#### Screenshot of the result

Monosnap General Settings | DPL CMS 2022-10-26 23-35-15

#### Checklist


composer.json
```diff
@@ -17,10 +17,10 @@
             "type": "package",
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-react",
-                "version": "0.3.0-rc7",
+                "version": "0.3.0-rc8",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/release-release%2F4/dist.zip",
+                    "url": "https://github.com/reload/dpl-react/releases/download/release-blacklist-search-result/dist.zip",
```
kasperg, line 23: This should be reverted before this PR is merged.

spaceo (commented):
Please remember to document the packages directory

spaceo (approved):
LGTM 👍

web/modules/custom/dpl_library_agency/src/FbsApiFactory.php
```diff
@@ -0,0 +1,62 @@
...
+    $this->client = $client;
+    $this->tokenHandler = $tokenHandler;
+  }
+
+  /**
+   * Assemble the API configuration.
+   */
+  protected function getConfiguration(): Configuration {
+    $configuration = (new Configuration())
+      ->setHost('https://fbs-openplatform.dbc.dk');
```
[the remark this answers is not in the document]
spaceo: We agreed upon a TODO describing that it could be configurable in the future.

web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
```diff
@@ -60,12 +86,54 @@ protected function getEditableConfigNames() {
...
+   *
+   * @param string $string
+   *   A string containing the English text to translate.
+   * @param mixed[] $args
+   *   Replacements to make after translation. Based on the first character of
+   *   the key, the value is escaped and/or themed.
+   * @param mixed[] $options
+   *   An associative array of additional options.
+   */
+  protected function t($string, array $args = [], array $options = []): TranslatableMarkup {
```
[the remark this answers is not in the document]
kasperg: We cannot do that because the type is not specified in the parent function.

packages/fbs/composer.json
```diff
@@ -0,0 +1,38 @@
+{
+    "name": "danskernesdigitalebibliotek/fbs",
```
turegjorup, line 2: A better name would be `danskernesdigitalebibliotek/fbs-php-client`

packages/fbs/.openapi-generator-ignore
```diff
@@ -0,0 +1,12 @@
+# OpenAPI Generator Ignore
+# Generated by openapi-generator https://github.com/openapitools/openapi-generator
+
+# Use this file to prevent files from being overwritten by the generator.
+# The patterns follow closely to .gitignore or .dockerignore.
+
```
turegjorup, line 6: Consider adding

```
docs/**/*.*
lib/**/*.*

!*/Api/ExternalV1AgencyidApi.*
!*/Model/AgencyBranch.*
```

to exclude files not needed

turegjorup (changes requested):
We should either
1. Generate a full blown FBS client and publish it to it's own repository and package
2. Keep the client here but limit the generation to ONLY the functionality needed

One way to do that is to expand the ignore file as shown. But I am unsure if I have excluded to much.

kasperg:
> We should either [...] Generate a full blown FBS client and publish it to it's own repository and package

I do not think publishing the FBS client to its own repository and package is a good idea at the moment.

1. I do not think it has the maturity to merit its' own package. It is based on an unofficial specification of the FBS proxy.
2. We do not have other users of the package that I am aware of.
3. The package it itself has little value. It is more or less the result of running a tool and committing the result. Others might as well do the same. The real value is the specification.
4. Publishing the results has its' own problems. We need Guzzle 6 for Drupal 9. Usage with other frameworks might require Guzzle 7. Instead of maintaining multiple versions in parallel centrally we might as well let each use case generate with their own appropriate configuration.

We have not published our generated TypeScript clients for FBS, Material List etc. either.

> We should either [...] Keep the client here but limit the generation to ONLY the functionality needed

I would not mind us doing using the approach you suggested so but it would be good to get a shared understanding of why we think this is needed.

We have not done so for our generated TypeScript REST clients at the moment.

turegjorup:
> I do not think publishing the FBS client to its own repository and package is a good idea at the moment.

Mostly agree, and not important relative to this PR. (Longer term I would prefer a published client dependent on a PSR-18 http client. I think this would still be easier to adopt in other projects than everybody duplicating auto generation. But that is out of scope of this PR).

> I would not mind us doing using the approach you suggested so but it would be good to get a shared understanding of why we think this is needed.

2 reasons. 

- Having excess code that we know will never be used is never a good idea IMHO. I don't believe there is anything such as "free" code maintenance wise.  Even if it is auto generated. So if we can scope this to the code actually needed then this is preferable. 
- In the NEXT architecture _most_ calls to external service should happen in the front end from the react clients. Taking care to actively exclude endpoint from the PHP clients clearly signals to developers where endpoints should be called from an architectural perspective. I would argue that something similar should be done in the TypeScript clients to either exclude this endpoint, or alter it so that the TypeScript clients knows to call Drupal for a list of branches. Because it's highly unlikely that a user facing part of NEXT should ever need to display excluded branches. Not doing so leaves it as a responsibility of the individual developers to know what not code to call and not to call.  Thsi is a little bit along the lines of thinking behind [PHPArkitect](https://github.com/phparkitect/arkitect) where you build rules to enforce architectural decisions.

kasperg:
@turegjorup I tried to address/reflect these decisions in [61d1184](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/153/commits/61d1184582d087794ca75726a93932a259552583).

turegjorup (approved):

PR #156, opened 2022-10-31
Expose campaigns

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-239

#### Description

This PR exposes an endpoint where an external application can fetch the most appropriate campaign from the based on the facets in a search result. The PR contains the following major changes:

- Configuration of a new content type: Campaign with fields (link, body and image) for defining where the link should direct the end user and the content for displaying it: Text and/or Image.
- A resource for exposing a campaign. This uses `POST` because of the size of the query data does not fit in the url and some part of the system seems to strip the body of `GET` requests.
- Cypress tests which ensures that the feature works as expected. The reason we use Cypress here is that the test involves much configuration (content type and rest resource setup) which cannot be easily handled using unit or kernel test cases.

The PR contains a number of changes we found necessary or suggest along the way:

- Additional UI modules are excluded from config exports
- The [`handy_cache_tags`](https://www.drupal.org/project/handy_cache_tags) Drupal module is added to support purging end points per content type
- Symfony packages needed to deserialize incoming JSON to PHP objects are added
- The `openapi_rest` module patch is updated with a fix
- Pa11y CI is updated to use a zip package af of the same SHA instead of a git checkout to reduce which software is need to use the package
- [API Cypress Plugin](https://github.com/filiphric/cypress-plugin-api) is added to facilitate easier end-to-end testing of APIs
- A hook is implemented for user 1 to ensure that language used on the admin interface for this user is always in English. This should ensure that the default configuration produced by this user is also always in English.

The PR contains expands upon #143 which was approved earlier.

#### Screenshot of the result

Screenshot 2022-10-31 at 14-39-23 Create Campaign DPL CMS

#### Checklist


web/modules/custom/dpl_campaign/src/Input/Facet.php
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Facet constructor.
+   *
+   * @param string $name
+   *   The name of the facet.
+   * @param \Drupal\dpl_campaign\Input\Value[] $values
+   *   The values of the facet.
+   */
+  public function __construct(string $name, array $values) {
+    $this->name = $name;
```
turegjorup, line 33: Why the difference in constructor style between? `Rule` and `Value` use PHP 8 style Constructor property promotion?

turegjorup (approved):
Question: Should there be OpenApi specs for this and other API's implemented in Drupal?

web/modules/custom/dpl_campaign/src/Input/Facet.php
kasperg: PHPStan (or the version that we currently use) does not support adding types to arrays when using constructor property promotion 😕 .

kasperg:
> Should there be OpenApi specs for this and other API's implemented in Drupal?

@turegjorup, there is [an OpenAPI spec for the APIs exposed by Drupal](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/release/4/openapi.json).

[We have documented the decision in an ADR](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/release/4/documentation/architecture/adr-006-api-specification.md).

PR #160, opened 2022-11-02
Drop registry credentials

danquah:
This is probably a leftover from a .lagoon.yml file from a "real" library site which needs to pull source-releases
from ghcr

#### Link to issue

https://reload.atlassian.net/browse/DDFDPDEL-212

#### Description

During the re-install of Lagoon I realized that we have a section in `.lagoon.yml` for getting privileged access to the Github container registry. This is a bit wird as the dpl-cms projct does not reference any images in that registry as of this point.

I've removed the section, and removed the personal access token from the dpl-cms project in Lagoon.

If I'm right this PR should be able to produce a functional PR site.

#### Checklist

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -53,11 +53,3 @@ environments:
       schedule: "M/15 * * * *"
       command: drush cron
       service: cli
-
-container-registries:
-  github:
-    username: any-user-works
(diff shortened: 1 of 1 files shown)


kasperg (approved):
:+1: LGTM

PR #165, opened 2022-11-07
Ensure Composer install does not modify files in VCS

kasperg:
#### Description

Installing the project should not infer unintended and/or unrelated file changes. It will usually be the result of problems with Drupal scaffold configuration.

Add a check in GitHub Actions to ensure that this is not the case and update the project accordingly.

#### Checklist


spaceo (approved):
LGTM

turegjorup (approved):
👍

PR #172, opened 2022-11-15
Adding title and id in campaign match endpoint

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-317

#### Description

We need title for statistics right now.
Adding id also because we might need that in the future too.

#### Checklist

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -222,4 +222,4 @@ jobs:
       - name: Download current specification
         run: task ci:openapi:download
       - name: Ensure specification has not drifted
-        run: git diff --exit-code openapi.json
+        run: git diff --ignore-space-at-eol --exit-code openapi.json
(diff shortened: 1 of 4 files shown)


kasperg (changes requested):
CI checks are failing.

- Remember to update the specification in VCS
- Remember to Cypress tests accordingly

kasperg (approved):
:+1: LGTM

.github/workflows/ci-tests.yml
```diff
@@ -222,4 +222,4 @@ jobs:
       - name: Download current specification
         run: task ci:openapi:download
       - name: Ensure specification has not drifted
-        run: git diff --exit-code openapi.json
+        run: git diff --ignore-space-at-eol --exit-code openapi.json
```
cableman, line 225: Why do we need to `ignore spaces at end of line` when comparing api spec. It should be generated by the same code and thereby be identical?

cableman (approved):
Approved with comment

.github/workflows/ci-tests.yml
kasperg: We tried to explain in the commit comment adding the change:

> We'll allow spaces in eol to ignore diffs based on editor cleanup settings.

PR #174, opened 2022-11-16
Disable git safe.directory protection on OSX

kasperg:
#### Description

[Because of a git security vulnerability a new `safe.directory` configuration was introduced](https://github.blog/2022-04-12-git-security-vulnerability-announced/).

This causes problems with volumes on OSX which has issues with handling file system permissions between Docker host and client.

Normally this is not an issue but Composer packages which are  distributed through git instead of zip files it is. Composer will try to clone the package on a mounted volume with dubious permissions. `drupal/coder` is such a package. This shows as running task dev:reset will fail with the following error:

```
fatal: detected dubious ownership in repository at '/app/vendor/drupal/coder'
```

To address the issue we configure git in the CLI container responsible for running Composer to mark all directories as safe effectively  disabling `safe.directory` handling. To limit the scope of this we only do this on OSX in development environment.

This has probably not been an issue earlier because we have been using images with an older version of Git without `safe.directory` handling.

#### Checklist

### DIFF
--- a/docker/cli/.gitconfig
+++ b/docker/cli/.gitconfig
@@ -1,0 +1,2 @@
+[safe]
+    directory = *
--- a/docker-compose.mac-nfs.yml
+++ b/docker-compose.mac-nfs.yml
@@ -1,4 +1,9 @@
-version: '3'
+version: "3"
+
+services:
+  cli:
+    volumes:
+      - "./docker/cli/.gitconfig:/home/.gitconfig"
 
 volumes:
   projectroot:


docker/cli/.gitconfig
```diff
@@ -0,0 +1,2 @@
+[safe]
```
spaceo, line 1: Consider writing a comment here why we want consider all directories to be safe and in which context.

spaceo (commented):
LGTM. Have one comment...

docker/cli/.gitconfig
kasperg: @spaceo I have tried to explain this in the commit comment. To you prefer an inline comment in the `.gitconfig` file?

spaceo: Hmm. Don't know if that is even possible. And it does also not really belong in the docker compose file. So lets keep it like this.

spaceo (approved):
LGTM

turegjorup (approved):

PR #177, opened 2022-11-24
Change dpl-* dependencies to use develop releases

spaceo:
#### Description

Since we change to develop branches we need to change it in composer.json accordingly

#### Checklist


kasperg (approved):
:+1: LGTM

turegjorup (approved):

PR #178, opened 2022-11-24
Add go-to-text placeholder

Adamik10:
#### Link to issue
DDFSOEG-301

#### Description
In order to maintain all the translatable system strings matching the dpl-react repository, I had to add a placeholder into the go-to-text system string. Also, the go-to-e-reolen-text is deprecated because of the introduction of string placeholders, so I removed it (it wasn't being used anywhere).

#### Checklist

n/a

### DIFF
--- a/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
+++ b/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
@@ -201,7 +201,6 @@ class DplReactAppsController extends ControllerBase {
         'find-on-shelf-modal-screen-reader-modal-description-text' => $this->t('Reservation modal screen reader description', [], $c),
         'genre-and-form-text' => $this->t('Genre', [], $c),
         'get-online-text' => $this->t('Get online', [], $c),
-        'go-to-e-reolen-text' => $this->t('Go to e-Reolen', [], $c),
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -202,7 +202,7 @@ public function work(string $wid): array {
         'genre-and-form-text' => $this->t('Genre', [], $c),
         'get-online-text' => $this->t('Get online', [], $c),
         'go-to-e-reolen-text' => $this->t('Go to e-Reolen', [], $c),
-        'go-to-text' => $this->t('Go to', [], $c),
+        'go-to-text' => $this->t('Go to @source', [], $c),
```
kasperg, line 205: Please consider if the text right above this change specifically for eReolen is still relevant now that we have placeholders.

kasperg (approved):
:+1: LGTM but please double check which texts are needed here before moving this along.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
Adamik10: Great point. I will remove it from dpl-react too to avoid confusion.

sinejespersen (approved):

PR #179, opened 2022-11-28
Digital Archive Service module

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-304

#### Description

This PR adds a new dpl_das module which provides integration between the DPL platform and the Digital Archive Service (Elba webservices) provided by the Royal Danish Library.

Besides implementing the modue with the integration the PR provides two main changes;

1. Implementing a Drupal authentication provider for Adgangsplatformen tokens. This is used to protect the endpoint against unauthorized access.
2. Refactoring Adgangsplatformen configuration access. To avoid passing out strings and unstructured arrays this central configuration is moved to a couple of classes.

Please read through the commits for additional commentary.

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_das/src/Elba/Client.php
+++ b/web/modules/custom/dpl_das/src/Elba/Client.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Drupal\dpl_das\Elba;
+
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\GuzzleException;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\RequestOptions;
+use Symfony\Component\Serializer\Encoder\EncoderInterface;
+
+/**
+ * Client for Elba webservices.
+ *
+ * These web services are provided by the Royal Danish Library.
+ *
+ * @see https://webservice.statsbiblioteket.dk/elba-webservices/
+ */
+class Client {
+
+  /**
+   * Constructor.
+   */
+  public function __construct(
+    private ClientInterface $client,
+    private EncoderInterface $encoder,
+    // Note that the uri is all lower case despite the provided documentation.
+    // This is intentional and required for the service to work.
+    private string $uri = "https://webservice.statsbiblioteket.dk/elba-webservices/services/placecopyrequest",
+  ) {}
+
+  /**
+   * Order a digital copy of an article to be sent to the patron by email.
+   */
+  public function placeCopy(PlaceCopyRequest $placeCopy): void {
+    // The XML encoder does not provide a way to add a namespace to objects.
+    // Convert it to an array to fix this.
+    $request = (array) $placeCopy;
+    $request['@xmlns'] = "http://statsbiblioteket.dk/xws/elba-placecopyrequest-schema";
+
+    $xml = $this->encoder->encode($request, "xml", [
+      // Existing integrations use XML 1.0. Keep doing to for consistency.
+      "xml_version" => "1.0",
+      "xml_encoding" => "utf-8",
+      "xml_root_node_name" => "placeCopyRequest",
+    ]);
+
+    try {
+      $this->client->send(new Request(
+        "POST",
+        $this->uri,
+        // Content type matches encoding.
+        ["Content-Type" => "application/xml; charset=UTF8"],
+        $xml
+      ), [
+        // This is intentionally set high because the service is known to be
+        // slow. Typical response times are about 10 seconds.
+        RequestOptions::TIMEOUT => 30,
+      ]);
+    }
+    catch (GuzzleException $e) {
+      throw new Exception($e);
+    }
+  }
+
+}
--- a/web/modules/custom/dpl_das/src/Elba/Exception.php
+++ b/web/modules/custom/dpl_das/src/Elba/Exception.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Drupal\dpl_das\Elba;
+
+/**
+ * An error returned by the Elba webservice.
+ */
+class Exception extends \RuntimeException {
+
+}
--- a/web/modules/custom/dpl_das/src/Form/SettingsForm.php
+++ b/web/modules/custom/dpl_das/src/Form/SettingsForm.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace Drupal\dpl_das\Form;
+
+use Drupal\Core\Form\ConfigFormBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+
+/**
+ * Configure Digital Archive Service settings for this site.
+ */
+class SettingsForm extends ConfigFormBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'dpl_das_settings';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getEditableConfigNames() {
+    return ['dpl_das.settings'];
+  }
+
+  /**
+   * Translates a string to the current language or to a given language.
+   *
+   * @param string $string
+   *   A string containing the English text to translate.
+   * @param mixed[] $args
+   *   Replacements to make after translation. Based on the first character of
+   *   the key, the value is escaped and/or themed.
+   * @param mixed[] $options
+   *   An associative array of additional options.
+   */
+  protected function t($string, array $args = [], array $options = []): TranslatableMarkup {
+    // Intentionally transfer the string originally passed to t().
+    // phpcs:ignore Drupal.Semantics.FunctionT.NotLiteralString
+    return parent::t($string, $args, array_merge($options, ['context' => 'Digital Article Service']));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form['webservice'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Webservice integration'),
+    ];
+    $form['webservice']['url'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Url'),
+      '#description' => $this->t('The service endpoint for "placeCopyRequest"'),
+      '#default_value' => $this->config('dpl_das.settings')->get('url') ?? "https://webservice.statsbiblioteket.dk/elba-webservices/services/placecopyrequest",
+    ];
+    $form['webservice']['username'] = [
+      '#type' => 'textfield',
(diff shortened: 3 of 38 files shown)


web/modules/custom/dpl_login/src/Routing/DisableCsrfCheckSubscriber.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * We hold the CMS to the same security levels as other business systems in the
+ * NEXT platform. Thus we deem a valid user token sufficient and we remove
+ * this requirement for all routes using the user token authentication
+ * provider.
+ *
+ * @see \Drupal\Core\Access\CsrfRequestHeaderAccessCheck
+ * @see \Drupal\dpl_login\UserTokenAuthProvider
+ */
+class DisableCsrfCheckSubscriber extends RouteSubscriberBase {
```
spaceo, line 22: Maybe `DangerouslyDisableCsrfCheckSubscriber`? 😜

web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Structured access to configuration for Adgangsplatformen.
+ *
+ * Adgangsplatformen is implemented as a plugin for the OpenID Connect Drupal
+ * module but we need access to the stored values elsewhere in the system.
+ * This class provides structured access to what would otherwise be an map
+ * of strings.
+ */
+class Config {
```
spaceo, line 17: Nice with an isolated "entity" for handling Adgangsplatformen config. 👍

web/modules/custom/dpl_login/src/UserTokenAuthProvider.php
```diff
@@ -0,0 +1,75 @@
...
+  ) {}
+
+  /**
+   * Extract a bearer token from the authorization header of a request.
+   */
+  private function getToken(Request $request) : ?string {
+    $header = $request->headers->get('Authorization');
+    if (!$header) {
+      return NULL;
+    }
```
spaceo, line 34: It is a matter of taste but I usually solve situations like this:
```php
    if (!$header = $request->headers->get('Authorization')) {
      return NULL;
    }

File: web/modules/custom/dpl_login/src/UserTokenAuthProvider.php [hunk 4]
```diff
@@ -0,0 +1,75 @@
...
+
+    preg_match('/^Bearer\s+(\w+)/', $header, $matches);
+    return $matches[1] ?? NULL;
```
spaceo, line 37: Can also just be:
```php
return preg_match('/^Bearer\s+(\w+)/', $header) ?? NULL;

File: web/modules/custom/dpl_das/src/Plugin/rest/resource/DigitalArticleOrderResource.php [hunk 5]
```diff
@@ -0,0 +1,172 @@
...
+ *     400 = {
+ *      "descriptions" = "Invalid input",
+ *     },
+ *     500 = {
+ *       "description" = "Internal server error",
+ *     },
+ *   }
+ * )
+ */
+class DigitalArticleOrderResource extends ResourceBase {
```
spaceo, line 71: I think we should have a test for the endpoint

web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php
spaceo: Consider making a unit test for the class

web/modules/custom/dpl_das/src/Elba/Client.php
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Constructor.
+   */
+  public function __construct(
+    private ClientInterface $client,
+    private EncoderInterface $encoder,
+    // Note that the uri is all lower case despite the provided documentation.
+    // This is intentional and required for the service to work.
+    private string $uri = "https://webservice.statsbiblioteket.dk/elba-webservices/services/placecopyrequest",
```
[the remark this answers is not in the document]
kasperg: Good question but I do not think it makes sense to change the current implementation:

1. [The `placeCopyRequest` is the only endpoint the implementation in the ding2 project has called](https://github.com/ding2/ding2/blob/master/modules/ting_das/includes/ting_das.admin.inc#L35) since the module was introduced there >1.5 years ago.
2. It is likely that this entire integration will be moved to FBI as they have also just [implemented Elba support](https://fbi-api.dbc.dk/documentation#Elba-services) and it is probably preferrable to move the implementation there in the future.

If we are yo change this I could split the url into two parts: The base url and `placecopyrequest`.

web/modules/custom/dpl_das/src/Elba/Exception.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Drupal\dpl_das\Elba;
+
+/**
+ * An error returned by the Elba webservice.
+ */
+class Exception extends \RuntimeException {
```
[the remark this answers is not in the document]
kasperg: Good point. Actually [the Drupal Coding standards state](https://www.drupal.org/docs/develop/standards/object-oriented-code):

> Classes and interfaces should have names that stand alone to tell what they do without having to refer to the namespace, read well, and are as short as possible without losing functionality information or leading to ambiguity. 

This means that `ElbaException` would probably be a more telling name. Based on my comment above regarding the future of this let me know if you want me to change this.

web/modules/custom/dpl_das/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,86 @@
...
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form['webservice'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Webservice integration'),
+    ];
+    $form['webservice']['url'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Url'),
+      '#description' => $this->t('The service endpoint for "placeCopyRequest"'),
+      '#default_value' => $this->config('dpl_das.settings')->get('url') ?? "https://webservice.statsbiblioteket.dk/elba-webservices/services/placecopyrequest",
```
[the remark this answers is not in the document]
kasperg: Good point. Based on the future of the integration do you want me to change this?

web/modules/custom/dpl_das/src/Plugin/rest/resource/DigitalArticleOrderResource.php
kasperg: Good point. Based on the future of the integration do you want me to implement this?

web/modules/custom/dpl_das/src/Plugin/rest/resource/DigitalArticleOrderResource.php
```diff
@@ -0,0 +1,163 @@
...
+    $instance->serializer = $container->get('dpl_das.serializer');
+    $instance->client = $container->get('dpl_das.client');
+    $instance->configManager = $container->get('config.manager');
+
+    return $instance;
+  }
+
+  /**
+   * Receive incoming requests to order digital articles.
+   */
```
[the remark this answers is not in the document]
kasperg: Personally I am not a fan of declaring `@throws` blocks. It is not something we require from our coding guidelines. If you like it I can add it but it may become somewhat random in regard to where we use it and where we do not. We could enforce it through code analysis if we wanted.

Based on the future of the integration do you want me to implement this?

web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Returns the configuration as an array.
+   *
+   * This is primarily for usage with the OpenID Connect plugin system which
+   * also expects array storage.
+   *
+   * @return string[]
+   *   Map of Adgangsplatformen configuration.
+   */
+  public function asArray() : array {
```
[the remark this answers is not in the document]
kasperg: Could you elaborate here? Serialization generally turns objects into string form. That is not what we are lokking for. Perhaps you are wondering whether PHP has an interface or the like we can implement to make an object pass as an array for function arguments etc. but that does not seem to be the case.

[Larry Garfield has a nice post about the topic](https://steemit.com/php/@crell/php-never-type-hint-on-arrays) but as long as Drupal still uses `array` type hints there does not seem to be much we can do about it.

web/modules/custom/dpl_login/src/UserTokenAuthProvider.php
kasperg: I have noticed and we differ in taste here :smile:

In my opinion assignments in conditions leads to less readable code and risk problems of mixing up assignment and comparison.

web/modules/custom/dpl_login/src/UserTokenAuthProvider.php
kasperg: I do not think so. `preg_match()` returns `0|1|false` depending on whether we have a match or not. We need it to return the actual match.

web/modules/custom/dpl_das/src/Plugin/rest/resource/DigitalArticleOrderResource.php
```diff
@@ -0,0 +1,163 @@
...
+    catch (UnexpectedValueException $e) {
+      throw new HttpException(400, "Invalid article order data: {$e->getMessage()}}");
+    }
+
+    $config = $this->configManager->getConfigFactory()->get('dpl_das.settings');
+
+    $request = new PlaceCopyRequest(
+      $config->get('username'),
+      $config->get("password"),
+      "775100",
```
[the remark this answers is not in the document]
kasperg: This has been addressed in c810ef56fdb91c79b29bdc03385bf3ca5e74202b.

web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php
kasperg: Implemented in eb712d5.

web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php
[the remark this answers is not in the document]
spaceo: 👍

web/modules/custom/dpl_login/src/UserTokenAuthProvider.php
spaceo: 👍

spaceo (approved):
LGTM 👍

PR #181, opened 2022-11-28
opsaetning af a11y check i gh actions

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-324

#### Description

This PR introduces a function CI setup for Pa11y.
When GH CI workflow is running a "Pa11y" action is running which runs:
* Sets up a running dpl-cms site
* Runs a `pa11y:ci` task invoking the yarn cmd:  `pa11y:ci`

The "Pa11y:ci" yarn cmd does following:
* Builds the pa11y js setup from .ts files
* Create mappings for various requests connected to the test pages defined in .pa11yci
* Runs the pa11y-ci node application

#### Checklist

This the initial implementation pa11y setup which enables us to run the Pa11y tests in our CI on the front page, search page and work page.

Some things still needs to be taken care of:
* Establish a setup for local development with Pa11y. We want to be able to run 1 or more task cmds that builds a running version of dpl-cms that is ready for local "ci testing" and that opens up a browser in a mode that allows us to "see" what pa11y is testing
* Make it possible to configure all the basurls for the various services the dpl-react and dpl-cms. In that way we can use http domains to circumvent browser ssl sercurity checks which blocks us testing what pa11y "sees" (as described in previous point)


pa11y/.gitignore
```diff
@@ -0,0 +1,3 @@
+build
+node_modules
```
kasperg, line 2: Please remove `node_modules`.

I think it is a leftover from when `pa11y` subfolder had it's own `package.json`.

pa11y/lib/general.ts
```diff
@@ -0,0 +1,11 @@
+import { WireMockRestClient } from "wiremock-rest-client";
+import { Options } from "wiremock-rest-client/dist/model/options.model";
+
+export const matchGraphqlQuery = (id: string) =>
+  `$.[?(@.query=~/.*query ${id}\\(.*/s)]`;
+
+export const wiremock = (baseUri?: string, options?: Options) => {
+  return new WireMockRestClient("http://wiremock", options);
```
kasperg, line 8: Consider if the wiremock host should be defined as an environment variable instead of hardcoded.

pa11y/mappings/common/createCommonMappings.ts
```diff
@@ -0,0 +1,90 @@
...
+          id: "870970-basis:134693959",
+          type: "pid",
+          imageUrls: {
+            small: {
+              url: "https://picsum.photos/98/160",
+              format: "jpeg",
+              size: "small",
+            },
+            large: {
+              url: "https://picsum.photos/142/172",
```
kasperg, line 23: Please document these picsum urls.

Why are these specific urls relevant in the current context? Does the url arguments refer to any expected sizes? Was picsum chosen because it supports features we need, do we use it elsewhere in the project or perhaps it is just random? :smile:

pa11y/mappings/common/createCommonMappings.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  // Mapping for tokens (anonymous session).
+  wiremock(baseUri, options).mappings.createMapping({
+    request: {
+      method: "GET",
+      urlPath: "/dpl-react/user-tokens",
+    },
+    response: {
+      headers: {
+        "Content-Type": "application/txt",
```
kasperg, line 52: Please change the content type to `application/javascript`.

[This matches the actual response from the CMS](https://github.com/reload/dpl-cms/blob/opsaetning-af-a11y-check-i-gh-actions-pa-dpl-cms-DDFSOEG-324/web/modules/custom/dpl_react/src/Controller/DplReactController.php#L83-L83).

pa11y/mappings/common/createCommonMappings.ts
```diff
@@ -0,0 +1,90 @@
...
+      },
+      body: 'window.dplReact = window.dplReact || {};\nwindow.dplReact.setToken("library", "fcd5c29a171f97b626d71eceffe1313f00a284b0")',
+    },
+  });
```
kasperg, line 56: Consider if this is still relevant and/or works.

We have previously specifically ignored requests to the CMS going through Wiremock. If we really want to ensure specific tokens are set we might need to mock the responses from Adgangsplatformen to the CMS instead.

pa11y/mappings/common/createCommonMappings.ts
```diff
@@ -0,0 +1,90 @@
...
+  // Mapping for covers.
+  wiremock(baseUri, options).mappings.createMapping({
+    request: {
+      method: "GET",
+      urlPattern: "/api/v2/covers.*",
+    },
+    response: {
+      jsonBody: [
+        {
+          id: "870970-basis:134693959",
```
kasperg, line 14: Consider if it is relevant to template the id in the response from the request.

This always returns a fixed id.

pa11y/mappings/common/createCommonMappings.ts
```diff
@@ -0,0 +1,90 @@
...
+      urlPattern: "/external/agencyid/catalog/availability/v3\\?recordid=.*",
+    },
+    response: {
+      transformers: ["response-template"],
+      jsonBody: [
+        {
+          recordId: "{{request.query.recordid}}",
+          reservable: "{{pickRandom true false}}",
+          available: "{{pickRandom true false}}",
+          reservations: "{{randomInt lower=0 upper=10}}",
```
kasperg, line 71: Consider if it is relevant to document why we use random of these values.

pa11y/mappings/search/createMappingsForSearch.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+    });
+  });
+
+  // Get campaign.
+  wiremock(baseUri, options).mappings.createMapping({
+    request: {
+      method: "POST",
+      urlPath: "/dpl_campaign/match",
+    },
```
kasperg, line 64: See comment regarding mocking requests from the frontend to the CMS.

pa11y/mappings/work/data/covers/covers.json
```diff
@@ -0,0 +1,13 @@
+[
+  {
+    "id": "870970-basis:134693959",
+    "type": "pid",
+    "imageUrls": {
+      "small": {
+        "url": "https://res.cloudinary.com/dandigbib/image/upload/t_ddb_cover_small/v1647612411/bogportalen.dk/9788702307566.jpg",
+        "format": "jpeg",
+        "size": "small"
```
kasperg, line 9: Consider if this file is still relevant.

As far as I can tell it has been superseeded by a mock returning images from picsum.

pa11y/yarn.lock
```diff
@@ -0,0 +1,822 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kasperg, line 1: Please remove this file if it is no longer relevant.

As far as I can tell is it made irrelevant by using the root `package.json`.

package.json
```diff
@@ -20,5 +20,10 @@
     "prettier": "^2.7.1",
     "typescript": "^4.8.3",
     "wiremock-rest-client": "^1.10.0"
+  },
+  "scripts": {
+    "pa11y:create-mappings": "tsc -p pa11y && node pa11y/build/createMappings.js",
+    "pa11y:ci": "tsc -p pa11y && node pa11y/build/createMappings.js && pa11y-ci",
```
kasperg, line 26: Consider if `pa11y:create-mappings` can be called from `pa11y:ci`.

As far as I can tell the commands are duplicated. Perhaps this is intentional.

kasperg (commented):
:+1: Lots of work when into this. Good stuff.

Apart from the comments please look into whether we can/should lint the PA11y TS code. There are examples for how we do this for Drupal and Cypress in https://github.com/reload/dpl-cms/blob/opsaetning-af-a11y-check-i-gh-actions-pa-dpl-cms-DDFSOEG-324/.github/workflows/ci-tests.yml#L171-L171.

We might also want to setup yarn scripts for this.

pa11y/mappings/common/createCommonMappings.ts
spaceo: I'll document it.
The equivalent to this in Cypress is done via fixtures.
This was done as an easy way to get a picture dummy with proper rights.
Otherwise one could look into if Wiremock supports bas64 encoded mocks.
But I prioritized my time.

pa11y/mappings/search/createMappingsForSearch.ts
spaceo: Check

package.json
spaceo: It could :)

pa11y/yarn.lock
spaceo: Done

pa11y/mappings/work/data/covers/covers.json
spaceo: Removed

pa11y/.gitignore
spaceo: 👍

pa11y/lib/general.ts
spaceo: 👍

pa11y/mappings/common/createCommonMappings.ts
spaceo: 👍

pa11y/mappings/common/createCommonMappings.ts
spaceo: 👍

pa11y/mappings/common/createCommonMappings.ts
spaceo: It has been documented...

turegjorup (approved):

kasperg (approved):

PR #182, opened 2022-11-29
Test lagoons handling of cron for PRs

danquah:
#### Link to issue

None

#### Description

Test of how Lagoon handles cron for pr-environments. Should not be merged as the PR is making a change to `.lagoon.yml` that is specific to this pr.


danquah:
I've verified that the environment created for this PR did indeed run cronjobs.

PR #183, opened 2022-11-29
Prevent hard errors when branches cannot be retrieved from FBS

kasperg:
#### Description

Some React apps require the CMS to expose a list of branches. If the site has no valid library token (e.g. due to missing or invalid  Adgangsplatformen configuration or because cron has not run) then fetching branches will fail with an exception bubbling up and a hard error.

A softer approach would be to log the problem and pass an empty set of branches. This would allow much of the functionality of the app to work. On the other hard we risk missing real problems which appear down the stream e.g. when patrons interact with the apps and cannot see any branches.

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_library_agency/src/Branch/FallbackBranchRepository.php
+++ b/web/modules/custom/dpl_library_agency/src/Branch/FallbackBranchRepository.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Drupal\dpl_library_agency\Branch;
+
+use Psr\Log\LoggerInterface;
+
+/**
+ * Repository with a fallback option.
+ *
+ * If an error/exception occurs when retrieving branches from one repository
+ * this will fall back to a secondary option.
+ */
+class FallbackBranchRepository implements BranchRepositoryInterface {
+
+  /**
+   * Constructor.
+   */
+  public function __construct(
+    private BranchRepositoryInterface $primaryRepository,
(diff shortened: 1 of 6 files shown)


web/modules/custom/dpl_library_agency/src/Branch/FallbackBranchRepository.php
```diff
@@ -0,0 +1,42 @@
...
+  ) {}
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getBranches(): array {
+    try {
+      return $this->primaryRepository->getBranches();
+    }
+    catch (\Exception $e) {
```
spaceo, line 31: It is unclear to me where this exception is being thrown. I did a bit of investigation, and perhaps I missed it. But a @throws in the docblock of the primaryRepository:getBranches() would be nice.

spaceo (commented):
LGTM. Have one question though.

web/modules/custom/dpl_library_agency/src/Branch/FallbackBranchRepository.php
kasperg: As any branch repository implemenetation can be set as the primary I cannot see how I can set this in the fallback implementation.

I have updated the DocBlock in the FBS repository with a `@throws` block.

spaceo (approved):

turegjorup (approved):

PR #185, opened 2022-11-29
Enable XDebug per service

kasperg:
#### Description

Currently XDebug is enabled centrally in a single environment variable which is used across services. This makes it easy to enable but causes problems because it will cause both web requests and CLI tools to use  XDebug.

Debugging CLI tools with XDebug can be helpful but also cause problems because CLI tools e.g. Drush are often invoked during development. With Xdebug enabled this can cause slowness and breakpoints to be triggered which were intended for web requests.

Splitting up the values allow developers to enable XDebug selectively e.g. only for web requests through PHP-FPM.

#### Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,8 +33,6 @@ x-environment:
     WEBROOT: web
     # Uncomment if you like to have the system behave like in production
     #LAGOON_ENVIRONMENT_TYPE: production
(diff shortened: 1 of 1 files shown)


spaceo (approved):
LGTM 👍

jeppekroghitk (approved):

PR #186, opened 2022-11-30
Feature/loan list v2

turegjorup:
#### Link to issue

https://platform.dandigbib.org/issues/5387

#### Description

Replaces https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/131

#### Screenshot of the result

#### Checklist


docker-compose.yml
```diff
@@ -72,8 +72,6 @@ services:
       lagoon.type: cli-persistent
       lagoon.persistent.name: nginx # mount the persistent storage of nginx into this container
       lagoon.persistent: /app/web/sites/default/files/ # location where the persistent storage should be mounted
-    environment:
-      << : *default-environment # loads the defined environment variables from the top
```
cableman, line 76: Why this  change to docker setup?

web/modules/custom/dpl_loans/src/Controller/DplLoansController.php
```diff
@@ -0,0 +1,76 @@
...
+    /** @var \Drupal\dpl_loans\Plugin\Block\LoanListBlock $plugin_block */
+    $plugin_block = $this->blockManager->createInstance('dpl_loans_list_block', $config);
+
+    // Some blocks might implement access check.
+    $access_result = $plugin_block->access($this->currentUser());
+
+    // Return empty render array if user doesn't have access.
+    // $access_result can be boolean or an AccessResult class.
+    if (is_object($access_result) && $access_result->isForbidden() || is_bool($access_result) && !$access_result) {
+      // You might need to add some cache tags/contexts.
```
cableman, line 65: Maybe we should lock into ensuring the block is cached? Or remove update the comments in this function.

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -0,0 +1,158 @@
...
+    $data = [
+      // Config.
+      "fbs-base-url" => $fbsConfig->get('base_url'),
+      "publizon-base-url" => $publizonConfig->get('base_url'),
+      // This overrides config.
+      "page-size-desktop" => "25",
+      // This overrides config.
+      "page-size-mobile" => "25",
+      // Urls.
+      // @todo update placeholder URL's
```
cableman, line 80: Why this placeholder to images.... why are feed-page-url an image?

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -0,0 +1,158 @@
...
+    $context = ['context' => 'Loan list'];
+    $contextAria = ['context' => 'Loan list (Aria)'];
+
+    $fbsConfig = $this->configFactory->get('dpl_fbs.settings');
+    $publizonConfig = $this->configFactory->get('dpl_publizon.settings');
+
+    $data = [
+      // Config.
+      "fbs-base-url" => $fbsConfig->get('base_url'),
+      "publizon-base-url" => $publizonConfig->get('base_url'),
```
cableman, line 74: Should we refactor this to use the same syntax as `DplReactAppsController::searchResultUrl()`?

cableman (changes requested):

docker-compose.yml
turegjorup: Lines were duplicated. See commmit https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/186/commits/c1f62c1e0f62620309deedaf33d6793b58d7cff1

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
turegjorup: Yes/No. The URL's we get here are external URLs from config. The URL they get in `DplReactAppsController::searchResultUrl()` are internal URLs from routes son not directly comparable.

I think we should have an established way of getting URLs across module boundaries. You could argue that `dpl_loans`should not call an internal(?) function in `dpl_react_apps`.

This should be handled on an architectural level and as such out of scope of this PR.

PR #188, opened 2022-12-07
Enable cron in non-production environments through automated cron

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFDRIFT-13

#### Description

Lagoon does not support cron on PR environments at the moment. We need cron to run to support renewal on library tokens at the moment and would like to have it available in general.

As a workaround we enable the automated cron module which runs cron at the end of web requests. It is currently configured with the default interval of 3 hours.

We would not use this approach for a production environment so we use configuration-by-Lagoon-environment-type to never run automated crons there even though the module might still be enabled.

#### Checklist


spaceo (approved):
I can approve this. But is this the de facto way of handling dev vs prod handling of modules?
Isn't there a way only to enable modules in the development environment?

kasperg:
> But is this the de facto way of handling dev vs prod handling of modules? Isn't there a way only to enable modules in the development environment?

Good question. As I see it there is no way in default way to handle this in Drupal Core. The de facto way is to use the [Config Split module](https://www.drupal.org/project/config_split). [We have tried to set this up previously to lift configuration management for the project in general but failed](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/documentation/architecture/adr-001-configuration-management.md#config-split).

I am not terribly minded trying again to address this problem when it is possible to manage the variation with what is already provided by the module and our existing setup.

cableman (approved):

cableman:
Agreed, that it would be nice to be able to enable some modules based on `env`, but the only way I have seen is having two different configurations sync folders, but this do not seem to be a good solution and time consuming to keep in sync.

PR #189, opened 2022-12-07
Make the environment creation process more robust

danquah:
#### Link to issue

https://reload.atlassian.net/browse/DDFDRIFT-16

#### Description

We've seen some PRs that are not able to trigger environment-creation in Lagoon. The culprit is most likely some details around when contributers has access to which secrets.

This PR implements some checks in our actions to make it more clear when actions are unable to run due to missing actions, and does some work to improve the basic documentation of our actions so that a developer that experiences these problems knows what to do.

#### Checklist

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,5 +1,6 @@
 # These workflows takes care of various CI tests.
-on: pull_request
+# on: pull_request
+on: workflow_dispatch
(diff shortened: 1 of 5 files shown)


danquah:
Closing this. We may consider making the checks for secrets availability a bit more up front, but I suspect we'll know whats wrong going forward if we see these workflows fail with missing secrets.

PR #191, opened 2022-12-12
Reinstall React and design system packages during reset

kasperg:
#### Description

These packages track the develop branch release builds by default. The content of these builds change over time even though the package version does not. During development we always want to ensure we have the latest version so we reinstall the packages every time we reset.

Disabling the Composer cache is important as the develop release builds always use the same url for downlading. This url may be cached within Composer so we also have to ensure that the Composer cache is bypassed.

This situation should not matter for official releases as they will use proper tagged versions of all packages.

#### Checklist

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -95,6 +95,10 @@ tasks:
       - cp -n .env.example .env || true
       # Build site.
       - task dev:cli -- composer install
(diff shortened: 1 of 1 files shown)


spaceo (approved):
LGTM 👍

cableman (approved):

PR #192, opened 2022-12-12
Quiet errors when fetching ports for CI services during development

kasperg:
#### Description

Not all Docker services are available at all times when running Task. This causes the output to contain numerous line in the format

no container found for varnish_1

To avoid this we can pipe error output from docker compose to /dev/null. This silences the errors but note that it could also make related errors harder to find in the future.

I do not know if this will work on Windows.

#### Checklist


spaceo (approved):
LGTM 👍

kasperg:
This is nice but it is blocking a PR environment. I will close this - for now at least.

jeppekroghitk (approved):

PR #193, opened 2022-12-12
Feature/loan list v2

sinejespersen:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-459

#### Description

FBS module
Publizon module
Loan list

#### Screenshot of the result

image

image

image

#### Checklist


docker-compose.yml
```diff
@@ -72,8 +72,6 @@ services:
       lagoon.type: cli-persistent
       lagoon.persistent.name: nginx # mount the persistent storage of nginx into this container
       lagoon.persistent: /app/web/sites/default/files/ # location where the persistent storage should be mounted
-    environment:
-      << : *default-environment # loads the defined environment variables from the top
```
cableman, line 76: Why this change in the docker setup... don't think it should be altered.

web/modules/custom/dpl_loans/dpl_loans.info.yml
```diff
@@ -0,0 +1,10 @@
+name: DPL loans
+description: Handle the user profile loan pages
+package: DPL
+dependencies:
+  - drupal:block
+  - dpl_user_profile:dpl_user_profile
+  - dpl_fbs:dpl_fbs
+  - dpl_publizon:dpl_publizon
```
cableman, line 8: Should this have a dependency to `dpl_react`?

web/modules/custom/dpl_loans/dpl_loans.module
```diff
@@ -0,0 +1,9 @@
+<?php
+
+/**
+ * @file
+ * Dpl_loans drupal module file.
+ *
+ * Demonstrates how to use the dpl_react module
+ * and render components.
+ */
```
cableman, line 9: Maybe update this block with correct comment :-)

web/modules/custom/dpl_loans/src/Form/LoanListSettingsForm.php
```diff
@@ -0,0 +1,107 @@
...
+    if (!filter_var($feesUrl, FILTER_VALIDATE_URL)) {
+      $form_state->setErrorByName('fees_page_url', $this->t('The url "%url" is not a valid URL.', ['%url' => $feesUrl]));
+    }
+
+    $materialUrl = $form_state->getValue('material_overdue_url');
+    if (!filter_var($materialUrl, FILTER_VALIDATE_URL)) {
+      $form_state->setErrorByName('material_overdue_url', $this->t('The url "%url" is not a valid URL.', ['%url' => $materialUrl]));
+    }
+
+    $pageSizeMobile = intval($form_state->getValue('page_size_mobile'));
```
cableman, line 82: `intval` might not be the best way to do this, as intval will try to convert the value to an integer and hence you are not validating the actually value that gets stored.

See https://www.php.net/manual/en/function.intval.php 

A better solution would be:
```
$pageSizeMobile =  $form_state->getValue('page_size_mobile')
if (is_interger($pageSizeMobile) && $pageSizeMobile <= 0`) { ... } 
```
See https://www.php.net/manual/en/function.is-integer.php

web/modules/custom/dpl_loans/src/Form/LoanListSettingsForm.php
```diff
@@ -0,0 +1,107 @@
...
+    if ($pageSizeMobile <= 0) {
+      $form_state->setErrorByName('page_size_mobile', $this->t('Page size mobile has to be a positive integer'));
+    }
+
+    $pageSizeDesktop = intval($form_state->getValue('page_size_desktop'));
```
cableman, line 87: Same here with intval

cableman (changes requested):

docker-compose.yml
sinejespersen: Did not notice when i reviewed the files, I think this is a leftover from Ture. I'll revert

config/sync/config_ignore.settings.yml
```diff
@@ -97,5 +97,6 @@ ignored_config_entities:
   - ~user.settings
   - ~varnish_purger.settings.65fc931232
   - ~varnish_purger.settings.30cd45a1b1
+  - loan-list.settings
```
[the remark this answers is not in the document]
sinejespersen: Should it start with tilde?

web/modules/custom/dpl_loans/dpl_loans.info.yml
sinejespersen: I dont know, should we?

web/modules/custom/dpl_loans/dpl_loans.module
sinejespersen: good catch, will do

web/modules/custom/dpl_loans/src/Form/LoanListSettingsForm.php
sinejespersen: Thanks for this feedback, I'll implement

web/modules/custom/dpl_loans/src/Form/LoanListSettingsForm.php
sinejespersen: -||- ^

docker-compose.yml
sinejespersen: Checked it out, it seems "environment" is repeated 
https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/docker-compose.yml#L66-L76

web/modules/custom/dpl_loans/dpl_loans.info.yml
sinejespersen: I've created a dependency on dpl_react, that must be the way to go

PR #201, opened 2022-12-19
Lighthouse tests using wiremock

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-351

#### Description

- use same mappings as the pa11y mappings
- refactor functionality to reside in wiremock dir because it is now used by both Pa11y and Lighthouse
- add same urls to Ligthouse config as the ones in the Pa11y configuration

#### Checklist


Taskfile.yml
```diff
@@ -220,32 +220,29 @@ tasks:
...
-      - docker compose {{ .DOCKER_COMPOSE_FILES_CI }} run --rm node-chrome yarn install --frozen-lockfile
-      - docker compose {{ .DOCKER_COMPOSE_FILES_CI }} run --rm -e HTTP_PROXY={{ .WIREMOCK_HOST_DOCKER }} node-chrome yarn pa11y:create-mappings
-
   ci:lighthouse:
     desc: Run Lighthouse to test performance
+    deps: [ci:prepare]
+    cmds:
+      - docker compose {{ .DOCKER_COMPOSE_FILES_CI }} run --rm -e HTTP_PROXY={{ .WIREMOCK_HOST_DOCKER }} node-chrome npx lhci autorun
+
+  ci:prepare:
```
kasperg, line 242: Consider renaming this task and provide a description.

`ci:prepare` is very generic and sounds like something we would setup for all `ci:*` tasks, but we actually only do so for Pa11y and Lighthouse. It is (currently) not something we use with e.g. Cypress.

kasperg (changes requested):
This PR deletes the `pa11y/screenshots/.gitkeep` file. That does not seem intentional.
As far as I can tell this [stops the Pa11y job from generating and uploading screenshots as artifacts](https://github.com/danskernesdigitalebibliotek/dpl-cms/actions/runs/3733308186/jobs/6333932024).

Taskfile.yml
spaceo: Yeah. Naming is hard. I choose the generic version because I couldn't come up with a better alternative.

spaceo: Found  a better name - I guess :)

ci:wiremock:create-mappings

kasperg (approved):
:+1: LGTM

pa11y/.gitignore
```diff
@@ -1,2 +1 @@
-build
 screenshots
```
sinejespersen, line 1: why this?

sinejespersen (approved):

pa11y/.gitignore
spaceo: Good question.
It is because the only thing that the P11y directory is "managing"/containing is the screenshots directory. When the Pa11y cmd is executed it will generate screenshots for the configured URL's - and they will end up in /pa11y/screenshots

PR #203, opened 2022-12-20
Update queue text to include placeholder for number

kasperbirch1:
#### Update queue text to include placeholder for number

Include a placeholder for the variable number. This placeholder will be replaced with the actual value of the number variable at runtime.

#### Checklist


kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #212, opened 2023-01-10
FBI API update

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DSC-16

#### Description

Update FBI client and fixtures with code and data corresponding to the to JED 0.8 update.

This brings data in FBI fixtures related to works and materials in line with changes in the update.

This is a followup on the change in danskernesdigitalebibliotek/dpl-react#302. CI results of updating the client alone in [4aa12fb](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/212/commits/4aa12fbba09ba1a1dec9540827a74d910d2db09a) shows that the current fixtures in our test cases will work with updated client code but we update the fixtures for good measure anyway.

The code also contains a fix to a potential bug found in the `ci:wiremock:browser` task which created a data directory in my workspace.

Monosnap tmp 2023-01-10 06-05-32

#### Checklist

Checks for the latest commit [0ada7b3](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/212/commits/0ada7b3e3162cfe1af633a93eb290ae6b00da04e) is marked with a red cross but as far as I can tell all checks pass.

dpl-cms 2023-01-10 09-08-53


composer.json
```diff
@@ -17,10 +17,10 @@
             "type": "package",
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-react",
-                "version": "0.3.0-rc10",
+                "version": "0.3.0-rc11",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/release-develop/dist.zip",
+                    "url": "https://github.com/reload/dpl-react/releases/download/release-fix%2Ffbi-update-take-2/dist.zip",
```
kasperg, line 23: This change should be removed from the PR. The client will be automatically updated when danskernesdigitalebibliotek/dpl-react#302 lands.

spaceo (approved):
LGTM 👍

wiremock/src/goggles.ts
```diff
@@ -42,7 +42,7 @@ launch({
     `--proxy-server=${httpProxy}`,
     `--proxy-bypass-list=${httpProxyBypassList}`,
     '--args',
-    '--user-data-dir="/tmp/chrome_dev_test"',
+    '--user-data-dir=/tmp/chrome_dev_test',
```
kasperg, line 45: Another option would be to put the user data dir inside `./.cache` and thus contain it within the project.

sinejespersen (approved):

PR #217, opened 2023-01-16
End2end test - search result

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-308

#### Description
This PR adds search result page to the user journey cypress end-to-end test. The test checks whether data is rendered and whether clicking on one of the results ends up in a redirect to the correct material page. 

#### Screenshot of the result
n/a

#### Checklist

Includes #213 due to not being able to run wiremock locally without this change + #216

### DIFF
--- a/cypress/e2e/user-journey.cy.ts
+++ b/cypress/e2e/user-journey.cy.ts
@@ -1,5 +1,6 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
+/* eslint-disable @typescript-eslint/no-var-requires */
 const autosuggestData = require("../../wiremock/src/mappings/search/data/fbi/autosugggest.json");
+const searchResultData = require("../../wiremock/src/mappings/search/data/fbi/searchWithPagination.json");
 
 describe("User journey", () => {
   it("Shows search suggestions & redirects to search result page", () => {
@@ -28,6 +29,32 @@ describe("User journey", () => {
       .click()
       .url()
       .should("include", "search?q=Harry%2520Potter");
+  });
+
+  it("Shows search results & redirects to material page", () => {
+    cy.createMapping({
+      request: {
+        method: "POST",
+        url: "/opac/graphql",
+      },
+      response: {
+        headers: {
+          "Content-Type": "application/json",
+        },
+        status: 200,
+        jsonBody: searchResultData,
+      },
+    });
+
+    cy.visit("/search?q=Harry%2520Potter")
+      .get('[data-cy="search-result-title"]')
+      .should("contain", "Showing results for “Harry Potter” (109)")
+      .get('[data-cy="search-result-list"]')
+      .children()
+      .eq(0)
+      .click()
+      .url()
+      .should("include", "work/work-of:870970-basis:25245784");
   });
 
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+describe("User journey", () => {
+  it("Shows search suggestions & redirects to search result page", () => {
+    cy.createMapping({
+      request: {
+        method: "POST",
+        url: "/opac/graphql",
+      },
+      response: {
+        headers: {
+          "Content-Type": "application/json",
```
spaceo, line 15: Please do the matching as suggested [here](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/216/files#r1083912879).

cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+      .get(".header__menu-search-input")
+      .focus()
+      .type("harry")
+      .get(".autosuggest")
+      .should("be.visible")
+      .get(".autosuggest__text")
+      .eq(0)
+      .click()
+      .url()
+      .should("include", "search?q=Harry%2520Potter");
```
spaceo, line 32: Please make use of data-cy attribute and in general follow the pattern we use in dpl-react

cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+        url: "/opac/graphql",
+      },
+      response: {
+        headers: {
+          "Content-Type": "application/json",
+        },
+        status: 200,
+        jsonBody: searchResultData,
+      },
+    });
```
spaceo, line 48: Please do the matching as suggested [here](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/216/files#r1083912879).

cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+    cy.visit("/search?q=Harry%2520Potter")
+      .get('[data-cy="search-result-title"]')
+      .should("contain", "Showing results for “Harry Potter” (109)")
+      .get('[data-cy="search-result-list"]')
+      .children()
+      .eq(0)
+      .click()
+      .url()
+      .should("include", "work/work-of:870970-basis:25245784");
+  });
```
spaceo, line 72: I think also you can make use of the getBySel cmd here

cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+        jsonBody: availabilityLabelsData,
+      },
+    });
+
+    cy.visit("/search?q=Harry%2520Potter")
+      .get('[data-cy="search-result-title"]')
+      .should("contain", "Showing results for “Harry Potter” (109)")
+      .get('[data-cy="search-result-list"]')
+      .children()
+      .eq(0)
```
spaceo, line 68: Just info: It is also possible to use `first()`

cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+      .click()
+      .url()
+      .should("include", "work/work-of:870970-basis:25245784");
```
spaceo, line 71: I think it would be nice to first make an assertion of the workid of the the first search result. Then it makes sense that the first clicked search result results in a redirect to the same work id.

spaceo (changes requested):
Looks good. A few comments ;)

Adamik10:
Added as a part of #216

PR #218, opened 2023-01-17
End2end - material page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-309

#### Description
This PR adds material page to the user journey cypress end-to-end test. The test checks whether key data that a user would be focusing on during their user journey is rendered and whether clicking on the reservation button opens a modal.

#### Screenshot of the result
n/a

#### Checklist

Contains #216 + #217

### DIFF
--- a/cypress/e2e/user-journey.cy.ts
+++ b/cypress/e2e/user-journey.cy.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const autosuggestData = require("../../wiremock/src/mappings/search/data/fbi/autosugggest.json");
 const searchResultData = require("../../wiremock/src/mappings/search/data/fbi/searchWithPagination.json");
+const availabilityLabelsData = require("../../wiremock/src/mappings/work/data/fbs/availability.json");
 
 describe("User journey", () => {
   it("Shows search suggestions & redirects to search result page", () => {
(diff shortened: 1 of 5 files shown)


cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,147 @@
...
+const patronData = require("../../wiremock/src/mappings/work/data/fbs/patron.json");
+
+describe("User journey", () => {
+  it("Shows search suggestions & redirects to search result page", () => {
+    cy.createMapping({
+      request: {
+        method: "POST",
+        url: "/opac/graphql",
+      },
+      response: {
```
spaceo, line 16: There is one thing I would like to talk with you about looking at these mappings. I find it unnecessary to create the same mappings as we are using in the Pa11y and Lighthouse tests.

cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,147 @@
...
+  it("Shows material page & can open reservation modal", () => {
+    materialPageMocking();
+
+    cy.visit("/work/work-of:870970-basis:25245784")
+      .contains("Harry Potter og Fønixordenen")
+      .get('[data-cy="material-header-buttons-physical"]')
+      .should("contain", "Reserve bog")
+      .click()
+      .get('[data-cy="modal"]')
+      .should("contain", "Harry Potter og Fønixordenen");
```
spaceo, line 141: I think we should use getBySel here

spaceo (changes requested):
We need to talk about the mappings. There is one comment about the actual test I would like you to look at :)

Adamik10:
Became a part of #216

PR #220, opened 2023-01-18
End2end - reservation

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-310
https://reload.atlassian.net/browse/DDFSOEG-311

#### Description
This PR adds the last part of the reservation flow to the user journey cypress end-to-end test - reservation modal. The test checks whether key data that a user would be focusing on during their user journey is rendered and whether the reservation can be followed through to a confirmation screen.

#### Screenshot of the result
n/a

#### Checklist

Contains #216 + #217 + #218

### DIFF
--- a/cypress/e2e/user-journey.cy.ts
+++ b/cypress/e2e/user-journey.cy.ts
@@ -66,9 +66,9 @@ describe("User journey", () => {
     });
 
     cy.visit("/search?q=Harry%2520Potter")
-      .get('[data-cy="search-result-title"]')
+      .get("[data-cy='search-result-title']")
       .should("contain", "Showing results for “Harry Potter” (109)")
-      .get('[data-cy="search-result-list"]')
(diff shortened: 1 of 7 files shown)


cypress/e2e/user-journey.cy.ts
```diff
@@ -0,0 +1,198 @@
...
+      .get("[data-cy='modal']")
+      .should(
+        "contain",
+        "The material is available and is now reserved for you!"
+      )
+      .get("[data-cy='reservation-success-close-button']")
+      .click()
+      .get("[data-cy='modal']")
+      .should("not.exist");
+  });
```
spaceo, line 193: I have the same comments as in previous material page PR

spaceo (changes requested):
I have the same comments as in previous material page PR

Adamik10:
Became a part of #216

PR #224, opened 2023-01-25
Add search header app SSR

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-391

#### Description

This builds on top of [the PR that introduces SSR rendering of apps](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/222), starting with the search result app.
Because of that the PR mentioned should be merged into develop first before adding this.

**What is being done here?**
The search header SSR is added which improves the Lighthouse score with 8-9 points by lowering the total blocking time.

#### Checklist


kasperg (approved):
I have only reviewed [2d43b02](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/224/commits/2d43b0262e47c47a0c6ff873eef36647e2fe1460) here but things look good.
I wonder what makes the difference in this case.

Remember to at least fix the accessibility issue.

PR #230, opened 2023-02-07
Performance optimisations

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-400

#### Description

This PR solves a wide range of performance issues. Both by updated dependencies: dpl-react and dpl-design-system.
But also by changing the html rendered by dpl-cms which includes:
* Server side rendered html for apps for loading state
* Pre-connect tags for cover urls
* Pre-load tags for fonts
* Inline css by using the inline_all_css module
* Activate all 3 Lighthouse tests (front, search, material)
* Lazy load svg icons

#### Checklist

Both dpl-react and dpl-design-system PR's are involved in this PR. When they have been merged the composer file will be updated
I closed [another PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/224) since the changes are included in here as well.


web/themes/custom/novel/templates/layout/html.html.twig
```diff
@@ -29,12 +29,20 @@
...
+<html{{html_attributes}}>
+	<head>
+		<head-placeholder token="{{ placeholder_token }}">
+			<title>{{ head_title|safe_join(' | ') }}</title>
+			<link rel="preload" fetchpriority="low" href="/themes/custom/novel/assets/dpl-design-system/fonts/gochi-hand-v19-latin-regular.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" fetchpriority="low" href="/themes/custom/novel/assets/dpl-design-system/fonts/lora-v26-latin-italic.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/lora-v26-latin-regular.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/noto-sans-jp-v42-latin-500.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/noto-sans-jp-v42-latin-700.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/noto-sans-jp-v42-latin-regular.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
```
kasperg, line 41: I think this warrants some [inline commenting](https://twig.symfony.com/doc/3.x/templates.html#comments) about why this is necessary.

1. We preload fonts to improve performance with regard to first contentful paint/largest contentful paint
2. We inline *all* CSS to improve performance with regard to first contentful paint/largest contentful paint. We choose to inline all CSS because the alternative, inlining the critical CSS, is more complex to setup and maintain. We do not serve so much CSS at the moment that inling everything should be a problem.
3. We have to preload all fonts because we inline all CSS. Inlining all CSS does support relative urls to font files by design. 
4. We mark fonts which are not likely to be used in the FCP/LCP with a low fetch priority. It would be preferable not to preload them at all but that is not possible due to 2.
5. We decide not to preload `*.woff` fonts even though they are included in the design system and in our CSS to avoid preloading too many files. All supported browsers as defined by our code guidelines support `woff2` anyway.

Phew 😅 . Some of this might even constitute an ADR.

web/themes/custom/novel/templates/layout/html.html.twig
```diff
@@ -29,12 +29,20 @@
...
+			<link rel="preconnect" href="http://cover.dandigbib.org"/>
+			<link rel="preconnect" href="https://res.cloudinary.com"/>
```
kasperg, line 43: Consider how this is necessary.

Preconnecting to two of the external APIs that we use is helpful but what about FBI, FBS, Publizon and Material List? Is there a reason they are not included?

One simple approach would be to hardcode them all and hope that others will remember to do the same. Another approach would be to let some of the module(s) which define these urls and add them in a `hook_preprocess_page` implementation.

[We could also move the information to a Link HTTP header](https://web.dev/preconnect-and-dns-prefetch/) and avoid "polluting" the DOM.

web/themes/custom/novel/novel.theme
```diff
@@ -35,6 +38,25 @@ function novel_theme(
...
+ * Implements hook_preprocess_HOOK().
+ *
+ * @param mixed[] $variables
+ *   The variables for the theme hook.
+ */
+function novel_preprocess_dpl_react_app__search_header(array &$variables): void {
+  $search_svg = file_get_contents(sprintf('%s/%s',
+    \Drupal::theme()->getActiveTheme()->getPath(),
+    '/assets/dpl-design-system/icons/collection/Search.svg'
+  ));
```
kasperg, line 51: If we end of doing more of this stuff it would be nice with Twig functions to help inline a file.

config/sync/config_ignore.settings.yml
```diff
@@ -31,6 +31,7 @@ ignored_config_entities:
   - ~file.settings
   - ~filter.format.plain_text
   - ~filter.settings
+  - ~inline_all_css.settings
```
kasperg, line 34: :+1: for remembering this :smile.

kasperg (approved):
:+1: Excellent. The implementation here is good and the results speak for themselves.

The primary this missing is to get some of your decisions regarding handling of fonts and CSS out of your head and into the project. Some of this might be included in code comments but given the complexity of this I think we need a more prominent position.

web/themes/custom/novel/templates/layout/html.html.twig
kasperg: It seems as if [Drupal has a neat way of attaching `<head>` content *and* HTTP headers for render arrays](https://www.drupal.org/node/2160069).

web/themes/custom/novel/templates/layout/header.html.twig
```diff
@@ -2,7 +2,7 @@
...
...UGisao2zO08ByaoWn0soFlDUtiIlrguersYPNV6vwgTWi4Y8uhvszf7Skfq+6bM84dLrj1gofxDw2edW73/ILhG0N3r25FE2+jXGQR46IGXvZF911Hp3s+gJjxou3DZrl/1aKO0WGkc4uHa9uCvulT4VjAq9XTLhYiPXN+0Cg/Q5F2wrghBs9Y6rZZy7wp3tcjhOu30vQ7aFBq+06MAdoCVJxzIFuPzZl2b8t4i1wPi4/K908McE+4va0ucRm22EK0CRLYtkzvkKf7zC/EowF25gdD3+ebpXDud10aBvuZ+27xYltI6Vrt5s6mHbWHBrsHPAVqQhznWyOqzO0NWLLBkSYp/zKNoi3PjWpb7X47YhkZrcr1QYt5tr1tc529bHPtlmXLjBFWTiUy9g6t8tFoWzVBLF3tRMKbSb/6DLWn2pdlvNNTpYWwqV7omlpEaVTXz3dllA4vdGa8SpraVtIJZ2W2HKwNYsl6LWbVkBWbT0bEq4UzMvpcmP1BYZTqBlBG2FDpmOupz2t83FdDbVZ/HrAfdelGhEhpxLrERVpO40AqKY5t3EUx2C+qDh0uDa2EWsa0y/yzc/6FrL7vkvDrxzF03MINfl6+tfHHyo8ERwimYQQ530MhuRBdne7FNVIhWIDp/GGpa5cOPTT51A/TTatrrmbldkjjLYpGDuEnqSPFxBzkNmL1QtMX+SMb1ybqkRrcC8oXzXgHo7UT7wI7QJGZrk9Ai6yatdJjXEmivYpUldE0CeWUxgSd3d52unVvIt0oc+12bhh307tdlMZBEVe91DfLuse4nNqxFNUfN6GcRcybIxh7ydA9VybUFQ3i/eAa71JfRsY2riAn+28ZY740oX8DFrYKoeP5lP9VmvN0uqaDy6T4a91LrpqEsUFjFR8TI0kVYrpuxen6/uTx1ty6Qde7jsK1TbCvZRyXqL5Sf2aSFCqgNCttzTOrqDPupD8m/z6lnL+++uYbSb+6jFZSOapNWcpqYuy/aLlySXfLTJ1UFrr9VdKPyGfEzyKNr4AmxqH0tZTvTeTaawWxS18/Wul3rCFXriTdbbSG88YctcNcbtigQziXH2lZ6oQ7VDamQXlhlBb5UY0TlcWoLC+V1Xk8D+y6bV6zxguxmIXz5iWdTPl/+jv2xktbiOwAAAAASUVORK5CYII=" alt="Logo image of libary"/>
```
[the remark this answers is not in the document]
spaceo: It's a remain from an experiment of lazy loading all svg whatsoever.
I'll delete them from the header

web/themes/custom/novel/templates/layout/html.html.twig
spaceo: It is to be able to serve images as fast as possible. You are right that it could be used for other services, but then again then we end up with having everything as high priority.
I'll leave a note about why I think it is necessary. But I do not think we should invest more time in this matter. We have already spent a lot of time in getting here. :)

web/themes/custom/novel/templates/layout/html.html.twig
spaceo: Great feedback. I'll write a comment about. And you are right about the ADR. Wrote a TODO about that.

web/themes/custom/novel/novel.theme
spaceo: yes

config/sync/config_ignore.settings.yml
spaceo: ;)

web/themes/custom/novel/novel.theme
```diff
@@ -35,6 +38,25 @@ function novel_theme(
...
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ *
+ * @param mixed[] $variables
+ *   The variables for the theme hook.
+ */
+function novel_preprocess_dpl_react_app__search_header(array &$variables): void {
+  $search_svg = file_get_contents(sprintf('%s/%s',
```
cableman, line 48: So the plan is to inline all svg icons etc. or is this some skeleton images?

cableman: When using `file_get_contents` you should check that it actually gets data and not is `false`.

cableman: How much is the performance gain by doing this with svg icons? And why is it only this svg icon?

web/themes/custom/novel/templates/layout/html.html.twig
```diff
@@ -29,12 +29,20 @@
...
+			<title>{{ head_title|safe_join(' | ') }}</title>
+			<link rel="preload" fetchpriority="low" href="/themes/custom/novel/assets/dpl-design-system/fonts/gochi-hand-v19-latin-regular.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" fetchpriority="low" href="/themes/custom/novel/assets/dpl-design-system/fonts/lora-v26-latin-italic.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/lora-v26-latin-regular.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/noto-sans-jp-v42-latin-500.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/noto-sans-jp-v42-latin-700.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preload" href="/themes/custom/novel/assets/dpl-design-system/fonts/noto-sans-jp-v42-latin-regular.woff2" as="font" type="font/woff2" crossorigin="anonymous"/>
+			<link rel="preconnect" href="http://cover.dandigbib.org"/>
```
cableman, line 42: I really think that these `preconnect` and `preload` urls should be configurable in drupal backend as they may change or be updated (or more being added later on).

cableman (changes requested):
Some questions and why is `wiremock/src/mappings/search/data/fbi/searchWithPagination.json` which I assume is test data removed in this PR? What have I missed?

web/themes/custom/novel/novel.theme
spaceo: > So the plan is to inline all svg icons etc. or is this some skeleton image?

This is to be able to render the search header fast. It is the same principle of showing the rendering server side before javascript kicks in with it's own rendering. In this particular case we show the search icon (and the search field) before the dpl-react app comes and does the same.

spaceo: > When using `file_get_contents` you should check that it actually gets data and not returns `false`.

Good point. I'll look into that.

web/themes/custom/novel/templates/layout/html.html.twig
spaceo: I agree. And I think we should do it. But considering the amount of time already put into this I think we should write a TODO here and get it done in the future.

cableman: Then make it a big TODO :-D

spaceo: Like this?:
```
    _____                   _______                   _____                   _______                 
     /\    \                 /::\    \                 /\    \                 /::\    \                
    /::\    \               /::::\    \               /::\    \               /::::\    \               
    \:::\    \             /::::::\    \             /::::\    \             /::::::\    \              
     \:::\    \           /::::::::\    \           /::::::\    \           /::::::::\    \             
      \:::\    \         /:::/~~\:::\    \         /:::/\:::\    \         /:::/~~\:::\    \            
       \:::\    \       /:::/    \:::\    \       /:::/  \:::\    \       /:::/    \:::\    \           
       /::::\    \     /:::/    / \:::\    \     /:::/    \:::\    \     /:::/    / \:::\    \          
      /::::::\    \   /:::/____/   \:::\____\   /:::/    / \:::\    \   /:::/____/   \:::\____\         
     /:::/\:::\    \ |:::|    |     |:::|    | /:::/    /   \:::\ ___\ |:::|    |     |:::|    |        
    /:::/  \:::\____\|:::|____|     |:::|    |/:::/____/     \:::|    ||:::|____|     |:::|    |        
   /:::/    \::/    / \:::\    \   /:::/    / \:::\    \     /:::|____| \:::\    \   /:::/    /         
  /:::/    / \/____/   \:::\    \ /:::/    /   \:::\    \   /:::/    /   \:::\    \ /:::/    /          
 /:::/    /             \:::\    /:::/    /     \:::\    \ /:::/    /     \:::\    /:::/    /           
/:::/    /               \:::\__/:::/    /       \:::\    /:::/    /       \:::\__/:::/    /            
\::/    /                 \::::::::/    /         \:::\  /:::/    /         \::::::::/    /             
 \/____/                   \::::::/    /           \:::\/:::/    /           \::::::/    /              
                            \::::/    /             \::::::/    /             \::::/    /               
                             \::/____/               \::::/    /               \::/____/                
                              ~~                      \::/____/                 ~~                      
                                                       ~~

cableman: That's beautiful.... let's go with that...

cableman (approved):

web/themes/custom/novel/novel.theme
kasperg: Quick note: `thecodingmachine/safe` ensures that `file_get_contents` will throw an exception if loading fails - not return `false`.

PR #234, opened 2023-02-13
Change opac to next + Remove .test from publizon service

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-416
https://reload.atlassian.net/browse/DSC-34

#### This PR change following:
- Change profile on `fbi-api` from 'opac' to 'next'
- Change `publizon` service to production

#### Checklist


kasperg (approved):
:+1: LGTM

kasperg (changes requested):
There are other places that reference the `opac` part of the url e.g. the Wiremock mappings.

This is probably also why all the tests that use the mappings fail.

Please correct this.

cableman (approved):
Note: have you test that http://pubhub-openplatform.dbc.dk actually works. It has returned 500 on almost all request for a very long time now...

kasperbirch1:
> 

I tested it in 'dpl-react' and it works there.
image

PR #236, opened 2023-02-16
ITKDev - loans list

cableman:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-459

#### Description

Added module that handles loan lists.

#### Screenshot of the result

N/A

#### Checklist

Requires merge of https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/235 before merge and contains all changes from that PR.

* PHPStan errors is due to Stan not knowing about Drupal forms. 
* pa11y that fails is in the search code outside this PR.
* cypress tests that fails are ind the search pages and outside this PR.

### DIFF
--- a/web/modules/custom/dpl_loans/src/Controller/DplLoansController.php
+++ b/web/modules/custom/dpl_loans/src/Controller/DplLoansController.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Drupal\dpl_loans\Controller;
+
+use Drupal\Core\Block\BlockManagerInterface;
+use Drupal\Core\Controller\ControllerBase;
+use Drupal\Core\Render\RendererInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Render loan list react app.
+ */
+class DplLoansController extends ControllerBase {
+
+  /**
+   * DplLoansController constructor.
+   *
+   * @param \Drupal\Core\Block\BlockManagerInterface $blockManager
+   *   Drupal block manager.
+   * @param \Drupal\Core\Render\RendererInterface $renderer
+   *   Drupal renderer service.
+   */
+  public function __construct(
+    private BlockManagerInterface $blockManager,
+    private RendererInterface $renderer
+  ) {
+  }
+
+  /**
+   * {@inheritdoc}
+   *
+   * @param \Symfony\Component\DependencyInjection\ContainerInterface $container
+   *   The Drupal service container.
+   *
+   * @return static
+   */
+  public static function create(ContainerInterface $container) {
+    return new static(
+      $container->get('plugin.manager.block'),
+      $container->get('renderer'),
+    );
+  }
+
+  /**
+   * Demo react rendering.
+   *
+   * @return mixed[]
+   *   Render array.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
+   */
+  public function list(): array {
+    $config = [];
+
+    /** @var \Drupal\dpl_loans\Plugin\Block\LoanListBlock $plugin_block */
+    $plugin_block = $this->blockManager->createInstance('dpl_loans_list_block', $config);
+
+    // Some blocks might implement access check.
+    $access_result = $plugin_block->access($this->currentUser());
+
+    // Return empty render array if user doesn't have access.
+    // $access_result can be boolean or an AccessResult class.
+    if (is_object($access_result) && $access_result->isForbidden() || is_bool($access_result) && !$access_result) {
+      // You might need to add some cache tags/contexts.
+      return [];
+    }
+
+    // Add the cache tags/contexts.
+    $render = $plugin_block->build();
+    $this->renderer->addCacheableDependency($render, $plugin_block);
+
+    return $render;
+  }
+
+}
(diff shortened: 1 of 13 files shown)


jeppekroghitk (approved):

kasperg (changes requested):
All comments from #239 apply here as well.

Given the similarity of the implementation and considerations that future implementation may look just the same seems to warrant some code of abstraction. I do not have any concrete abstractions at the moment.

cableman:
Fixed related errors found in https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/239 and the PHPStan errors have been fixed there (and will just create merge errors if repeated here)

jeppekroghitk (approved):

web/modules/custom/dpl_loans/src/Controller/DplLoansController.php
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
+   */
+  public function list(): array {
+    $config = [];
+
+    /** @var \Drupal\dpl_loans\Plugin\Block\LoanListBlock $plugin_block */
+    $plugin_block = $this->blockManager->createInstance('dpl_loans_list_block', $config);
+
+    // Some blocks might implement access check.
```
spaceo, line 59: This is really minor. But why do you write "blocks" in plural when we know there is only one block (`dpl_loans_list_block`) that is being used. Which is used (`dpl_loans_list_block::build()`) in order to render the list in yhis method. If I have understood it correctly :)

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * {@inheritDoc}
+   *
+   * @return mixed[]
+   *   The app render array.
+   */
+  public function build() {
+    $loanListSettings = $this->configFactory->get('dpl_loan_list.settings');
+    $context = ['context' => 'Loan list'];
+    $contextAria = ['context' => 'Loan list (Aria)'];
```
spaceo, line 78: Great with multiple contexts! I think we have to few in the material app (namely one).

One thing that I would like to mention is that, with the translation server task we are looking into using a tool for scanning the codebase for translation functions. It is probably using some kind of build in reflection functionality in PHP. But one thing I have noticed is that:
* You have to write literal strings for the source string (the first argument for `t()`). That's not something new, but I have seen places in our common codebase where `t()` is being used when errors are thrown.
*Context should be written literally everytime a translation function is being used. So it is not possible to reuse `$context` as it is being done here.

This is just to mention it. And I guess there is quite some refactoring to be done if we stick with the current idea of how we should scan the code. Alternatively we could put the whole data prop array (at least the text part) in .inc files to clean it up a bit.

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -0,0 +1,165 @@
...
+   * Drupal config factory.
+   *
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  private ConfigFactoryInterface $configFactory;
+
+  /**
+   * LoanListBlock constructor.
+   *
+   * @param array $configuration
```
spaceo, line 31: You need to specify `mixed[]` as type here in order to make PhpStan quiet.
You could argue that it is not the right way to go.
But I think that is a separate discussion and my main concern is to eliminate PHPstan errors.

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -0,0 +1,165 @@
...
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, ConfigFactoryInterface $configFactory) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->configuration = $configuration;
+    $this->configFactory = $configFactory;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
```
spaceo, line 49: Same as I wrote in previous comment.

spaceo (changes requested):
Looks good! A few minors though...

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
cableman: Why scan the code base... will Drupal not build an database table over all translatable strings that could be "dump" and used. Repeating the context string (in some cases 100 of times in the code base) will only lead to C&P errors and typos.

I think it would be much better idea to reuse the Drupal parts that already collect translations from the code base.

spaceo: It is because we want to be able to register all strings without rendering the pages/apps.
It is the same principle that the translation server of drupal.org, https://localize.drupal.org, is using.

spaceo (approved):
LGTM 👍

cableman:
Resolved merge conflict in composer lock

PR #249, opened 2023-02-22
Ignore iterable type PHPStan errors for Drupal Form API

kasperg:
#### Description

Ignore iterable type PHPStan errors for Drupal Form API.

Drupal Form API makes extensive use for arrays which we cannot provide more detailed typing of.

#### Checklist

n/a


kasperg:
Deployment fails because no PR environment is created. Perhaps it is because there is no actual code change in this PR. I do not know if it makes sense to merge regardless of deployment failing or not.

spaceo (approved):

jeppekroghitk (approved):

PR #251, opened 2023-02-22
Remove digital archive service from CMS

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DSC-32

#### Description

This PR removes the Digital Archieve Service module from the CMS. This is the result of the responsibility of integrating with the Danish Royal Library Elba webservices is moved from the CMS to the FBI platform.

This allows React components expand the usage of FBI instead of adding business logic and integrations in the CMS. Consequently this PR is related to danskernesdigitalebibliotek/dpl-react#355.

This PR removes all code and configuration related to the module.

During the process I identified a problem where disabled REST resources where exposed in the generated OpenAPI specification. For good measure I included a patch to address this. While not necessarily relevant here because we delete the configuration entirely the handling of disabled REST resources by the OpenAPI REST module can also cause [other problems](https://www.drupal.org/project/openapi_rest/issues/3277858). I have included the patch for good measure. 

#### Checklist

PHPStan analysis fails due to errors in unrelated code. It is addressed in #249.

I am not sure why the user journey test fails. It is likely due to unrelated errors in this PR.


spaceo (approved):
Approved.
The PHPStan errors have been dealt with in another PR, right?

cableman (approved):

PR #252, opened 2023-02-22
Configure all REST endpoints to use the dpl_login_user_tokens provider

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-376

#### Description

The frontend will pass a bearer token for the library or user by default. The UserTokenAuthProvider applies to such requests but for authentication to work it also has to be a valid authentication provider for the route.

Consequently we configure the campaign and proxy endpoints to use this instead.

In practice both endpoints are accessible by anonymous users but for some reason the authentication providers must match anyhow. If not the resources will return 403 and campaigns will not be displayed.

Update configuration accordingly.

#### Checklist

The user journey tests fails errors unrelated to this PR.

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -5,8 +5,8 @@ module:
   config_filter: 0
   config_ignore: 0
   dpl_campaign: 0
-  dpl_fbs: 0
   dpl_das: 0
+  dpl_fbs: 0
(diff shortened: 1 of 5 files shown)


spaceo (approved):
LGTM 👍

jeppekroghitk (approved):

PR #260, opened 2023-03-02
skip the `Can open reservation modal & reserve a material` test.

kasperbirch1:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-467

#### This PR requests that we skip the `Can open reservation modal & reserve a material` test.

As it stands, the current test seems to be experiencing some issues when it comes to clicking the "Reserve" button and invoking the reservation API. We are currently unsure of the root cause of this issue.

Therefore, I propose that we skip this test until we are able to resolve the issue. 

#### Checklist


spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #264, opened 2023-03-03
Do not use self closing in material skeleton

spaceo:
#### Description

For some reason we need separate closing html elements in order to render the empty skeleton screen elements right-

#### Checklist


Adamik10 (approved):
Very nice, teehee.
In regards to the failing tests, you can refer to Kasper Birk's fixes in PR #260

PR #268, opened 2023-03-07
Try to prevent NoCorrespondingEntityClassException for VarnishPurgerSettings 

kasperg:
The module name as defined by the .info.yml file is varnish_purger but the package name on drupal.org is varnish_purge. This causes the module to be downloaded to `web/modules/contrib/varnish_purge by default.

While the module seems to work as is we have been seeing errors as shown below when trying to install the site. This occurs occasionally both on CI and sometimes locally. We have not been able to recreate the error consistently.

VarnishPurgerSettings is a plugin and the error seems to occur because it has not been detected properly. According to the documentation for namespace resolution the base namespace for Drupal\varnish_purger\ should be modules/varnish_purger/src/. 
https://www.drupal.org/docs/develop/standards/php/psr-4-namespaces-and-autoloading-in-drupal-8

With this change the directory for the namespace Drupal\varnish_purger\ becomes modules/contrib/varnish_purger/src/. This brings it in accordance with the documentation. The contrib directory should not make a difference.

Update composer.lock after the change.

[...]
Notice: >  [notice] Checked da translation for schemata. Notice: >  [notice] Checked da translation for varnish_purge. Notice: >  [notice] Checked da translation for dpl_cms. Notice: >  [notice] Message: Checked available interface translation updates for 16 projects.
> 
> PHP Fatal error:  Uncaught Drupal\Core\Entity\Exception\NoCorrespondingEntityClassException: The Drupal\varnish_purger\Entity\VarnishPurgerSettings class does not correspond to an entity type. in /app/web/core/lib/Drupal/Core/Entity/EntityTypeRepository.php:98
> Stack trace:
> #0 /app/web/core/lib/Drupal/Core/Entity/EntityBase.php(487): Drupal\Core\Entity\EntityTypeRepository->getEntityTypeFromClass('Drupal\\varnish_...')

#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -122,6 +122,7 @@
             "web/themes/custom/novel/assets/{$name}": ["danskernesdigitalebibliotek/dpl-design-system"],
             "web/core": ["type:drupal-core"],
             "web/libraries/{$name}": ["type:drupal-library"],
+            "web/modules/contrib/varnish_purger": ["drupal/varnish_purge"],
             "web/modules/contrib/{$name}": ["type:drupal-module"],
             "web/profiles/contrib/{$name}": ["type:drupal-profile"],
             "web/themes/contrib/{$name}": ["type:drupal-theme"],
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


kasperg:
https://github.com/danskernesdigitalebibliotek/dpl-cms/actions/runs/4353356391/jobs/7608119298 shows that this approach does not work.

PR #289, opened 2023-03-28
Allow github bot to autocommit translation changes

spaceo:
#### Description

This PR adds a PAT for the dpl-platform-bot to allow it to make commits on the develop branch.
We ran into the problem because the develop branch is protected. 
This should resolve the [issue](https://github.com/stefanzweifel/git-auto-commit-action#push-to-protected-branches).

#### Checklist


kasperg (approved):
:+1: LGTM

PR #290, opened 2023-03-29
Instant Loan module

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DSC-12

#### Description

Add instant loan module

- Configuration form added
- Specific permission for editing instant loan configuration added
- The configuration is added to the data prop of the material app
    
#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_instant_loan/dpl_instant_loan.links.menu.yml
+++ b/web/modules/custom/dpl_instant_loan/dpl_instant_loan.links.menu.yml
@@ -2,4 +2,4 @@ dpl_instant_loan.settings_form:
   title: "Instant Loan Settings"
   route_name: dpl_instant_loan.settings
   description: "Configure promotion of materials available for instant loans."
-  parent: system.admin_config_services
+  parent: dpl_library_agency.settings
--- a/web/modules/custom/dpl_instant_loan/dpl_instant_loan.links.menu.yml
+++ b/web/modules/custom/dpl_instant_loan/dpl_instant_loan.links.menu.yml
@@ -1,5 +1,5 @@
 dpl_instant_loan.settings_form:
   title: "Instant Loan Settings"
   route_name: dpl_instant_loan.settings
-  description: "Change settings of the instant loan feature"
+  description: "Configure promotion of materials available for instant loans."
   parent: system.admin_config_services
--- a/web/modules/custom/dpl_instant_loan/dpl_instant_loan.links.menu.yml
+++ b/web/modules/custom/dpl_instant_loan/dpl_instant_loan.links.menu.yml
@@ -1,0 +1,5 @@
+dpl_instant_loan.settings_form:
+  title: "Instant Loan Settings"
+  route_name: dpl_instant_loan.settings
+  description: "Change settings of the instant loan feature"
+  parent: system.admin_config_services
--- a/web/modules/custom/dpl_instant_loan/dpl_instant_loan.permissions.yml
+++ b/web/modules/custom/dpl_instant_loan/dpl_instant_loan.permissions.yml
@@ -1,0 +1,3 @@
+administer instant loan configuration:
+  title: "Administer instant loan configuration"
+  description: "Access for administering of the instant loan configuration."
--- a/web/modules/custom/dpl_instant_loan/dpl_instant_loan.routing.yml
+++ b/web/modules/custom/dpl_instant_loan/dpl_instant_loan.routing.yml
@@ -2,7 +2,7 @@ dpl_instant_loan.settings:
   path: "/admin/config/services/instant-loan"
   defaults:
     _form: '\Drupal\dpl_instant_loan\Form\DplInstantLoanSettingsForm'
-    _title: "Instant Loan Settings"
+    _title: "Instant loan settings"
   requirements:
     _permission: "administer instant loan configuration"
   options:
--- a/web/modules/custom/dpl_instant_loan/dpl_instant_loan.routing.yml
+++ b/web/modules/custom/dpl_instant_loan/dpl_instant_loan.routing.yml
@@ -1,5 +1,5 @@
 dpl_instant_loan.settings:
-  path: "/admin/config/services/instant-loan"
+  path: "/admin/config/dpl-library-agency/instant-loan"
   defaults:
     _form: '\Drupal\dpl_instant_loan\Form\DplInstantLoanSettingsForm'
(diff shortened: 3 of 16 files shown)


web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -427,4 +428,14 @@ public static function ensureUrlIsString(string|GeneratedUrl $url): string {
     return $url;
   }
 
+  /**
+   * Get the instant loan configuration.
+   *
+   * @return mixed[]
+   *   The instant loan configuration.
+   */
+  public static function getInstantLoanConfig() {
```
kasperg, line 437: Please add the return type to the method instead of in the Docblock.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -427,4 +428,14 @@ public static function ensureUrlIsString(string|GeneratedUrl $url): string {
...
+    return \Drupal::configFactory()->get('dpl_instant_loan.settings')->get() ?? [];
```
kasperg, line 438: The approach to using configuration here is different compared to the one we use for branches.

Other settings are wrapped in services and injected into the controller. The controller then again has methods to convert the settings to JSON as expected by the apps. The service also knows how to use the setting to control caches.

This setting is retrieved directly from a static method and then passed to the app. It does not deal with caching when the configuration changes.

Personally I appreciate the type handling for the approach used with branch settings I recognize that it may be over-complicated. [There are already other settings where it has not been implemented](https://github.com/reload/dpl-cms/blob/main/web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php#L210-L210), [Drupal configuration objects can also be used with cache handling on their own](https://github.com/drupal/drupal/blob/f965e40015127c57b48c853bd60bb81a7d5cd3bb/core/lib/Drupal/Core/Config/ConfigBase.php#L27) and [serialization to JSON is also handled elsewhere](https://github.com/reload/dpl-cms/blob/main/web/modules/custom/dpl_react/dpl_react.module#L129-L129).

With all this in mind I think there are at least some changes that should be made to the current implementation:

1. Use `$this->config('dpl_instant_loan.settings')` as provided by the controller class to retrieve the config in the controller instead of using `\Drupal::configFactory()`
2. Add the configuration object as a cachable dependency to the render array for the material app
3. Consider whether it is possible to document the shape of the instant loans config expected by the React application. I do not have any good suggestion. One option would be to add it to the `getInstantLoanConfig()` docblock but if we need a `Config` object for cache management then we will need to change the signature of this function anyway.

web/modules/custom/dpl_instant_loan/dpl_instant_loan.routing.yml
```diff
@@ -0,0 +1,9 @@
+dpl_instant_loan.settings:
+  path: "/admin/config/services/instant-loan"
+  defaults:
+    _form: '\Drupal\dpl_instant_loan\Form\DplInstantLoanSettingsForm'
+    _title: "Instant Loan Settings"
```
kasperg, line 5:
```suggestion
    _title: "Instant loan settings"
```

web/modules/custom/dpl_instant_loan/src/Form/DplInstantLoanSettingsForm.php
```diff
@@ -0,0 +1,84 @@
...
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $config = $this->config('dpl_instant_loan.settings');
+
+    $form['enabled'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Enabled', [], ['context' => 'dpl_instant_loan']),
+      '#description' => $this->t(
+        'Should the instant loan functionality be activated?',
```
kasperg, line 39:
```suggestion
        'Should materials available for instant loans be promoted to patrons?',
```

web/modules/custom/dpl_instant_loan/src/Form/DplInstantLoanSettingsForm.php
```diff
@@ -0,0 +1,84 @@
...
+        [],
+        ['context' => 'dpl_instant_loan']
+      ),
+      '#default_value' => $config->get('enabled'),
+    ];
+
+    $form['match_string'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Match String', [], ['context' => 'dpl_instant_loan']),
```
kasperg, line 48: Consider if we have a dependency between fields here.

How will the React app work if `match_string` and/or `threshold` are empty if the feature is enabled? Should they in this case be required?

kasperg (changes requested):
:+1:

A few changes are needed here in relation to how the configuration form is integrated into the admin interface and how configuration is passed to the app.

There were also a few places where the descriptions got a little too generic to my taste. Feel free to use my suggestions.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
spaceo: 👍

web/modules/custom/dpl_instant_loan/dpl_instant_loan.permissions.yml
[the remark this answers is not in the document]
spaceo: True 👍

web/modules/custom/dpl_instant_loan/src/Form/DplInstantLoanSettingsForm.php
spaceo: Added state handling in the form. 👍

web/modules/custom/dpl_instant_loan/dpl_instant_loan.links.menu.yml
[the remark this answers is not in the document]
spaceo: 👍

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
spaceo: Added e config service 👍

kasperbirch1 (approved):

cableman (approved):

kasperg (approved):

PR #293, opened 2023-03-29
Revert "Fix input passing to webhook action in translation workflow"

kasperg:
#### Description

This action does in fact use env instead of with for input.

This reverts commit 18d8429bf29a6df7871ed82c84218662ba57eb45.

#### Checklist


spaceo (approved):
LGTM 👍

PR #294, opened 2023-03-30
Add literal context to translation strings

spaceo:
#### Description

Adds literal context to translation strings to the ones that did not have it already.

#### Checklist


cableman:
It would be really smart if Drupal had had a default context with module name...

spaceo:
Thx for the feedback. You are more than welcome in streamlining all the contexts 😜

PR #296, opened 2023-04-13
Fixing user-journey reservation test

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-504

#### Description
Make sure event handler is present on reserve-btn by making sure that modal is fully mounted.
Inspired by [this article](https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/).

#### Checklist

There are several changes from latest demo branch and instant loan PR that need to me merged before this fully works

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -148,13 +148,14 @@ jobs:
         run: task ci:reset
       - name: Run Cypress
         run: task ci:cypress
-      - name: Archive vidoes
+      - name: Archive videoes
+        if: always()
         uses: actions/upload-artifact@v3
         with:
           name: cypress-videos
           path: cypress/videos
       - name: Archive screenshots
-        if: ${{ failure() }}
+        if: always()
         uses: actions/upload-artifact@v3
         with:
           name: cypress-screenshots
--- a/cypress/e2e/withMappings/user-journey.cy.ts
+++ b/cypress/e2e/withMappings/user-journey.cy.ts
@@ -52,6 +52,7 @@ describe("User journey", () => {
     // We have to wait for the modal to be fully rendered
     // and the event listeners to be attached.
     // Read more: https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/
+    // TODO: Consider using the pipe package in the future...
     // eslint-disable-next-line cypress/no-unnecessary-waiting
     cy.wait(2000);
     cy.getBySel("reservation-modal-submit-button").click();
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -148,13 +148,14 @@ jobs:
...
-      - name: Archive vidoes
+      - name: Archive videoes
+        if: always()
         uses: actions/upload-artifact@v3
         with:
           name: cypress-videos
           path: cypress/videos
       - name: Archive screenshots
-        if: ${{ failure() }}
+        if: always()
```
Adamik10, line 158: I'm just wondering - is there a reason we now always archive screenshots? Even if the test doesn't fail?

cypress/e2e/withMappings/user-journey.cy.ts
```diff
@@ -36,18 +35,25 @@ describe("User journey", () => {
...
-    cy.getBySel("modal")
+    // We have to wait for the modal to be fully rendered.
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(2000);
+    cy.getBySel("reservation-modal-parallel")
       .should("be.visible")
       .and("contain", "Harry Potter og Fønixordenen");
+    // We have to wait for the modal to be fully rendered
+    // and the event listeners to be attached.
+    // Read more: https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/
```
Adamik10, line 54: Very nice that you added this link! Should we consider using .pipe() like they do? I know it's an extra package, but we could just add it to devDependencies maybe?

docker-compose.yml
```diff
@@ -84,8 +82,8 @@ services:
         CLI_IMAGE: *cli-image # Inject the name of the cli image
     depends_on:
       - php
-    << : *default-volumes # loads the defined volumes from the top
-    << : *default-user # uses the defined user from top
+    # loads the defined volumes and user from the top
+    <<: [*default-volumes, *default-user]
```
Adamik10, line 86: I am unsure as to what these changes do, but if you're confident in them, I trust your judgement,

Adamik10 (approved):
Approved with some comments :)

.github/workflows/ci-tests.yml
spaceo: It is a bit vague for me now. But I believe I added it because otherwise we could not see the assets when test were failing.

cypress/e2e/withMappings/user-journey.cy.ts
spaceo: That could be an idea yes. I think I'll add a todo about for now, so we don't use more time on it.

docker-compose.yml
spaceo: Docker compose was failing in CI because the syntax has changed regarding merging defaults

PR #300, opened 2023-04-14
Fix docker compose validation error

spaceo:
Make shure we have unique yaml keys.
Inspired by this post:
https://github.com/docker/compose/issues/10411#issuecomment-1489657328

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,8 +72,6 @@ services:
       lagoon.type: cli-persistent
       lagoon.persistent.name: nginx # mount the persistent storage of nginx into this container
       lagoon.persistent: /app/web/sites/default/files/ # location where the persistent storage should be mounted
-    environment:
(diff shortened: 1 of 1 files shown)


kasperg (approved):
:+1: LGTM

I expect this is referred to errors seen in e.g. #299.

Since the problems originate in the setup provided by Lagoon you could consider also mentioning the problem and potentially providing a fix in there: https://github.com/lagoon-examples/drupal9-base/blob/main/docker-compose.yml.

In general I do not like problems such as this occurring unrelated to our code changes. I cannot see how we could do anything different here. As I understand it a fixed `ubuntu-22.04` runner would suffer from the same problem.

cableman (approved):

donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

PR #304, opened 2023-04-20
ITKDev - Patron redirect handling

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5372

#### Description

Redirect user to configured patron login method on configurable paths if user is not authenticated.

#### Screenshot of the result

N/A

#### Checklist

N/A


sinejespersen (approved):

spaceo (commented):
Would like to talk about this in person. Think it will be easier than messages back and forth.

PR #308, opened 2023-04-28
Use placeholder in search loading title

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-519

#### Description

Since dpl-react was changed to use query placeholder in the search title we need to do the same in dpl-cms.

Reference:
https://github.com/danskernesdigitalebibliotek/dpl-react/commit/eb14285a868acd27e457d5b89b08b5eb77a65b8e

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
+++ b/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
@@ -135,7 +135,7 @@ class DplReactAppsController extends ControllerBase {
       'results-text' => $this->t('results', [], ['context' => 'Search Result']),
       'show-more-text' => $this->t('Show more', [], ['context' => 'Search Result']),
       'show-results-text' => $this->t('Show results', [], ['context' => 'Search Result']),
-      'showing-results-for-text' => $this->t('Showing results for', [], ['context' => 'Search Result'])->__toString(),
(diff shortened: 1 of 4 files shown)


web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -135,7 +135,7 @@ public function search(): array {
       'results-text' => $this->t('results', [], ['context' => 'Search Result']),
       'show-more-text' => $this->t('Show more', [], ['context' => 'Search Result']),
       'show-results-text' => $this->t('Show results', [], ['context' => 'Search Result']),
-      'showing-results-for-text' => $this->t('Showing results for', [], ['context' => 'Search Result'])->__toString(),
+      'showing-results-for-text' => $this->t('Showing results for "@query"', [], ['context' => 'Search Result'])->__toString(),
```
kasperg, line 138: Consider if we really need to call `__toString()` here.

We do not do so for other texts here. I see that we need it for the handling in `novel_preprocess_dpl_react_app__search_result()`. An alternate approach might be to move all the handling there á la:

```php
$loading_text = $variables['data']['showing-results-for-text'] ? (string) $variables['data']['showing-results-for-text'] : "";
```

sinejespersen (approved):

PR #327, opened 2023-05-22
itkdev: reservation list settings

sinejespersen:
#### Link to issue

n/a

#### Description

reservation list config added

### DIFF
--- a/web/modules/custom/dpl_reservations/src/Form/ReservationListSettingsForm.php
+++ b/web/modules/custom/dpl_reservations/src/Form/ReservationListSettingsForm.php
@@ -74,6 +74,41 @@ class ReservationListSettingsForm extends ConfigFormBase {
       '#step' => 1,
     ];
 
+    $form['settings']['reservation_detail_allow_remove_ready_reservations_config'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow removing ready reservations', [], ['context' => 'Reservation list (settings)']),
+      '#default_value' => $config->get('reservation_detail_allow_remove_ready_reservations_config'),
+    ];
+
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_reservations/src/Form/ReservationListSettingsForm.php
```diff
@@ -74,6 +74,41 @@ public function buildForm(array $form, FormStateInterface $form_state): array {
...
+      '#title' => $this->t('Allow settings interest period to 6 months', [], ['context' => 'Reservation list (settings)']),
+      '#default_value' => $config->get('interest_period_six_months_config_text'),
+    ];
+
+    $form['settings']['interest_period_one_year_config_text'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Allow settings interest period to 12 months', [], ['context' => 'Reservation list (settings)']),
+      '#default_value' => $config->get('interest_period_one_year_config_text'),
+    ];
+
```
spaceo, line 111: I think we can agree on that the implementation of interest period config as text was wrong from the beginning. It should never have been anything else than config. But so be it.
But using a data group like this should be solved via [checkboxes](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Checkboxes.php/class/Checkboxes/10) in Drupal Form Api.

spaceo (changes requested):
I think we can agree on that the implementation of interest period config as text was wrong from the beginning. It should never have been anything else than config. But so be it.
But using a data group like this should be solved via [checkboxes](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Checkboxes.php/class/Checkboxes/10) in Drupal Form Api.

PR #334, opened 2023-05-30
Itkdev: interest period config

sinejespersen:
#### Link to issue

n/a

#### Description

Change format of interest periods

https://github.com/danskernesdigitalebibliotek/dpl-react/pull/463

#### Screenshot of the result

image

#### Checklist

n/a

### DIFF
--- a/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
+++ b/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
@@ -156,7 +156,7 @@ class DplReactAppsController extends ControllerBase {
   }
 
   /**
-   * Get a string with interest periods
+   * Get a string with interest periods.
    *
    * @return string
    *   A string with interest periods
--- a/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
+++ b/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
@@ -156,6 +156,16 @@ class DplReactAppsController extends ControllerBase {
   }
 
   /**
+   * Get a string with interest periods
+   *
+   * @return string
+   *   A string with interest periods
+   */
+  public static function getInterestPeriods(): string {
+    return \Drupal::configFactory()->get('dpl_library_agency.general_settings')->get('interest_periods_config');
+  }
+
+  /**
    * Render work page.
    *
    * @param string $wid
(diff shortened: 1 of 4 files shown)


cableman (approved):

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -155,6 +155,16 @@ public function search(): array {
...
   }
 
+  /**
+   * Get a string with interest periods.
+   *
+   * @return string
+   *   A string with interest periods
+   */
+  public static function getInterestPeriods(): string {
+    return \Drupal::configFactory()->get('dpl_library_agency.general_settings')->get('interest_periods_config');
```
spaceo, line 165: I think you should add the general settings as a dependency like the other settings:
https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/0c6d60916b6425243ef1c9b0ce3871b2eab9c4ee/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php#L44-L48
To keep it persistent and make it easier to test.

spaceo (changes requested):
Looks good. But please look at my comment.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
cableman: That is not possible here as the function is static and don't have access to `$this`.

cableman (approved):

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
spaceo: Ok I have been digging into the various ways we are handling settings and it is not consistent.
I understand why you need to solve it like this, but I think we can make it more consistent by creating services for the different areas of the configuration (like `DplInstantLoanSettings`) in the future.

spaceo (approved):
Approved but please look at my comment

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
cableman: I hear you but I don't think that this PR should change how the general settings is handled. But "yes" to the idea to change this into the settings pattern now applied in may modules.

cableman: Todo added and ticket on the way about fixing this into the settings pattern

PR #339, opened 2023-06-01
itkdev: dashboard props updates

sinejespersen:
#### Link to issue

n/a

#### Description

A lot of props changed in, so this pr matches that: 
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/462

Also, added missing config and moved ereolen-url-config to GS

and moved ereolen_my_page_url to general settings and added it to the yml instead of hardcoded config 🐶
#### Screenshot of the result

n/a 

#### Checklist

n/a


web/modules/custom/dpl_patron_page/src/Plugin/Block/PatronPageBlock.php
```diff
@@ -103,7 +97,7 @@ public function build() {
       'pause-reservation-start-date-config' => $dateConfig,
       'pincode-length-max-config' => $patron_page_settings->get('pincode_length_max'),
       'pincode-length-min-config' => $patron_page_settings->get('pincode_length_min'),
-      'text-notifications-enabled-config' => (int) $this->textNotificationsEnabled(),
+      'text-notifications-enabled-config' => (int) $this->reservationSettings->smsNotificationsIsEnabled(),
```
cableman, line 100: This will not work you have to loadConfig first into variable as with the `patron_page_settings`

cableman (changes requested):

cableman (approved):

web/modules/custom/dpl_dashboard/src/Form/DashboardSettingsForm.php
```diff
@@ -108,6 +108,12 @@ public function buildForm(array $form, FormStateInterface $form_state): array {
       '#default_value' => $config->get('reservations_url') ?? '',
     ];
 
+    $form['settings']['fees_url'] = [
+      '#type' => 'url',
+      '#title' => $this->t('Fees url', [], ['context' => 'Dashboard (settings)']),
+      '#default_value' => $config->get('fees_url') ?? '',
+    ];
```
spaceo, line 115: Still find it odd with no validation. Will the app work if this is just partially filled?

web/modules/custom/dpl_patron_page/src/Plugin/Block/PatronPageBlock.php
spaceo, line 100: How was it with this setting? Fo we have it twice in the admin now or did someone fix it?

sinejespersen: This is me fixing it

web/modules/custom/dpl_dashboard/src/Form/DashboardSettingsForm.php
sinejespersen: We will discuss this with fddf during the e2e-test

PR #342, opened 2023-06-07
itkdev: props updated

sinejespersen:
#### Link to issue

n/a

#### Description

Linked to this: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/474

Updated props to match new modals

#### Screenshot of the result

n/a

#### Checklist

n/a


web/modules/custom/dpl_patron_menu/src/Plugin/Block/PatronMenuBlock.php
```diff
@@ -55,15 +64,18 @@ public function __construct(array $configuration, $plugin_id, $plugin_definition
...
    */
   public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
     return new static(
       $configuration,
       $plugin_id,
       $plugin_definition,
       $container->get('config.factory'),
+      $container->get('dpl_library_agency.branch_settings'),
+      $container->get('dpl_library_agency.branch.repository'),
+      \Drupal::service('dpl_patron_menu.settings')
```
spaceo, line 78: It should be possible to get it from the container. Isn't it possible to get it like:
`$container->get('dpl_library_agency.branch.repository')`?

spaceo (changes requested):
Looks good but please look at my comment about the `PatronMenuBlock ::create` method

cableman (approved):

web/modules/custom/dpl_patron_menu/src/Plugin/Block/PatronMenuBlock.php
cableman: I think it would be, but this follows the pattern used in all the other modules.

spaceo (approved):
LGTM 👍

PR #355, opened 2023-06-12
Feature/remove favorites list mc config

sinejespersen:
#### Link to issue

n/a

#### Description

Remove unused config
add missing props

#### Screenshot of the result

n/a

#### Checklist

n/a


cableman (approved):

web/modules/custom/dpl_dashboard/src/Plugin/Block/DashboardBlock.php
```diff
@@ -102,11 +102,14 @@ public function build(): array {
       'ereolen-my-page-url' => $generalSettings->get('ereolen_my_page_url'),
 
       // Texts.
+      'reservation-details-others-in-queue-text' => $this->t('Others are queueing for this material', [], ['context' => 'Dashboard']),
```
spaceo, line 105: Please double check with the search app strings if some trnaslations could be resused. Stupid to translate it twice. Or what do you think?

spaceo (approved):
LGTM 👍

PR #356, opened 2023-06-12
ITKDev - Disable anonymous page cache for redirects

cableman:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-629

#### Description

Response built from the ThrustedRedirectReponse class is not cached.

But the problem here is the page cache for anonymous requests, which caches all responses, even redirects and no matter if they are cacheable or not.

This is a big problem for /user/me/* as it will create a redirect loop for login, which make open id connect throw an 500 back. 

#### Screenshot of the result

N/A

#### Checklist

N/A

### DIFF
--- a/web/modules/custom/dpl_patron_redirect/src/EventSubscriber/RedirectPatronSubscriber.php
+++ b/web/modules/custom/dpl_patron_redirect/src/EventSubscriber/RedirectPatronSubscriber.php
@@ -4,6 +4,7 @@ namespace Drupal\dpl_patron_redirect\EventSubscriber;
 
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Config\ImmutableConfig;
+use Drupal\Core\PageCache\ResponsePolicy\KillSwitch;
 use Drupal\Core\Path\CurrentPathStack;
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_patron_redirect/src/EventSubscriber/RedirectPatronSubscriber.php
```diff
@@ -74,6 +74,12 @@ public function checkAuthStatus(RequestEvent $event): void {
         // ignored.
         $_SESSION['openid_connect_destination'] = $path;
 
+        // Response built from the ThrustedRedirectReponse class is not cached.
+        // But the problem here is the page cache for anonymous requests, which
+        // caches all responses, even redirects and no matter if they are
+        // cacheable or not. This will kill that cache.
+        \Drupal::service('page_cache_kill_switch')->trigger();
```
spaceo, line 81: Should be added in constructor and in `*.services.yml`

spaceo (changes requested):

spaceo (approved):
John 👍

sinejespersen (approved):

PR #357, opened 2023-06-12
Clear cache when interface language changes

spaceo:
With this change all apps gets their cache invalidated when user interface text is changed

#### Link to issue

#### Description

Discovered that interface texts did not change when an administrator updated translation.

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_react/dpl_react.module
+++ b/web/modules/custom/dpl_react/dpl_react.module
@@ -142,6 +142,11 @@ function dpl_react_preprocess_dpl_react_app(array &$variables): void {
 
   $variables['#cache']['tags'][] = 'dpl_react_app';
   $variables['#cache']['tags'][] = 'dpl_react_app:' . $variables['name'];
+
+  // When changes to the interface language occurs then skip cache.
+  if (!in_array('languages:language_interface', $variables['#cache']['contexts'] ?? [])) {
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_react/dpl_react.module
```diff
@@ -142,6 +142,11 @@ function dpl_react_preprocess_dpl_react_app(array &$variables): void {
 
   $variables['#cache']['tags'][] = 'dpl_react_app';
   $variables['#cache']['tags'][] = 'dpl_react_app:' . $variables['name'];
+
+  // When changes to the interface language occurs then skip cache.
+  if (!in_array('languages:language_interface', $variables['#cache']['contexts'] ?? [])) {
+    $variables['#cache']['contexts'][] = 'languages:language_interface';
+  }
```
kasperg, line 149:
```suggestion
  // Apps typically use *-text properties passing translated static texts from Drupal
  // to apps. Add language as a cache context to invalidate caches when translations 
  // are updated.
  $variables['#cache']['contexts'][] = 'languages:language_interface';
```

I am not sure it make a difference adding checks to whether the cache context is already there (it should not) or whether the contexts array is already set (we do not do the same for cache tags).

kasperg (approved):
LGTM

PR #361, opened 2023-06-20
ITKDev - Set missing prop-types for patron reg.

sinejespersen:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-662

#### Description

Patron registration was missing some configuration.

#### Screenshot of the result

N/A

#### Checklist

N/A


cableman (approved):

web/modules/custom/dpl_patron_reg/src/Plugin/Block/PatronRegistrationBlock.php
```diff
@@ -139,19 +150,21 @@ public function build(): array {
       'pincode-length-min-config' => $patron_page_settings->get('pincode_length_min'),
       'redirect-on-user-created-url' => $config->get('redirect_on_user_created_url'),
       'user-token' => $userToken,
+      'login-url' => $userInfoEndpoint,
+      'text-notifications-enabled-config' => (int) $this->reservationSettings->smsNotificationsIsEnabled(),
```
spaceo, line 154: There is two different config keys at play right now.
The one on this line and [this one](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/0ce9f5510150be5aa822bd8326a6ae3af7eefb6f/web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php#L193).
Should be just one config key to avoid confusion.

spaceo (changes requested):

PR #363, opened 2023-06-24
Fix locale update cache invalidation

spaceo:
#### Link to issue

#### Description
Invalidate locale cache tag when running locale-update:
Since the locale cache tag is not invalidated when running
`drush locale-update` we need to make sure it happens ourselves
by injecting our own batch finished callback.

#### Checklist


kasperg (approved):
:+1: Given the rather complex situation I think this is a good approach.

Adamik10 (approved):
LGTM

PR #365, opened 2023-06-26
Set selected site language to 'en' in CI context

spaceo:
#### Link to issue

#### Description

We have tests asserting various UI texts. So we need to have some static strings.

#### Checklist


kasperg (approved):
:+1: LGTM

Adamik10 (approved):
LGTM

PR #373, opened 2023-07-18
5887: differentiate translation texts

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5887

#### Description

Different strings for translations


jeppekroghitk (approved):

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -490,12 +490,12 @@ public static function getBlockedSettings(): array {
...
+      'blocked-patron-w-title-text' => t('Your user is blocked (w)', [], ['context' => 'Blocked user']),
+      'blocked-patron-w-body-text' => t('You therefore cannot reserve, borrow or renew loans. Please contact the library for further information (w)', [], ['context' => 'Blocked user']),
       'blocked-patron-o-title-text' => t('Blocked reason O modal title', [], ['context' => 'Blocked user']),
       'blocked-patron-o-body-text' => t('Blocked patron o body text', [], ['context' => 'Blocked user']),
-      'blocked-patron-u-title-text' => t('Your user is blocked', [], ['context' => 'Blocked user']),
-      'blocked-patron-u-body-text' => t('You therefore cannot reserve, borrow or renew loans. Please contact the library for further information', [], ['context' => 'Blocked user']),
+      'blocked-patron-u-title-text' => t('Your user is blocked (u)', [], ['context' => 'Blocked user']),
+      'blocked-patron-u-body-text' => t('You therefore cannot reserve, borrow or renew loans. Please contact the library for further information (u)', [], ['context' => 'Blocked user']),
```
kasperg, line 498: Please use the same pattern as for all blocked texts to ensure consistency.

I do not understand what `d`, `s`, `f`, `w` and `u`means and I guess it is outside the scope of this PR to provide this information. Preceeding lines add the special letters in the middle of the text e.g. "Blocked patron d title text". Here it is added in a parenthesis at the end.  

Personally I perfer the last option but the most important thing is to keep it consistent. This makes it easier to editors to look up all variations of the text.

I suggest that you change the title and body texts in this PR to the existing forms e.g. "Blocked patron w title text" and "Blocked patron w body text".

Alternately change the existing texts to the form introduced in this PR e.g. "Your user is blocked (d)", "You therefore cannot reserve, borrow or renew loans. Please contact the library for further information (d)".

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -146,8 +146,8 @@ public function build() {
       'loan-list-status-badge-danger-text' => $this->t("Expired", [], ['context' => 'Loan list']),
       'loan-list-status-badge-warning-text' => $this->t("Expiring soon", [], ['context' => 'Loan list']),
       'loan-list-title-text' => $this->t("Your loans", [], ['context' => 'Loan list']),
-      'loan-list-to-be-delivered-digital-material-text' => $this->t("Due date @date", [], ['context' => 'Loan list']),
-      'loan-list-to-be-delivered-text' => $this->t("Due date @date", [], ['context' => 'Loan list']),
+      'loan-list-to-be-delivered-digital-material-text' => $this->t("Due date digital @date", [], ['context' => 'Loan list']),
+      'loan-list-to-be-delivered-text' => $this->t("Due date physical @date", [], ['context' => 'Loan list']),
```
kasperg, line 150: Generally I am not too keen on having texts which are not in a form which is suitable for end users as the source. 

In these cases it would be preferable to either see if we can find a more suitable text or use the context to differ.  I do not have a good solution in the current case so let's just go with this for now.

kasperg (changes requested):
:+1: but please look into the blocked texts.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
jeppekroghitk: I fixed this. Went with the parentheses at the end for them all.

PR #374, opened 2023-08-07
Analyse profile with PHPStan and PHPCS

kasperg:
#### Description

Currently we only look in custom modules and themes. We need to look in the profile as well.

Check out the commits and action results to see the problems raised which have subsequently been fixed.

#### Checklist

I originally expected this to identify the problem in #372. It turns out this is not the case. [Apparently it is because PHPStan will not raise errors with undefined variables and `??`/null coalescence operator](https://phpstan.org/r/a54d4eeb-c276-479d-8102-9b5c56a4ff80).


spaceo (approved):
LGTM 👍

PR #375, opened 2023-08-11
5863: Removing some unused link inputs

jeppekroghitk:
...and replacing some with hardcoded counterparts in dpl-react

#### Link to issue

[Please add a link to the issue being addressed by this change.](https://platform.dandigbib.org/issues/5863)

#### Description

Some of the links previously to be defined by each library branch, is now to be hardcoded. e.g (/user/me/loans, /user/me/fees etc)

#### Screenshot of the result

N/A

#### Checklist

### DIFF
--- a/web/modules/custom/dpl_dashboard/src/Form/DashboardSettingsForm.php
+++ b/web/modules/custom/dpl_dashboard/src/Form/DashboardSettingsForm.php
@@ -66,54 +66,6 @@ class DashboardSettingsForm extends ConfigFormBase {
       '#tree' => FALSE,
     ];
 
-    $form['settings']['intermediate_url'] = [
-      '#type' => 'url',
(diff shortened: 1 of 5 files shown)


kasperg (changes requested):
I do not understand the reasoning behind all of this change.

1. I agree with the fact that it is unreasonable to expect library site administrators to define urls within their own system that are not expect to be changable by an editor anyway.
2. I still think it is relevant for developers to be able to pass information about urls from the CMS (where developers assemble components to an entire application) to the React components (which do not know where they might live.

All in all I agree with the removal of configuration forms etc in `DashboardSettingsForm` etc. I do not agree with not passing this information ass props from Drupal to React components in `DashboardBlock` etc.

Passing urls as information about how components should link elsewhere resembles what we already do for the search url.

PR #378, opened 2023-08-17
ITKDev: Fixed missing user token in patron reg. flow

cableman:
#### Link to issue

#### Description

Changed login flow when it's part of the user registration flow.

#### Screenshot of the result

N/A

#### Checklist

N/A

### DIFF
--- a/web/modules/custom/dpl_login/dpl_login.module
+++ b/web/modules/custom/dpl_login/dpl_login.module
@@ -21,12 +21,32 @@ use function Safe\sprintf as sprintf;
  *   User info from external service.
  * @param mixed[] $context
  *   Various openid_connect context. Tokens etc.
+ *
+ * @return void
+ *
+ * @throws \Drupal\dpl_login\Exception\AccessTokenCreationException
+ * @throws \Safe\Exceptions\StringsException
+ * @throws \Drupal\Core\TempStore\TempStoreException
  */
 function dpl_login_openid_connect_userinfo_alter(array &$userinfo, array $context): void {
 
-  // If we cannot resolve uinque id we cannot continue.
+  // If we cannot resolve unique id we cannot continue.
   if (!$id = $userinfo['attributes']['uniqueId'] ?? FALSE) {
+    // If this is n user registration flow we still need the user token to
+    // create the user in FBS from React.
+    /** @var Drupal\Core\TempStore\PrivateTempStore $store */
+    $store = \Drupal::service('tempstore.private')->get('dpl_patron_reg');
+
+    if ($store->get('isRegistrationFlow')) {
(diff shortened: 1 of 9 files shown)


sinejespersen (approved):

jeppekroghitk (approved):

web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -21,12 +21,30 @@ use function Safe\sprintf as sprintf;
...
+    // create the user in FBS from React.
+    /** @var Drupal\Core\TempStore\PrivateTempStore $store */
+    $store = \Drupal::service('tempstore.private')->get('dpl_patron_reg');
+
+    if ($store->get('isRegistrationFlow')) {
+      $access_token = AccessToken::createFromOpenidConnectContext($context);
+      /** @var Drupal\dpl_login\UserTokensProvider $token_provider */
+      $token_provider = \Drupal::service('dpl_login.user_tokens');
+      $token_provider->setAccessToken($access_token);
+      $store->delete('isRegistrationFlow');
```
spaceo, line 43: Suggestion: Instead of reading/deleting from the tempstore in an imperative way, it would be nice with some methods in the dpl_patron_reg namespace: Eg.: Something::userIsInRegistrationFlow() and Something::clearUserRegistrationFlowStatus() and Something::setUserIsInRegistrationFlowStatus().

web/modules/custom/dpl_patron_reg/src/Controller/DplPatronRegController.php
```diff
@@ -145,6 +149,11 @@ public function authRedirect(Request $request, string $client_name): TrustedRedi
     $url->setAbsolute();
     $url = $url->toString();
 
+    // Set special cookie to detect this is patron registration flow. This is
+    // used in dpl_login.
+    $store = $this->privateTempStore->get('dpl_patron_reg');
+    $store->set('isRegistrationFlow', TRUE);
```
spaceo, line 155: Suggestion: Instead of reading/deleting from the tempstore in an imperative way, it would be nice with some methods in the dpl_patron_reg namespace: Eg.: Something::userIsInRegistrationFlow() and Something::clearUserRegistrationFlowStatus() and Something::setUserIsInRegistrationFlowStatus().

spaceo (commented):
Great that you found a solution!
I have a couple of comments about the same matter

spaceo:
I am also assigning @kasperg : He could have some opinions about this as well.

kasperg (changes requested):
:+1: for creating the ADR for this. It is a complex issue.

I think the relationship between the `dpl_login` and `dpl_patron_reg` modules needs to be resolved (or at least investigated) before this can be merged.

PR #381, opened 2023-08-23
Ddfdpdel 216 manglende t function pa tekst strenge

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFDPDEL-216

#### Description

Adds translation strings that were missing.
And added the alert text in a global preprocess hook. In the future we can have other strings/properties that are global and used across the apps.

This PR relates to:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/518

#### Checklist


Adamik10 (approved):
LGTM. How about the availability text? Wasn't that a part of this? 🤔 Or maybe it already is in DPL CMS. I just wanted to mention it in case.

web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -49,3 +49,17 @@ function dpl_react_apps_preprocess_page(array &$variables): void {
...
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ *
+ * @param mixed[] $variables
+ *   The variables for the theme hook.
+ */
+function dpl_react_apps_preprocess_dpl_react_app(array &$variables): void {
+  // Since we need these texts in all the apps we add them here.
```
kasperg, line 60: :+1: nice commenting.

web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -49,3 +49,17 @@ function dpl_react_apps_preprocess_page(array &$variables): void {
...
+  $variables['attributes'] += [
+    'data-alert-error-close-text' => t('Close', [], ['context' => 'React apps global: Error handling']),
+    'data-alert-error-message-text' => t('An error occurred', [], ['context' => 'React apps global: Error handling']),
+  ];
```
kasperg, line 64:
```suggestion
  $variables['attributes'] += [
    'data-alert-error-close-text' => t('Close', [], ['context' => 'React apps (Global error handling)']),
    'data-alert-error-message-text' => t('An error occurred', [], ['context' => 'React apps (Global error handling)']),
  ];
```

As I see it we are struggling a bit with subsections of contexts. [Other places I see us using `([subsection])` to delimit subsections such as ARIA](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/e314165d6278cd4624e45c3c92e289a832b55a84/web/modules/custom/dpl_fees/src/Plugin/Block/FeesListBlock.php#L79). For the sake of consistency this suggests continuing with that pattern instead of doing something else.

kasperg (approved):
:+1: I have one suggestion for you to consider. Otherwise LGTM.

web/modules/custom/dpl_react_apps/dpl_react_apps.module
spaceo: Good point 👍

PR #382, opened 2023-08-24
Add multiple match configuration for instant loans

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DSC-69

#### Description

It should be possible to match materials by multiple strings so the configuration form element has been changed from a text input field to a textarea.

#### Checklist

The PR is connected to the changes in this [PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/519).

### DIFF
--- a/web/modules/custom/dpl_instant_loan/src/Form/DplInstantLoanSettingsForm.php
+++ b/web/modules/custom/dpl_instant_loan/src/Form/DplInstantLoanSettingsForm.php
@@ -87,7 +87,7 @@ class DplInstantLoanSettingsForm extends ConfigFormBase {
     $form['match_strings'] = [
       '#type' => 'textarea',
       '#title' => $this->t('Match Strings', [], ['context' => 'dpl_instant_loan']),
-    // Set the number of visible rows for the textarea.
+      // Set the number of visible rows for the textarea.
       '#rows' => 5,
       '#description' => $this->t('Text used to identify materials which are available for instant loans.<br/> You can write multiple strings - each on a spearate line.<br/> To find a match one of the strings must be present in the material group of such materials.'),
       '#default_value' => implode("\n", $config->get('match_strings')),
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_instant_loan/src/Form/DplInstantLoanSettingsForm.php
```diff
@@ -84,15 +84,13 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
-      '#description' => $this->t(
-        'Text used to identify materials which are available for instant loans. This text must be present in the material group of such materials.',
-        [],
-        ['context' => 'dpl_instant_loan']
-      ),
-      '#default_value' => $config->get('match_string'),
+    $form['match_strings'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Match Strings', [], ['context' => 'dpl_instant_loan']),
+    // Set the number of visible rows for the textarea.
```
kasperg, line 90:
```suggestion
      // Set the number of visible rows for the textarea.
```

Indentation seems off here.

web/modules/custom/dpl_instant_loan/src/Form/DplInstantLoanSettingsForm.php
```diff
@@ -84,15 +84,13 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#rows' => 5,
+      '#description' => $this->t('Text used to identify materials which are available for instant loans.<br/> You can write multiple strings - each on a spearate line.<br/> To find a match one of the strings must be present in the material group of such materials.'),
```
kasperg, line 92: :+1: for elaborating, but the context should not be removed.

kasperg (approved):
:+1: but retain the context in the description translation.

Adamik10 (approved):
👍

PR #388, opened 2023-09-18
Disable preprocessing of JavaScript from React component libraries

kasperg:
#### Description

Preprocessing means that the files will be aggregated if enabled. Aggregation disables sourcemaps which makes JavaScript errors significantly harder to debug.

We could toggle this through configuration but we do not want to  disable aggregation for other JavaScript files.

This may degrade performance because multiple files will have to be downloaded. On the other hand features like HTTP/2 and browser caching should make this less of a problem in practice. Not using aggregation should also enable browsers to cache the files once downloaded as they will not change across pages.

#### Checklist


spaceo (approved):
LGTM 👍

PR #393, opened 2023-09-29
Fix syntax for Dependabot Drupal Core groups

kasperg:
#### Description

There is no include-patterns option. It is just called patterns.

The problem causes other PRs to fail e.g. https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/374/checks?check_run_id=17258750091


spaceo (approved):
LGTM 👍

PR #399, opened 2023-10-08
Add renewal feedback texts

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-167

#### Description
Adds renewal feedback texts to apps.
It is a mess what belongs where with all the text props. This is an attempt to make sure that all the renewal response texts are being available for the corresponding apps.

Things to remmeber for myself:
* Switch back dpl-react dep to develop
* Make sure this has been approved before merge of this PR: https://reload.atlassian.net/browse/DDFSOEG-553

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -65,7 +65,7 @@ environments:
       schedule: "M/30 * * * *"
       command: drush locale-check && drush locale-update
       service: cli
-  pr-396:
+  pr-399:
     cronjobs:
     - name: drush cron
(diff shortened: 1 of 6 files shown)


Adamik10 (approved):
It's going to be a bit messy trying to figure out what texts are where. I think you had a good point at the scrum that we could add all of them into all apps. Then maybe just divide them with comments to have some sort of understanding of where they belong.

PR #405, opened 2023-10-10
Set PR password to a fixed password from a lagoon

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-580

#### Description

This PR makes it possible to set a fixed password for the drupal admin user in our PR environments.


Adamik10 (approved):
Yay! Very good step forward! ✨ 👣

PR #410, opened 2023-10-11
Deny login for blocked users

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-151

#### Description

This PR implements denied login for patrons who are blocked in FBS for certain reasons. Implementing this has been a somewhat length process requiring the following steps:

1. Updating the generated FBI Client to include the Patron API
2. Refactoring the DPL FBS module to move API handling here from the Library Agency module and avoiding cyclic dependencies with Login and/or Library Token modules
3. Implementing blocking in Login module through OpenID Connect module hooks
4. Rendering system messages in the theme to show error messages if login is blocked

I recommend going commit by commit.

#### Screenshot of the result

Monosnap Log ind | DPL CMS — Privat browsing 2023-10-12 22-30-01

Note that system messages are currently not styled. I hope we can get something at a later point through work on other tickets.


web/modules/custom/dpl_library_agency/src/Branch/FbsBranchRepository.php
```diff
@@ -13,22 +14,29 @@ class FbsBranchRepository implements BranchRepositoryInterface {
...
 
   /**
    * {@inheritdoc}
    *
    * @throws \DanskernesDigitaleBibliotek\FBS\ApiException
    */
   public function getBranches(): array {
+    // Use an empty token if we don't have one. This may cause the API to throw
+    // an exception, but that is acceptable.
+    $api = $this->apiFactory->getAgencyApi($this->tokenHandler->getToken() ?? '');
```
spaceo, line 36: I think it would be prettier if the AgencyApi could handle that there should be a token for its requests instead of passing it as an argument.

web/modules/custom/dpl_fbs/src/FbsApiFactory.php
```diff
@@ -50,4 +51,14 @@ public function getAgencyApi(string $token): ExternalV1AgencyidApi {
     );
   }
 
+  /**
+   * Generate a patron API instance.
+   */
+  public function getPatronApi(string $token): ExternalAgencyidPatronsApi {
```
spaceo, line 57: As I wrote in an earlier comment I think it would be nice if the factory gets it token by itself. Could it be an idea to do it in a shared baseClass?

web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -42,6 +44,39 @@ function dpl_login_openid_connect_userinfo_alter(array &$userinfo, array $contex
...
+  $api_factory = \Drupal::service('dpl_fbs.api_factory');
+  $api = $api_factory->getPatronApi($token->token);
+  $patron = $api->getPatronInformationByPatronIdV2();
+
+  $patron_block_status = $patron->getPatron()?->getBlockStatus() ?? [];
+  $block_login = array_map(function (BlockStatus $block_status) {
+    return $block_status->getBlockedReason();
+  }, $patron_block_status);
+
+  // Return FALSE to block login.
```
spaceo: Thx for the comment here 😅

web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -69,12 +70,31 @@ function dpl_login_openid_connect_pre_authorize(AccountInterface|FALSE $account,
...
+    try {
+      return BlockedReason::from($block_status->getBlockedReason());
+    }
+    catch (UnexpectedValueException $e) {
+      \Drupal::logger('dpl_login')->warning('Unable to map blocked reason @reason: @message', [
+        '@reason' => $block_status->getBlockedReason(),
+        '@message' => $e->getMessage(),
+      ]);
+      return BlockedReason::UNKNOWN();
+    }
```
spaceo: Nice construction!

spaceo (approved):
LGTM 👍 
Great work
I have some comments for you to consider

PR #412, opened 2023-10-12
Show login status in page title

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-86

#### Description

Add "Logged in" suffix in the HTML head title when users are logged in.

This applies for all users - both patrons and library staff.

This assumes that patrons will be logged in as both Drupal users and
with Adgangsplatformen tokens as is currently the case.

Update our PHPStan setup to not complain about typing for preprocess variable arrays.

#### Screenshot of the result

Monosnap admin | DPL CMS | Logged in 2023-10-12 10-03-22


Adamik10 (approved):
Good stuff 💯

PR #432, opened 2023-10-19
Fix permissions for translation workflows

kasperg:
#### Description

Publishing to GitHub Pages using with peaceiris/actions-gh-pages requires general permission to commit to git - not specifically for Pages.

See https://github.com/peaceiris/actions-gh-pages#user-content-%EF%B8%8F-first-deployment-with-github_token

The current setup causes GitHub Actions to fail.

#### Screenshot of the result

dpl-cms@a1d1e6b 2023-10-19 14-35-00

See example of failing action with the current setup.

### DIFF
--- a/.github/workflows/publish-source.yml
+++ b/.github/workflows/publish-source.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #435, opened 2023-10-20
Exclude all Drupal packages from Dependabot Composer packages group

kasperg:
#### Description

With this update with now both exclude both Drupal Core and Contrib. It was not excluded before as shown in #422 

Experience shows that it is not a good idea to do automatic updates of contrib when we do not have sufficient test coverage.


spaceo (approved):
LGTM 👍

PR #436, opened 2023-10-23
Defaults in dpl-react configuration (review but do not merge, DDF needs it for testing)

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-246

#### Description

This PR is trying get rid of failing dpl-react apps when various data-props has not been configured.
* The urls that belongs to the various modules through their *.routing.yml files has been moved their corresponding modules via the `hook_dpl_react_apps_data()` hook.
* Defaults have been set as constants on settings objects so they can be used in both config forms and as rendering fallback data-props.
* App url handling functions have been moved to dpl_reat_apps.module. We could also create a service for it? But anyways, this is an attempt to try to handle urls passed to apps in a unified manner.
* Some naming/logic cleanup around `reservation_sms_notifications_enabled`. I think I have done it right, but it should be double checked with an (manual) integration test.

[This](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/633) should be merged first.
Then the dpl-react dependency should point at develop.
When this PR has been approved it should NOT be merged! - because  DDF needs a testing environment
Then this PR can be merged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
             "type": "package",
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-react",
-                "version": "0.3.0-rc42",
+                "version": "0.3.0-rc43",
                 "type": "drupal-library",
                 "dist": {
                     "url": "https://github.com/reload/dpl-react/releases/download/release-DDFLSBP-246-vi-skal-sorge-for-at-vi-har-defaults-pa-al-configuration-af-dpl-react-apps/dist.zip",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
                 "version": "0.3.0-rc42",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/release-develop/dist.zip",
+                    "url": "https://github.com/reload/dpl-react/releases/download/release-DDFLSBP-246-vi-skal-sorge-for-at-vi-har-defaults-pa-al-configuration-af-dpl-react-apps/dist.zip",
                     "type": "zip"
                 },
                 "require": {
--- a/web/modules/custom/dpl_dashboard/src/Plugin/Block/DashboardBlock.php
+++ b/web/modules/custom/dpl_dashboard/src/Plugin/Block/DashboardBlock.php
@@ -93,7 +93,7 @@ class DashboardBlock extends BlockBase implements ContainerFactoryPluginInterfac
       // Urls.
       // Cannot find that route. Does it exist?
       'intermediate-url' => '/user/me/intermediates',
-      'ereolen-my-page-url' => $generalSettings->get('ereolen_my_page_url') ?? GeneralSettingsForm::EREOLEN_MY_PAGE_URL,
+      'ereolen-my-page-url' => dpl_react_apps_format_app_url($generalSettings->get('ereolen_my_page_url'), GeneralSettingsForm::EREOLEN_MY_PAGE_URL),
 
       // Texts.
       'reservation-details-others-in-queue-text' => $this->t('Others are queueing for this material', [], ['context' => 'Dashboard']),
--- a/web/modules/custom/dpl_dashboard/src/Plugin/Block/DashboardBlock.php
+++ b/web/modules/custom/dpl_dashboard/src/Plugin/Block/DashboardBlock.php
@@ -5,6 +5,9 @@ namespace Drupal\dpl_dashboard\Plugin\Block;
 use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Url;
+use Drupal\dpl_dashboard\DplDashboardSettings;
+use Drupal\dpl_library_agency\Form\GeneralSettingsForm;
 use Drupal\dpl_react\DplReactConfigInterface;
 use Drupal\dpl_react_apps\Controller\DplReactAppsController;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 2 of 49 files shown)


composer.json
```diff
@@ -20,7 +20,7 @@
                 "version": "0.3.0-rc42",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/release-develop/dist.zip",
+                    "url": "https://github.com/reload/dpl-react/releases/download/release-DDFLSBP-246-vi-skal-sorge-for-at-vi-har-defaults-pa-al-configuration-af-dpl-react-apps/dist.zip",
```
kasperg, line 23: Remember to remove this once both are approved.

web/modules/custom/dpl_dashboard/src/Plugin/Block/DashboardBlock.php
```diff
@@ -79,22 +81,18 @@ public function build(): array {
...
       // Urls.
-      'dpl-cms-base-url' => DplReactAppsController::dplCmsBaseUrl(),
-      'fees-page-url' => '/user/me/fees',
+      // Cannot find that route. Does it exist?
       'intermediate-url' => '/user/me/intermediates',
-      'reservations-url' => '/user/me/reservations',
-      'physical-loans-url' => '/user/me/loans',
-      'search-url' => DplReactAppsController::searchResultUrl(),
-      'ereolen-my-page-url' => $generalSettings->get('ereolen_my_page_url'),
+      'ereolen-my-page-url' => dpl_react_apps_format_app_url($generalSettings->get('ereolen_my_page_url'), GeneralSettingsForm::EREOLEN_MY_PAGE_URL),
```
kasperg, line 95: Please remember to add a dependency on `dpl_react_apps` and `dpl_library_agency` when we use classes and functions from these modules.

web/modules/custom/dpl_loans/src/DplLoansSettings.php
```diff
@@ -9,6 +9,10 @@
  */
 class DplLoansSettings extends DplReactConfigBase {
 
+  const MATERIAL_OVERDUE_URL = '';
```
kasperg, line 12: Do urls have to be actual urls instead of empty strings?

web/modules/custom/dpl_reservations/src/Plugin/Block/ReservationListBlock.php
```diff
@@ -85,16 +89,18 @@ public function build(): array {
...
       // Config.
       'interest-periods-config' => DplReactAppsController::getInterestPeriods(),
-      'page-size-desktop' => $config->get('page_size_desktop'),
-      'page-size-mobile' => $config->get('page_size_mobile'),
-      'pause-reservation-start-date-config' => $generalSettings->get('pause_reservation_start_date_config') ?? '',
-      'reservation-detail-allow-remove-ready-reservations-config' => $generalSettings->get('reservation_detail_allow_remove_ready_reservations_config'),
-      'threshold-config' => $generalSettings->get('threshold_config'),
+      'page-size-desktop' => $config->get('page_size_desktop') ?? DplReservationsSettings::PAGE_SIZE_DESKTOP,
+      'page-size-mobile' => $config->get('page_size_mobile') ?? DplReservationsSettings::PAGE_SIZE_MOBILE,
+      'pause-reservation-start-date-config' => $generalSettings->get('pause_reservation_start_date_config') ?? GeneralSettingsForm::PAUSE_RESERVATION_START_DATE_CONFIG,
```
kasperg, line 99: Please remember to add undeclared dependencies here.

web/modules/custom/dpl_login/dpl_login.info.yml
```diff
@@ -7,3 +7,4 @@ core_version_requirement: ^9
 
 dependencies:
   - openid_connect:openid_connect
+  - dpl_react_apps:dpl_react_apps
```
kasperg, line 10: Consider if this dependency can be handled on other ways. 

I am not too keen on it as it creates a dependency on a module which (should) also depend on `dpl_login`. I get what you want to do but I fear a circular relationship where I personally see DPL Login as a dependency on the React Apps and not the other way around.

One way to do this would be to let DPL React Apps depend on DPL Login and generate the urls in its own `dpl_react_apps_dpl_react_apps_data()` implementation.

web/modules/custom/dpl_patron_reg/src/Plugin/Block/PatronRegistrationBlock.php
```diff
@@ -104,7 +104,7 @@ public function build(): array {
 
     // Get user info endpoint from OpenIdConnect configuration.
     $configuration = $this->configFactory->get('openid_connect.settings.adgangsplatformen');
-    $userInfoEndpoint = $configuration->get('settings')['userinfo_endpoint'];
+    $userInfoEndpoint = $configuration->get('settings')['userinfo_endpoint'] ?? '/';
```
kasperg, line 107: Consider if there is a mixing between handling of default urls.

This seems to uses a `/` as a default value for a url. Other places a blank space is used.

web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -114,3 +130,48 @@ function dpl_react_apps_texts_renewal(): array {
...
   ];
 }
+
+/**
+ * Make sure that generated url is a string.
+ *
+ * @param string|\Drupal\Core\GeneratedUrl $url
+ *   Drupal generated Url object.
+ */
+function dpl_react_apps_ensure_url_is_string(string|GeneratedUrl $url): string {
```
kasperg, line 140: :+1: nice use of union types.

web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -114,3 +130,48 @@ function dpl_react_apps_texts_renewal(): array {
...
+ *
+ * @return string
+ *   Formatted url.
+ */
+function dpl_react_apps_format_app_url(string|NULL $url, string $fallbackUrl = ''): string {
+  if (!empty($url)) {
+    return $url;
+  }
+
+  return $fallbackUrl ?: dpl_react_apps_get_dpl_cms_base_url();
```
kasperg, line 176: Consider adding a comment about why we think it is a good idea to fall back to the base url if no url is provided.

kasperg (changes requested):
:+1: good stuff. I have some considerations for you.

Remember to consider module dependencies when referencing classes.

composer.json
spaceo: Yes 👍 
image

web/modules/custom/dpl_dashboard/src/Plugin/Block/DashboardBlock.php
spaceo: 👍

web/modules/custom/dpl_reservations/src/Plugin/Block/ReservationListBlock.php
spaceo: 👍

web/modules/custom/dpl_login/dpl_login.info.yml
spaceo: You're right I'll move it around

PR #447, opened 2023-10-30
Downgrade to lhci 0.8

kasperg:
#### Description

We upgraded to Lighthouse 0.12 as suggested by Dependabot in #434. However this has introduced multiple problems:

First:

Lighthouse CI/lhci 0.10 introduces a change where it will respect the
HTTP_PROXY url environment variable for API requests. Apparently this
includes uploading reports to temporary storage.
https://github.com/GoogleChrome/lighthouse-ci/releases/tag/v0.10.0
https://github.com/GoogleChrome/lighthouse-ci/pull/727

As we setup Wiremock as a general HTTP proxy for Lighthouse requests to
mock API requests, report uploads will also pass through Wiremock.
This results in an error when uploading reports:

FetchError: request to https://us-central1-lighthouse-infrastructure.cloudfunctions.net/saveHtmlReport failed, reason: self signed certificate in certificate chain

This is also visible in the output for the PR which introduced the upgrade #434 even though the action as a whole succeeded: https://github.com/danskernesdigitalebibliotek/dpl-cms/actions/runs/6588193517/job/17899940545

To address this we either need to move away from browser proxying so
we only use WireMock for specific external systems or setup the browser
proxy to provide a trusted certificate chain.
https://wiremock.org/docs/proxying/#browser-proxying-of-https

Second:

Running with lhci 0.9 upgrades to Lighthouse 9.3.0. This reveals new 
problems which causes our tests to fail: https://github.com/danskernesdigitalebibliotek/dpl-cms/actions/runs/6695006011/job/18201354535

LCHI 0.8 was the original version we used before Dependabot started
suggesting updates. Revert to this to ensure we have a working version.
Then we will update LHCI at a later point in time when we have room to
fix these issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "license": "AGPL-3.0",
   "devDependencies": {
     "@apidevtools/swagger-cli": "^4.0.4",
-    "@lhci/cli": "^0.12.0",
+    "@lhci/cli": "^0.9",
     "@types/node": "^18.11.11",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.38.0",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "license": "AGPL-3.0",
   "devDependencies": {
     "@apidevtools/swagger-cli": "^4.0.4",
-    "@lhci/cli": "^0.9",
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #448, opened 2023-10-30
Reduce Lighthouse throttling from 2.9x to 1x

kasperg:
#### Description

This is a result of changing our target device from a mid-range mobile to a high-end mobile.

Please see the code comments for additional commentary.


spaceo:
🎉 🎉 🎉 🎉 🎉 🎉 🎉

spaceo (approved):

PR #454, opened 2023-11-01
Ddflsbp 229 brugeroprettelse laner der er oprettet i forvejen vises en fejlside

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-229

#### Description

Users that already have been created should be redirected to the dashboard.
This PR adds a handover of the dashboard and userinfo urls to the react applications

This can be merged when it's [sibling](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/653) have been merged.

### DIFF
--- a/web/modules/custom/dpl_react_apps/dpl_react_apps.module
+++ b/web/modules/custom/dpl_react_apps/dpl_react_apps.module
@@ -78,6 +78,9 @@ function dpl_react_apps_dpl_react_apps_data(array &$data): void {
     'auth' => dpl_react_apps_ensure_url_is_string(
       Url::fromRoute('dpl_login.login')->toString()
     ),
+    'dashboard' => dpl_react_apps_ensure_url_is_string(
+      Url::fromRoute('dpl_dashboard.list')->toString()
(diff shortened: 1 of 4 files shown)


web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -53,6 +54,14 @@ function dpl_react_apps_preprocess_page(array &$variables): void {
  * Implements hook_dpl_react_apps_data().
  */
 function dpl_react_apps_dpl_react_apps_data(array &$data): void {
+  $adgangsplatformen_config = \Drupal::service('dpl_login.adgangsplatformen.config');
+
+  $data['configs'] += [
+    'agency' => json_encode([
+      'id' => $adgangsplatformen_config->getAgencyId(),
+    ]),
+  ];
```
kasperg, line 63: Consider if there is a reason to make this an object. It seems like we could use a string which is simpler.

kasperg (approved):
:+1: LGTM

PR #456, opened 2023-11-02
DDFLSBP-165 - Removed the field material_overdue_url from LoanListSet…

Dresse:
…tingsForm

#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-165

#### Description

Removed the field material_overdue_url from LoanListSettingsForm

#### Screenshot of the result

Screenshot 2023-11-02 at 15 16 25

### DIFF
--- a/web/profiles/dpl_cms/translations/da.po
+++ b/web/profiles/dpl_cms/translations/da.po
@@ -2393,14 +2393,6 @@ msgid "Basic settings"
 msgstr "Grundlæggende indstillinger"
 
 msgctxt "Loan list (settings)"
-msgid "Material overdue url"
-msgstr "Link til overskredne lån"
-
-msgctxt "Loan list (settings)"
-msgid "The link to the material overdue page"
-msgstr "Link til siden med overskredne lån"
-
-msgctxt "Loan list (settings)"
 msgid "Page size mobile"
 msgstr "Antal listeelemeter på side (mobil)"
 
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -102,7 +102,6 @@ public function build() {
 
       // Urls.
       'ereolen-my-page-url' => dpl_react_apps_format_app_url($generalSettings->get('ereolen_my_page_url'), GeneralSettingsForm::EREOLEN_MY_PAGE_URL),
-      'material-overdue-url' => dpl_react_apps_format_app_url($loanListSettings->get('material_overdue_url'), DplLoansSettings::MATERIAL_OVERDUE_URL),
```
kasperg, line 105: Please reconsider this change.

We cannot simply remove the url. The React applications still need to know where to direct users who want to see their list of overdue materials (probably loans) and this is passed through this value.

To be honest I do not know what the url for this should be but to me it seems reasonable to use the [`dpl_loans.list` route](https://github.com/reload/dpl-cms/blob/main/web/modules/custom/dpl_loans/dpl_loans.routing.yml#L1-L1) for now.

web/profiles/dpl_cms/translations/da.po
```diff
@@ -2392,14 +2392,6 @@ msgctxt "Loan list (settings)"
...
 msgstr "Grundlæggende indstillinger"
 
-msgctxt "Loan list (settings)"
-msgid "Material overdue url"
-msgstr "Link til overskredne lån"
-
-msgctxt "Loan list (settings)"
-msgid "The link to the material overdue page"
-msgstr "Link til siden med overskredne lån"
-
```
kasperg, line 2402: :+1: for remembering to clean up unnecessary translations.

kasperg (changes requested):
:+1: nicely done.

Your handling of the overdue url needs some reconsideration.

kasperg (approved):
:+1: LGTM

PR #457, opened 2023-11-05
Demo week 44 data props

spaceo:
#### Description

Data props that was found necessary to add to accommodate latest changes in the dpl-react apps.


Adamik10 (approved):
Loads of translations! NICE! :))))

PR #458, opened 2023-11-06
fix/ereol-pickup-indication

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-213

#### Description
Added a translation for generic text shown on digital reservation cards in the list of user reservations.

#### Screenshot of the result
n/a

You can disregard the failing Lagoon integration test - there are currently problems with Lagoon.


spaceo (commented):
Have you tested that the text kicks in the different views in dpl-cms no matter where you are coming from/accessing the views directly by typing the url in the browser?
The reason why I ask is that the redux/translation setup has some "flaws" in that regard. Which would take me too long to explain here :)

spaceo (approved):
LGTM 👍

PR #470, opened 2023-11-15
DDFLSBP-150 - Changed the 'threshold_config' textfield into 'expirati…

Dresse:
…on_warning_days_before_config' number field. Also did some refactoring where the field is being used.

#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-150

#### Description

This PR is only for a smaller part of DDFLSBP-150 (the threshold changes). 

I have changed the "threshold_config" textfield form field into "expiration_warning_days_before_config" numbers field. This mean that we no longer set both a danger and warning configuration, but only the warning configuration. The changes for the react apps have already been done and merged in this issue: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/669 

#### Screenshot of the result
Screenshot 2023-11-15 at 09 24 12

### DIFF
--- a/web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
+++ b/web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
@@ -87,7 +87,7 @@ class LoanListBlock extends BlockBase implements ContainerFactoryPluginInterface
       "page-size-mobile" => $loanListSettings->get('page_size_mobile') ?? DplLoansSettings::PAGE_SIZE_MOBILE,
 
       // Config.
-      "expiration-warning-days-before" => $generalSettings->get('expiration_warning_days_before_config') ?? GeneralSettingsForm::EXPIRATION_WARNING_DAYS_BEFORE_CONFIG,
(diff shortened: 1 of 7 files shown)


web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
```diff
@@ -98,7 +87,7 @@ public function build() {
       "page-size-mobile" => $loanListSettings->get('page_size_mobile') ?? DplLoansSettings::PAGE_SIZE_MOBILE,
 
       // Config.
-      "threshold-config" => $this->getThresholdConfig(),
+      "expiration-warning-days-before" => $generalSettings->get('expiration_warning_days_before_config') ?? GeneralSettingsForm::EXPIRATION_WARNING_DAYS_BEFORE_CONFIG,
```
spaceo, line 90: You miss the `-config` suffix right?
Shouldn't it be: `expiration-warning-days-before-config`?

spaceo (changes requested):
Looks good to me, but I think you need to correct one thing.

web/modules/custom/dpl_loans/src/Plugin/Block/LoanListBlock.php
Dresse: You are right. This should have been fixed now. :)

spaceo (approved):
LGTM 👍

PR #472, opened 2023-11-15
Add ereolen-homepage-url prop to all dpl-react apps

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-160
https://reload.atlassian.net/browse/DDFLSBP-199

#### Description
Adding url data prop for ereol-homepage for [this PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/654).

#### Screenshot of the result
n/a
n/a

### DIFF
--- a/web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
+++ b/web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
@@ -28,6 +28,7 @@ class GeneralSettingsForm extends ConfigFormBase {
   const REDIRECT_ON_BLOCKED_URL = '';
   const BLOCKED_PATRON_E_LINK_URL = '';
   const EREOLEN_MY_PAGE_URL = '';
+  const EREOLEN_HOMEPAGE_URL = '';
   const PAUSE_RESERVATION_START_DATE_CONFIG = '';
 
   /**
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
```diff
@@ -156,7 +157,14 @@ public function buildForm(array $form, FormStateInterface $form_state): array {
...
       '#description' => $this->t('My page in ereolen', [], ['context' => 'Library Agency Configuration']),
-      '#default_value' => $config->get('ereolen_my_page_url'),
+      '#default_value' => $config->get('ereolen_my_page_url') ?? self::EREOLEN_MY_PAGE_URL,
+    ];
+
+    $form['reservations']['ereolen_homepage_url'] = [
+      '#type' => 'url',
+      '#title' => $this->t('Ereolen home link', [], ['context' => 'Library Agency Configuration']),
+      '#description' => $this->t('Home page in ereolen', [], ['context' => 'Library Agency Configuration']),
+      '#default_value' => $config->get('ereolen_homepage_url') ?? self::EREOLEN_HOMEPAGE_URL,
```
kasperg, line 167: Please consider the configuration here.

I see that it is the responsibility of the CMS to inject these urls but:

1. Is it necessary that they are configurable? I am not sure that Ereolen changes url structures very often.
2. If they are to be configurable I suggest that we provide a sensible default for these values based on the current version of the site.

web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
```diff
@@ -156,7 +157,14 @@ public function buildForm(array $form, FormStateInterface $form_state): array {
       '#type' => 'url',
       '#title' => $this->t('Ereolen link', [], ['context' => 'Library Agency Configuration']),
       '#description' => $this->t('My page in ereolen', [], ['context' => 'Library Agency Configuration']),
-      '#default_value' => $config->get('ereolen_my_page_url'),
+      '#default_value' => $config->get('ereolen_my_page_url') ?? self::EREOLEN_MY_PAGE_URL,
+    ];
```
kasperg, line 161: :+1: Nice of you to remember this default.

kasperg (changes requested):
:+1: Nice work adding this to the CMS. I have a quibble regarding how much we expect libraries to configure on their own for the site to work.

web/modules/custom/dpl_library_agency/src/Form/GeneralSettingsForm.php
Dresse: I also have a small comment, but this is regarding the translations in:

Should the new strings not be added and translated here?

kasperg (approved):
:+1. LGTM

PR #474, opened 2023-11-15
Add reservation-details-config to all apps

spaceo:
#### Description

Since it is a mess what is being added to what and we need the configuration in several apps we add the configuration to all apps in order to be sure that it is not missing.


Dresse (approved):
LGTM

PR #476, opened 2023-11-16
Add reservation-list-digital-pickup-text to patron menu & patron page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-213

#### Description

Turns out the text is in multiple apps.

#### Screenshot of the result
n/a

n/a


web/modules/custom/dpl_patron_page/src/Plugin/Block/PatronPageBlock.php
```diff
@@ -155,6 +155,7 @@ public function build() {
       'patron-page-phone-input-message-text' => $this->t('The phone number must be 6 to 15 characters in length and should be comprised solely of numbers or begin with a +', [], ['context' => 'Patron page']),
       'patron-page-handle-response-information-text' => $this->t('Your changes are saved.', [], ['context' => 'Patron page']),
       'patron-page-loading-text' => $this->t('Loading...', [], ['context' => 'Patron page']),
+      'reservation-list-digital-pickup-text' => $this->t('Online access', [], ['context' => 'Patron page']),
```
kasperg, line 158: Nice to see that you remember to update the context. It would have been easy just to copy/paste the line of code from one place to the other.

kasperg (approved):
:+1: LGTM

Dresse (changes requested):
Looks good, however I also think the translations should be added in:

Should look something like this I reckon?

``msgctxt "Patron menu"
msgid "Online access"
msgstr "Skal hentes online"``

Dresse (approved):

PR #477, opened 2023-11-16
Update and fix documentation for new developers.

rasben:
It was suprisingly painless to setup the site, but I did get a very cryptic error that was fixed by running `docker compose pull` first. So - I've added that to the `dev:reset`

I've also updated the README to the latest, swankiest information

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -94,6 +94,11 @@ tasks:
       # causes the process to fail. Wait and retry if that is the case
       - docker compose {{ .DOCKER_COMPOSE_FILES }} down --volumes --remove-orphans || (sleep 3 && docker-compose down)
 
+  dev:pull:
+    summary: Pull latest docker images.
+    cmds:
+      - docker compose {{ .DOCKER_COMPOSE_FILES }} pull
+
   dev:reset:
     desc: Create local development setup in a clean state
     cmds:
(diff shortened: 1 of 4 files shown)


Taskfile.yml
```diff
@@ -109,6 +114,8 @@ tasks:
       - task dev:cli -- composer reinstall danskernesdigitalebibliotek/dpl-design-system danskernesdigitalebibliotek/dpl-react --no-cache
       # Build dev scripts
       - task dev:cli -- $(cd dev-scripts/dpl-react; composer install)
+      # Pull the images (necessary for the first reset)
+      - task dev:pull
```
kasperg, line 118: I wonder what the error message you saw was? Other developers have set up the project without running into this issue.

Taskfile.yml
```diff
@@ -292,9 +299,9 @@ tasks:
       - task: source:push
 
   lagoon:drush:uli:
-    desc: Get an Drupal one-time login link for a Lagoon environment through Drush
+    desc: Get an Drupal one-time user login link for a Lagoon environment through Drush
     cmds:
-      - task lagoon:cli -- drush uli
+      - task lagoon:cli -- drush user-login
```
kasperg, line 304: :+1: I like it that we use full command names instead of shorthands in scripts.

kasperg (approved):

PR #478, opened 2023-11-17
Fix cookies in cypress

kasperg:
#### Description

Previously we had disabled some Cypress tests because we could not get login to work because of an issue with Cypress.

I took another look at it and as I see it it works again now:

1. I could just reenable a couple of the CMS tests and they ran locally
2. By tailing the Wiremock logs I could see that there was an unmatched request to FBS when logging in with Adgangsplatformen. This causes the login to fail - probably because of #410. I added a mock to fix this. Now tests which require login with Adgangsplatformen also runs locally.

I wonder why this change occurs. Perhaps it has something to do with the upgrade to Cypress 13.5.0 in #468. Anyway it is nice to have more Cypress tests working again.


spaceo (approved):
LGTM 👍

PR #479, opened 2023-11-17
Use ARM-compatible Docker images for better support on new Macs

kasperg:
#### Description

Docker will issue warnings when starting database and Wiremock images because they are not available for ARM architecture.

To address this we update to images which are.

For Wiremock the change matches what we already want to do for the React components in https://github.com/danskernesdigitalebibliotek/dpl-react/pull/632/commits/2cbb10f0c3d165317b17f19039e1f97085c40fde.

For the database image we explicitly have to change to a different image as Lagoon for whatever reason does not provide an ARM-compatible build  for uselagoon/mariadb-drupal. 

Switching to an image that is, uselagoon/mariadb-10.6-drupal, seems to be a good idea, so we are aware of what version of mariadb we are actually using.

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -34,7 +34,7 @@ services:
     # We use wiremock-gui as it exposes a UI for inspecting the state of
     # WireMock. This can make debugging easier. It can seamlessly be replaced
     # with the official wiremock/wiremock image.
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #488, opened 2023-11-21
Drupal 10

kasperg:
#### Link to issue

#### Description

#### Screenshot of the result


rasben (changes requested):
- Everything in `packagess/fbs-client` looks like contrib code, that should not be committed.
- I dont really understand why you add all ignoreErrors in phpstan-baseline.neon. Why not just fix them? legacyConfig() is our own methid anyway?

Otherwise, it looks alright.

rasben (approved):

PR #493, opened 2023-11-24
Enable dev *_ui modules, when running dev:reset.

kasperbirch1:
This also adds it as a standalone command, that we can run after running `drush config-import`

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -138,8 +138,15 @@ tasks:
       - task dev:cache:clear:all
       # Ensure site is reachable and warm any caches
       - task dev:cli -- curl --silent --show-error --fail --output /dev/null http://varnish:8080/
+      # Enable dev modules.
+      - task dev:enable-dev-tools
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -138,9 +138,16 @@ tasks:
...
       - task dev:cli -- curl --silent --show-error --fail --output /dev/null http://varnish:8080/
+      # Enable dev modules.
+      - task dev:enable-dev-tools
       # Show a one-time login to the local site.
       - task dev:cli -- drush user-login
 
+  dev:enable-dev-tools:
+    desc: Enable dev modules and settings, which are not to be used in Prod. They are config-ignored
+    cmds:
+      - task dev:cli -- drush install -y field_ui views_ui
```
kasperg, line 149: In my opinion we could enable all the modules listed in `config_exclude_modules` here.

kasperg (approved):
:+1: I do not know who is the primary developer here but I like it!

PR #497, opened 2023-11-27
Add Byline to `Article` Content-Type

kasperbirch1:
Depend on: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/352
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-9

#### Description
This pull request introduces a byline feature to the article content-type. The byline automatically includes the author's name and the date of creation, drawing these details from the user who has written the article. Additionally, this feature offers the flexibility to replace the default author link with a custom text if desired.

#### Screenshot of the result
image
image

**Override author**
image
image


web/modules/custom/dpl_article/dpl_article.module
```diff
@@ -0,0 +1,55 @@
...
+use Drupal\node\Entity\Node;
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function dpl_article_preprocess_node(array &$variables): void {
+  $node = $variables['node'] ?? NULL;
+
+  // Process only if the node is a valid Node entity.
+  if (!($node instanceof Node)) {
```
kasperg, line 18: Please replace not with `NodeInterface`.

In general when doing object oriented programming you often want to code against interfaces instead of implementations.

web/modules/custom/dpl_article/dpl_article.info.yml
```diff
@@ -0,0 +1,5 @@
+name: "dpl_article"
```
kasperg, line 1:
```suggestion
name: "Article"
```

The name of the module is what will be shown to administrators in the module list. Using a human readable name is always nice.

config/sync/core.date_format.article_author_date.yml
```diff
@@ -0,0 +1,8 @@
+uuid: cbff7fb1-23be-469b-a2c1-ae834ec4c55e
+langcode: en
+status: true
+dependencies: {  }
+id: article_author_date
+label: 'Article author date'
+locked: false
+pattern: 'd. F y'
```
kasperg, line 8: Consider if it is a good idea to provide a date format with a very specific name which implies a specific use case.

In the design do we see dates with the same granularity (day, month, year) presented differently in other contexts. If so then that might imply yes. If not then we should probably create something with a name that invites reuse.

If in doubt here then talk to a colleague.

web/modules/custom/dpl_article/dpl_article.info.yml
```diff
@@ -0,0 +1,5 @@
...
+type: module
+description: "Functionality related to creating article content."
+package: "DPL"
+core_version_requirement: ^9 || ^10
```
kasperg, line 5:
```suggestion
core_version_requirement: ^10
```

Now that we are on Drupal 10 and do not expect to downgrade we might as well specify for the future.

config/sync/field.field.node.article.field_show_overwrite_author.yml
```diff
@@ -0,0 +1,23 @@
...
+dependencies:
+  config:
+    - field.storage.node.field_show_overwrite_author
+    - node.type.article
+id: node.article.field_show_overwrite_author
+field_name: field_show_overwrite_author
+entity_type: node
+bundle: article
+label: 'Overwrite the author'
+description: "By default, the author is set to the Drupal user that owns the content.<br>\r\nIf you want to override this, with a manual text, you can check this."
```
kasperg, line 13: Please update this field to use [override instead of overwrite](https://stackoverflow.com/questions/8651562/overwrite-or-override).

As I see it this description gets it right, the label (and field id) gets it wrong.

kasperg (changes requested):
I took a look here.

Good job but there is still some work to do.

web/themes/custom/novel/templates/fields/username.html.twig
```diff
@@ -0,0 +1,5 @@
+{% if link_path -%}
+  <a{{ attributes.addClass('link-tag', 'text-body-small-regular') }}>{{ name }}{{ extra }}</a>
```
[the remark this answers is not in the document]
kasperbirch1: as I see it, it is necessary because `link-tag' has a different font size than `text-body-small-regular'

config/sync/core.entity_form_display.node.article.default.yml
```diff
@@ -16,29 +17,57 @@ targetEntityType: node
...
+    type: boolean_checkbox
+    weight: 27
     region: content
-  field_show_overwrite_author:
-    weight: 3
     settings:
       display_label: true
     third_party_settings: {}
-    type: boolean_checkbox
+  field_paragraphs:
```
kasperg, line 35: This seems to be new. Is it something that someone else forgot to commit?

config/sync/core.entity_form_display.node.article.default.yml
```diff
@@ -3,8 +3,9 @@ langcode: en
 status: true
 dependencies:
   config:
-    - field.field.node.article.field_overwrite_author
-    - field.field.node.article.field_show_overwrite_author
+    - field.field.node.article.field_override_author
```
kasperg, line 6: :+1: I appreciate you taking the time to change this.

web/themes/custom/novel/templates/layout/node--article--full.html.twig
```diff
@@ -1,10 +1,18 @@
...
+
+		<p class="article-header__info">
+			{{ 'By @username'|t({'@username':  author_name}) }}
+
+			<time class="article-header__date">
+				{{ node.created.value|format_date('date_full_month') }}
+			</time>
+		</p>
+
+		{{ content|without('field_subtitle') }}
```
kasperg, line 15: Content is already printed on L18. Did you mean for this line to replace what is currently on L18?

kasperg (approved):
:+1: LGTM

PR #506, opened 2023-11-28
Feature/ddflsbp 194/mapp id should be have default value

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-194

#### Description

The ID field in MAPP settings, should always be filled with the value "476651662471322". Right now, each library have to go in and add that value. This PR adds the value as part of the configuration in dpl_mapp.settings.yml.

#### Screenshot of the result

Screenshot 2023-11-28 at 23 12 35

N/A


spaceo (approved):
LGTM 👍

PR #515, opened 2023-11-30
XDebug improvements

kasperg:
#### Description

This PR adds a task to enable XDebug temporarily for the PHP container and documents it.

[According to the documentation XDebug is only enabled if the `XDEBUG_ENABLE` environment variable for the PHP container is true](https://docs.lagoon.sh/using-lagoon-advanced/setting-up-xdebug-with-lagoon/
). Having to uncomment a line is a bit inelegant as it requires file changes and hard to discover.

Instead of doing this we instead allow `XDEBUG_ENABLE` to be set from the outside and fall back to `false` if not. This ensures that XDebug is still disabled by default.

Now we can create a new task which sets `XDEBUG_ENABLE` to `true` while recreating the containers and waiting for further input.

When the developer is done containers are recreated without the environment variable set and thus without XDebug enabled.

You should be able to test whether this works by following the documentation :smile:

The PR environment does not have XDebug enabled which should ensure that this only affects local environments.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -157,6 +157,12 @@ tasks:
     cmds:
       - task dev:cli -- drush install -y field_ui views_ui
 
+  dev:enable-xdebug:
+    cmds:
+      - XDEBUG_ENABLE=true task dev:start
+      - read -p "Press enter to disable Xdebug"
+      - task dev:start
+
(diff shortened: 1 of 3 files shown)


kasperbirch1 (approved):
Sounds good

PR #517, opened 2023-12-01
DDFLSBP-274 - Added view-fees-and-compensation-rates url to dpl_react…

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-274

#### Description

Added missing view-fees-and-compensation-rates url to dpl_react_apps_data.

#### Screenshot of the result

N/A

N/A

### DIFF
--- a/web/modules/custom/dpl_react_apps/dpl_react_apps.info.yml
+++ b/web/modules/custom/dpl_react_apps/dpl_react_apps.info.yml
@@ -8,6 +8,7 @@ dependencies:
   - dpl_library_agency:dpl_library_agency
   - dpl_instant_loan:dpl_instant_loan
   - dpl_dashboard:dpl_dashboard
+  - dpl_fees:dpl_fees
 
 type: module
 core_version_requirement: ^9


web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -11,6 +11,7 @@
 
 use Drupal\Core\GeneratedUrl;
 use Drupal\Core\Url;
+use Drupal\dpl_fees\DplFeesSettings;
```
spaceo, line 14: Since you are introducing dpl_fees settings in the dpl_react_apps module you need to add the dpl_fees module as a dependency [here](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/a6de7e5518445fae08658aba3bc41a915a8d3220/web/modules/custom/dpl_react_apps/dpl_react_apps.info.yml#L11).

spaceo (changes requested):
Looks good. However one thing needs to be done

spaceo (approved):
LGTM 👍

PR #523, opened 2023-12-04
Add missing translations to reservation list block

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-325

#### Description

This PR adds two missing translations to the reservation list block:
reservation-list-pause-reservation-button-text
reservation-list-pause-reservation-on-hold-text

#### Screenshot of the result
n/a

n/a


web/modules/custom/dpl_reservations/src/Plugin/Block/ReservationListBlock.php
JacobArrow, line 1: We should also add `pauseReservationModalCancelButtonLabelText` which was introduced here #681

Adamik10: yessss thank youuu!! :)))))

JacobArrow (approved):
SantaSantaClausGIF

PR #534, opened 2023-12-07
Make sure that every build has unique assets.

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-328

#### Description

By setting the build name as an ARG in cli.docker the following layers are not cached and therefor composer install is run everytime.

That ensures that the composer dependencies always are fresh 🥗...

### DIFF
--- a/lagoon/cli.dockerfile
+++ b/lagoon/cli.dockerfile
@@ -1,5 +1,9 @@
 FROM uselagoon/php-8.1-cli-drupal:latest
 
+# Make sure that every build has unique assets.
+# By setting the build name as an ARG the following layers are not cached.
+ARG LAGOON_BUILD_NAME
+
 COPY composer.* /app/
 COPY assets /app/assets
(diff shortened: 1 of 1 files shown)


kasperg (approved):
Really great that you have found a solution to this problem. I would really like to see more elaborate commenting. It might be a 1 line fix but it is a problem that is somewhat unique to this project and have haunted us for months if not years.

I would like to get a summary of:

- What is the problem for us? Something about continually updated branch releases which do not require changes to composer.* and thus enabling Docker layer caching.
- How we solved it? A unique build argument to bust the cache
- How did we come about the actual solution? As far as I can tell [`LAGOON_BUILD_NAME` is not mentioned in the documentation](https://docs.lagoon.sh/using-lagoon-the-basics/build-and-deploy-process/) so what are we going into here? A reference to Lagoon source code would be helpful.

I think the code comment is OK but the commit comment needs to be expanded.

PR #538, opened 2023-12-09
Event price

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-57

#### Description

This PR adds handling of ticket categories and prices. The primary changes are:

- Adding a new paragraph "Ticket category" with fields for name (string) and price (decimal)
- Implementing logic for formatting prices in a new service
- Rendering prices as a part of the event metadata

Please check out the individual commits for additional commentary.

#### Screenshot of the result

Monosnap Quam non enim obcaecati iste | DPL CMS 2023-12-11 09-18-04

In the process I found that we need to tweak our PHPStan setup to avoid false positives when rendering fields and altering forms.

### DIFF
--- a/config/sync/field.field.paragraph.event_ticket_category.field_ticket_category_price.yml
+++ b/config/sync/field.field.paragraph.event_ticket_category.field_ticket_category_price.yml
@@ -1,0 +1,23 @@
+uuid: 5bf5715f-f899-4364-83e9-087733d804e0
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.paragraph.field_ticket_category_price
+    - paragraphs.paragraphs_type.event_ticket_category
+id: paragraph.event_ticket_category.field_ticket_category_price
+field_name: field_ticket_category_price
+entity_type: paragraph
+bundle: event_ticket_category
+label: Price
+description: ''
+required: true
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  min: 0.0
+  max: null
+  prefix: DKK
+  suffix: ''
+field_type: decimal
--- a/config/sync/field.field.paragraph.event_ticket_category.field_ticket_category_price.yml
+++ b/config/sync/field.field.paragraph.event_ticket_category.field_ticket_category_price.yml
@@ -18,6 +18,6 @@ default_value_callback: ''
 settings:
   min: 0.0
   max: null
-  prefix: DKK
+  prefix: ''
   suffix: ''
 field_type: decimal
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -5,6 +5,7 @@
  * Module file for DPL Event.
  */
 
+use Drupal\Core\Form\FormStateInterface;
 use Drupal\dpl_event\EventState;
 use Drupal\job_scheduler\Entity\JobSchedule;
 use Drupal\node\NodeInterface;
@@ -116,6 +117,15 @@ function dpl_event_set_occurred_callback(JobSchedule $job): void {
     $event->set("field_event_state", EventState::Occurred);
     $event->save();
   }
+}
+
+/**
+ * Implements hook_form_node_event_form_alter().
+ */
+function dpl_event_form_node_event_form_alter(array &$form, FormStateInterface $form_state, string $form_id): void {
+  // Use the generic "Access" value as the default for the first ticket
+  // category when creating events.
+  $form["field_ticket_categories"]["widget"][0]["subform"]["field_ticket_category_name"]["widget"][0]["value"]["#default_value"] = t('Access', [], ['context' => 'dpl_event']);
 }
 
 /**
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -172,3 +172,19 @@ function dpl_event_preprocess_field(array &$variables): void {
     $variables['items'][0]['content']['#title'] = $event_state->label();
   }
 }
+
+/**
+ * Implements hook_preprocess_field__field_ticket_category_price().
+ *
+ * Adds an array of formatted prices.
+ */
+function dpl_event_preprocess_field__field_ticket_category_price(array &$variables) : void {
+  /** @var \Drupal\dpl_event\PriceFormatter $price_formatter */
+  $price_formatter = \Drupal::service('dpl_event.price_formatter');
+  /** @var \Drupal\Core\Field\FieldItemListInterface $items */
+  $items = $variables['element']['#items'];
+  $prices = array_map(function (array $price) use ($price_formatter): string {
+    return $price_formatter->formatPrice($price["value"]);
+  }, $items->getValue());
+  $variables['prices'] = $prices;
+}
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -125,7 +125,11 @@ function dpl_event_set_occurred_callback(JobSchedule $job): void {
 function dpl_event_form_node_event_form_alter(array &$form, FormStateInterface $form_state, string $form_id): void {
   // Use the generic "Access" value as the default for the first ticket
   // category when creating events.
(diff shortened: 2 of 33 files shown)


web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -117,3 +118,28 @@ function dpl_event_set_occurred_callback(JobSchedule $job): void {
...
   }
 }
+
+/**
+ * Implements hook_form_node_event_form_alter().
+ */
+function dpl_event_form_node_event_form_alter(array &$form, FormStateInterface $form_state, string $form_id): void {
+  // Use the generic "Access" value as the default for the first ticket
+  // category when creating events.
+  $form["field_ticket_categories"]["widget"][0]["subform"]["field_ticket_category_name"]["widget"][0]["value"]["#default_value"] = t('Access', [], ['context' => 'dpl_event']);
```
rasben, line 128: You need to check that this array key already exists, and only set the value if it does.

Also, you match on `[0]` - can there sometimes be more than one item?

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -0,0 +1,44 @@
...
+ * have double fields with values as strings.
+ */
+class PriceFormatter {
+
+  /**
+   * Constructor.
+   */
+  public function __construct(
+    protected TranslationInterface $translation,
+  ) {}
```
rasben, line 21: isnt it a mistake that this is empty?
shouldnt it contain something like

```php
$this->translation = $translation;
```
?

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Format a single price.
+   */
+  public function formatPrice(string $price_string): string {
+    $context = ['context' => 'dpl_event'];
+
+    $price = BigDecimal::of($price_string);
+    return match(TRUE) {
```
rasben, line 30: Uh..? Why do you use `match` with a hardcoded value?

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -0,0 +1,44 @@
...
+   */
+  public function __construct(
+    protected TranslationInterface $translation,
+  ) {}
+
+  /**
+   * Format a single price.
+   */
+  public function formatPrice(string $price_string): string {
+    $context = ['context' => 'dpl_event'];
```
rasben, line 27: either rename it `$translation_context`, or move the array into the `translate()` as a parameter

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -0,0 +1,44 @@
...
+
+    $price = BigDecimal::of($price_string);
+    return match(TRUE) {
+      // Events with 0 cost should show "Free" instead of a numeric price.
+      $price->isEqualTo(0) => $this->translation->translate("Free", [], $context),
```
rasben, line 32: does this also work if someone has written 0.00?

web/themes/custom/novel/templates/fields/field--node--field-ticket-categories--event.html.twig
```diff
@@ -0,0 +1,8 @@
+{% if items|length == 1 %}
+  {# Events with a single ticket category should only show the price. #}
+  <span>{{ items.0.content['#paragraph'].field_ticket_category_price|view }}</span>
```
rasben, line 3: i think you can just do

```suggestion
  <span>{{ entity_type. field_ticket_category_price.0 }}</span>
```

rasben: or, alternatively: (This is just mock code, it probably doesnt work, but i hope you get the idea)

```
{% set view_mode = (items|length == 1) ? 'default' : 'single' %}


  {% for item in items %}
    {{ item.content|view(view_mode) }}
  {% endfor %}

**Review by rasben — CHANGES_REQUESTED** [review]

File: config/sync/field.field.paragraph.event_ticket_category.field_ticket_category_price.yml [hunk 7]
```diff
@@ -0,0 +1,23 @@
...
+entity_type: paragraph
+bundle: event_ticket_category
+label: Price
+description: ''
+required: true
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  min: 0.0
```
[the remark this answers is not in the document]
kasperg: This is the result of setting a minimum value of 0. Drupal for whatever reason stores this as 0.0 for a decimal value.

The actual formatting is handled by the `PriceFormatter` instance. The test shows that numbers with be formatted with two decimals even if an editor only inputs it with a single decimal.

web/modules/custom/dpl_event/src/PriceFormatter.php
kasperg: This is intentional [constructor property promotion](https://php.watch/versions/8.0/constructor-property-promotion).

web/modules/custom/dpl_event/src/PriceFormatter.php
kasperg: Because `match` needs a boolean input but [you can still use it for non-identity checks](https://www.php.net/manual/en/control-structures.match.php#weakreference.basic-example). It is basically another way of doing `if+elseif+else`.

web/modules/custom/dpl_event/src/PriceFormatter.php
kasperg: Yes it does, because `BigDecimal` handles this complexity for os. The unit test also shows that this is the case.

web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -117,3 +118,28 @@ function dpl_event_set_occurred_callback(JobSchedule $job): void {
...
+
+/**
+ * Implements hook_preprocess_field__field_ticket_category_price().
+ *
+ * Adds an array of formatted prices.
+ */
+function dpl_event_preprocess_field__field_ticket_category_price(array &$variables) : void {
+  /** @var \Drupal\dpl_event\PriceFormatter $price_formatter */
+  $price_formatter = \Drupal::service('dpl_event.price_formatter');
+  /** @var \Drupal\Core\Field\FieldItemListInterface $items */
```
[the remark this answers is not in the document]
kasperg: I do not think that is a good idea.

If we add an`instanceof`check this means that if a developer updates services definitions, then the code will still work technically but the correct formatting will not be used. It is kind of a silent failure. Since a change in services is initiated by a developer then it should also be easy for developers to become aware of any problems a change might cause. This will indeed be the case if rendering the event page fails. This is also be easy for any automated test to catch.

For user input it is another matter :smile:.

web/modules/custom/dpl_event/src/PriceFormatter.php
kasperg: I opted for `$translation_options` as that is what the variable is used for. It contains the context.

web/themes/custom/novel/templates/fields/field--node--field-ticket-categories--event.html.twig
kasperg: The first approach does not work. It will just render the first ticket category with both name and price.

I get the idea of your suggestion about introducing a new `single` view mode but it gets quite specialized. It will also force us to consider this view mode for other paragraphs where I do not think it will be relevant.

web/modules/custom/dpl_event/dpl_event.module
kasperg: I updated this with a check that fails hard if the element cannot be located.

This is based on the same reasoning as above.

web/modules/custom/dpl_event/src/PriceFormatter.php
rasben: To be completely honest, I feel like this is a very complicated way of doing something, that is very hard to read/maintain.

Maybe it's just cause I've never used that function before, but it seems super abstract.
I get that it might be less code than a more simple if/else, but is it worth it?

I'll leave the decision up to you - I'll approve regardless :)

rasben (approved):
Approved, but with a small final comment.
I'll leave it up to you if you want to revisit it or not.

PR #542, opened 2023-12-10
Create 'medias' paragraph, that display 1-2 irregular medias. DDFFORM-73

rasben:
- Creating a 'medias' paragraph, that allows two media images.
- Create a custom dpl_paragraphs module, that alters any second
  media to have an alternate image style.
- Clean-up and streamline the media styles, to only match what
  we actually use.
- Add templates that sets correct HTML tags for figure/captions.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-73


kasperbirch1 (approved):
LGTM 👍

PR #546, opened 2023-12-11
Add `place` field to `event` content type

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-18

#### Description
This pull request adds a `place` field to the event content type. The field is currently implemented as a multi-string, but it will be replaced with a checkbox list that includes all libraries in the future. 

#### Screenshot of the result
image

### DIFF
--- a/web/themes/custom/novel/templates/layout/node--event--full.html.twig
+++ b/web/themes/custom/novel/templates/layout/node--event--full.html.twig
@@ -20,6 +20,16 @@
       </div>
     </div>
     <dl class="list-description event-description__info list-description--event">
+      {% if content.field_event_place.0 %}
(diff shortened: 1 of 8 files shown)


web/themes/custom/novel/templates/layout/node--event--full.html.twig
```diff
@@ -20,6 +20,16 @@
       </div>
     </div>
     <dl class="list-description event-description__info list-description--event">
+      {% if content.field_event_place.0 %}
+      <div class="list-description__item">
+        <dt class="list-description__key">
+          {{ "Place" | trans }}
```
rasben, line 26: You need to add a translation for this.

PR #559, opened 2023-12-14
Allow editors to choose focal points for image-cropping. DDFFORM-72

rasben:
- Using the 'focal_points' module, which also comes with a preview functionality, that allows the editor to see which image styles will be created, based on the crop.
- Also includes 'media_library_edit' module, to improve the editor experience when editing content.
- A few minor admin CSS tweaks, to make the focal point preview play nice in the GIN admin theme.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,6 +90,7 @@
         "drupal/devel": "~5.0.2",
         "drupal/enum_field": "^1.0",
         "drupal/field_group": "^3.4",
+        "drupal/focal_point": "^2.0",
         "drupal/gin": "^3.0@RC",
         "drupal/gin_toolbar": "^1.0@RC",
         "drupal/handy_cache_tags": "^1.4",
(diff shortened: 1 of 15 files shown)


config/sync/field.field.media.image.field_media_image.yml
```diff
@@ -22,15 +22,15 @@ settings:
   handler_settings: {  }
   file_directory: '[date:custom:Y]-[date:custom:m]'
   file_extensions: 'png gif jpg jpeg'
-  max_filesize: ''
+  max_filesize: 256mb
```
kasperg, line 25: Consider why it is necessary to set this value.

- 256MB seems quite high for a maximum file size
- I would not spot any reasoning in the commit comment.

composer.json
```diff
@@ -212,6 +214,9 @@
                 "Configure core to not check permissions before accepting an updated module": "patches/disable-core-upload-permission-check.diff",
                 "3073554: Add a token for publication status": "https://www.drupal.org/files/issues/2019-08-09/3073554_10.patch"
             },
+            "drupal/focal_point": {
+                "Preview link accidentally closes the media library\n": "https://www.drupal.org/files/issues/2020-10-11/preview_link_accidentally_closes_the_media_library-3162210-19.patch"
```
kasperg, line 218: Please include the node id for the patch in the title like we do for the other external patches.

Is it intentional that you include a newline `\n`?

kasperg (approved):
:+1: LGTM

I have a couple of comments that you may want to consider before you merge.

PR #576, opened 2023-12-19
Add paragraph field to event and enable the link paragraph

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-94

#### Description
This pull request adds a new 'paragraph' field to our event content type. This feature enables the use of reusable paragraphs, enhancing content flexibility across different types. Currently, this update focuses on integrating 'link and the text body paragraphs', which are essential for our events.

#### Screenshot of the result
image

### DIFF
--- a/config/sync/core.entity_form_display.node.event.default.yml
+++ b/config/sync/core.entity_form_display.node.event.default.yml
@@ -8,6 +8,7 @@ dependencies:
     - field.field.node.event.field_event_description
     - field.field.node.event.field_event_image
     - field.field.node.event.field_event_link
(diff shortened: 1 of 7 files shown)


kasperg (changes requested):
I think It looks good but it could use a bit of extra configuration.

Try to compare the editorial UI for adding paragraphs to articles:

billede

With events:

Monosnap Create Event | DPL CMS | Logged in 2023-12-20 10-07-47

Try to look into the configuration and ensure that the one for events matches the one from articles so we get a consistent approach.

kasperg (approved):

PR #579, opened 2023-12-19
DDFLSBP-362 - Changed default_interest_period_config to a single value

Dresse:
Link to issue
https://reload.atlassian.net/browse/DDFLSBP-362

Description
The configuration for default_interest_period_config was unusable as it was formatted as an "stringformatted" array. But as the default_interest_period field is a select field, it would only need a value. Now the default_interest_period_config has been formatted correctly to only contain that value.


spaceo (approved):
LGTM 👍

PR #583, opened 2023-12-20
Dedupe shared react app data attributes

spaceo:
#### Description

Although adding load to the rendering process we can cut away a substantial amount of DOM rendering by removing data attributes from the other apps that has already been used.

We can do this because the react apps shares the data of some of the data attributes in a shared store.

### DIFF
--- a/web/modules/custom/dpl_react/dpl_react.module
+++ b/web/modules/custom/dpl_react/dpl_react.module
@@ -10,10 +10,10 @@
 
 use Drupal\Component\Serialization\Json;
 use Drupal\dpl_react\DplReactInterface;
-
-use function Safe\file_get_contents as file_get_contents;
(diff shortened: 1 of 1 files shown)


kasperg (commented):
Without further arguments I am not a fan of this approach.

It was not the intention that data attributes set on one app should be usable in another app. To me this adds a new level of complexity where you have to look for data attributes in all apps to figure out what is actually available in the individual app.

I do not see the need for this complexity. We have not faced any actual issues regarding the size of our DOM yet and we have gone pretty far already.

spaceo:
Agreeing in the arguments against this soultion.
The motivation for the changes was driven partly by seeing how much unecessary DOM that is being produced and partly as an experiment if it was possible to dedupe the data :)

PR #585, opened 2023-12-21
Introduce `dpl_identity` module for identity color customization

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-100

#### Description
This pull request introduces the `dpl_identity module, designed to enable users to choose a preferred identity color.

Upon selecting a color, the module dynamically inserts a style tag into the HTML head. This action effectively overrides the predefined style variables, allowing for customized color schemes.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-cms/assets/49920322/09cbf5d4-c76c-4090-8e9c-bdf54c23f9ed

### DIFF
--- a/web/modules/custom/dpl_identity/dpl_identity.routing.yml
+++ b/web/modules/custom/dpl_identity/dpl_identity.routing.yml
@@ -1,7 +1,7 @@
-dpl_identity.color:
-  path: "/admin/config/dpl_identity"
+dpl_identity.settings:
+  path: "/admin/config/dpl_identity/settings"
   defaults:
     _form: '\Drupal\dpl_identity\Form\IdentityConfigForm'
-    _title: "Identity Color"
+    _title: "Identity Settings"
   requirements:
     _permission: "administer site configuration"
--- a/web/modules/custom/dpl_identity/dpl_identity.routing.yml
+++ b/web/modules/custom/dpl_identity/dpl_identity.routing.yml
@@ -1,0 +1,7 @@
+dpl_identity.color:
+  path: "/admin/config/dpl_identity"
+  defaults:
+    _form: '\Drupal\dpl_identity\Form\IdentityConfigForm'
+    _title: "Identity Color"
+  requirements:
+    _permission: "administer site configuration"
--- a/web/modules/custom/dpl_identity/src/Form/IdentityConfigForm.php
+++ b/web/modules/custom/dpl_identity/src/Form/IdentityConfigForm.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace Drupal\dpl_identity\Form;
+
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines a form for setting identity configuration.
+ */
+class IdentityConfigForm extends FormBase {
+
+  /**
+   * The config factory service.
+   *
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $configFactory;
+
+  /**
+   * Constructs a new IdentityConfigForm object.
+   *
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config factory service.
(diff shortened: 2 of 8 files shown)


web/modules/custom/dpl_identity/src/Form/IdentityConfigForm.php
```diff
@@ -0,0 +1,94 @@
...
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   *
+   * @return array
+   *   The form structure.
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
```
kasperg, line 74: Please move the return value type to the function signature and leave out this parameter documentation.

web/modules/custom/dpl_identity/src/Form/IdentityConfigForm.php
```diff
@@ -0,0 +1,94 @@
...
+    return 'dpl_identity_config_form';
+  }
+
+  /**
+   * Form submission handler.
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
```
kasperg, line 54: Please leave out this parameter documentation.

In the current situation the documentation does not provide any value compared to what is already better defined in the function signature.

web/modules/custom/dpl_identity/src/Form/IdentityConfigForm.php
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    // Load existing configuration for the form.
+    $config = $this->configFactory->get('dpl_identity.settings');
+
+    $form['identity_color'] = [
+      '#type' => 'color',
+      '#title' => $this->t('Identity Color'),
+      '#default_value' => $config->get('identity_color') ?? '#365342',
+      '#description' => $this->t('Choose library identity color.'),
```
kasperg, line 82: Please [add a readable version of the module name as a context to translated strings](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/45fb37c42adf3e77157033b44509d100d7d83e0b/web/modules/custom/dpl_url_proxy/src/Form/ProxyUrlConfigurationForm.php#L135-L135). 

This makes it easier for translators to know the original of the texts they are working on.

web/modules/custom/dpl_identity/dpl_identity.routing.yml
```diff
@@ -0,0 +1,7 @@
+dpl_identity.color:
+  path: "/admin/config/dpl_identity"
+  defaults:
+    _form: '\Drupal\dpl_identity\Form\IdentityConfigForm'
+    _title: "Identity Color"
```
kasperg, line 5: Consider whether this form should be solely focused on colors or whether we expect it to expand into other areas in the future.

If so then color should probably be removed from much of all this configuration.

It is notably not part of the `IdentityConfigForm` class name.

kasperg (changes requested):
:+1: Nice first stab at this. I have some comments here and there.

Notably this is configuration which should be handled by config ignore.

I think it is worth mentioning this in [the ticket about reinstating config ignore](https://reload.atlassian.net/browse/DDFFORM-82).

web/modules/custom/dpl_identity/src/Form/IdentityConfigForm.php
```diff
@@ -0,0 +1,62 @@
...
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    // Load existing configuration for the form.
+    $config = $this->config('dpl_identity.settings');
+
+    $form['identity_color'] = [
+      '#type' => 'color',
+      '#title' => $this->t('Identity Color', [], ['context' => 'Identity settings']),
+      '#default_value' => $config->get('identity_color'),
+      '#description' => $this->t('Choose library identity color.'),
```
kasperg, line 42: Remember to add the context both places.

kasperg (approved):
:+1: LGTM. 

One final thing needs to be changed before this is merged.

PR #586, opened 2023-12-21
DDFFORM 59 arrangement liste

LasseStaus:
#### Link to issue

[jira link](https://reload.atlassian.net/browse/DDFFORM-59)
[design system PR](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/389)

#### Description

This PR
- Adds search_api module, a server and index for events. 
- View for events, and a list-teaser view mode that will display the events in a list. 
- Relevant wrapper templates and preprocess for serving calculated ticket-prices for event-list-teaser items. 
- Adds module date_range_formatter for formatting dates. 
- Refactors the way dates were displayed for a single event. (event full content / single page)

Currently the list only supposed to show events that have not yet completed. As such the filter is configured to the date.end_date offset to 00:00. 

The index is also set to an events end_date.

You will have to create events manually to see that it works as intended. Please specifically check that prices are calculated & displayed correctly. 

Go commit by commit

#### Screenshot of the result

image

I haven't worked a lot with views, so be observant whether i have missed something in the configuration / the way the view is structured that is relevant. 

I.e. is the path correctly set, any settings i have missed etc.


web/themes/custom/novel/novel.theme
```diff
@@ -86,3 +86,71 @@ function novel_preprocess_dpl_react_app__search_result(array &$variables): void
...
+
+    // Collect prices from ticket categories.
+    foreach ($ticket_categories as $category) {
+      if ($category->hasField('field_ticket_category_price') && !$category->get('field_ticket_category_price')->isEmpty()) {
+        $price = $category->get('field_ticket_category_price')->value;
+        $prices[] = (floatval($price) == intval($price)) ? intval($price) : floatval($price);
+      }
+    }
+
+    $variables['ticket_price_display'] = novel_get_ticket_price_display($prices);
```
kasperg, line 115: Please reconsider all this price handling logic in the theme.

We already have [a class for handling price formatting](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/b47d1118ca6c79a5c062af5c41ba5d08d321cc4f/web/modules/custom/dpl_event/src/PriceFormatter.php#L14-L14). It handles formatting single prices. I think we should extend it to handle an array of prices. This provides a number of benefits:

  - The class uses `BigDecimal` to handle juggling of floats and ints for you.
  - You have a translator available so you can translate strings. The current ones in the code are hardcoded to Danish.
  - We can implement a unit test of the code by extending the current test. This makes it a lot easier to verify that the code it working as intended
  - This moves the code away from the template. I have previously heard @rasben request that we move logic to modules instead of the `.theme` file where possible.

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
```diff
@@ -0,0 +1,31 @@
...
+      {% for item in content.field_event_place['#items'] %}
+          <p class="event-list-item__location">{{ item.value }}</p>
+      {% endfor %}
+    </div>
+    <div class="event-list-item__schedule">
+        <time class="event-list-item__time">
+          {{ content.field_event_date['#items'].0.start_date|date("H:i") }}
+          -
+          {{ content.field_event_date['#items'].0.end_date|date("H:i") }}
+        </time>
```
kasperg, line 24: Please remember to add the [`datetime` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#valid_datetime_values) for `<time>` elements.

The actual time for the event is spread across two elements here - the date(s) and the time. Perhaps it is more relevant to add it to the event date field.

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
```diff
@@ -0,0 +1,31 @@
...
+      {% set image_uri = content.field_event_image['#items'].entity.field_media_image.entity.fileuri %}
+      {% set image_url = file_url(image_uri) %}
+        <img src="{{ image_url }}" alt="{{ content.label }}" class="event-list-item__image" />
+    {% endif %}
+  </div>
+  <div class="event-list-item__content">
+    <div aria-pressed="false" class="tag tag--fill cursor-pointer tag--small event-list-item__tag">foredrag</div>
+    <p class="event-list-item__date">{{ content.field_event_date|render|striptags }}</p>
+    <h3 class="event-list-item__title">{{ label }}</h3>
+    <p class="event-list-item__description">{{ content.field_event_description|render|striptags|trim }}</p>
```
kasperg, line 13: Please reconsider the `render|striptags|trim` construct.

I guess you have added this to avoid wrapper divs around the field content. I would prefer that we use a different approach:

1. Override the field template to remove the wrapper divs
2. Get the raw value from the entity

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
```diff
@@ -0,0 +1,31 @@
+<a href="{{ content.event_link }}" class="event-list-item arrow__hover--right-small">
+  <div class="event-list-item__image-container">
+    {% if content.field_event_image['#items'] is not empty %}
+      {% set image_uri = content.field_event_image['#items'].entity.field_media_image.entity.fileuri %}
+      {% set image_url = file_url(image_uri) %}
+        <img src="{{ image_url }}" alt="{{ content.label }}" class="event-list-item__image" />
+    {% endif %}
+  </div>
+  <div class="event-list-item__content">
+    <div aria-pressed="false" class="tag tag--fill cursor-pointer tag--small event-list-item__tag">foredrag</div>
```
kasperg, line 10: Please remove `aria-pressed="false"`.

I do not think it makes sense in this context as the element is no longer a `button`.

kasperg, line 10: Consider removing this element entirely.

We have not implemented the correct value to display here so we might as well leave it out if we do not have something to display. The primary reason for leaving it in would be if keeping it out meant that the layout would be messed up.

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
```diff
@@ -0,0 +1,31 @@
+<a href="{{ content.event_link }}" class="event-list-item arrow__hover--right-small">
+  <div class="event-list-item__image-container">
+    {% if content.field_event_image['#items'] is not empty %}
+      {% set image_uri = content.field_event_image['#items'].entity.field_media_image.entity.fileuri %}
+      {% set image_url = file_url(image_uri) %}
+        <img src="{{ image_url }}" alt="{{ content.label }}" class="event-list-item__image" />
+    {% endif %}
+  </div>
```
kasperg, line 8: Please reconsider this construct.

I think I see what you want to do but I think this belongs in the template for the media or file entity if possible.

We need to get the right classes into the template. That might be tricky. Perhaps @rasben can help here or we can look at it together.

One problem which seems to be a side effect of the current approach is that you use the title of the event as the `alt` text instead of the alt text for the media entered by the editor.

web/themes/custom/novel/templates/views/views-view--events.html.twig
```diff
@@ -0,0 +1,40 @@
...
+  dom_id ? 'js-view-dom-id-' ~ dom_id,
+  'event-list-page'
+]
+%}
+<div{{ attributes.addClass(classes) }}>
+  {{ title_prefix }}
+  {{ title }}
+  {{ title_suffix }}
+
+  <h1 class="event-list-page__heading">Arrangementer</h1>
```
kasperg, line 12: Please remember to the default text in English and make it translatable.

config/sync/core.entity_view_display.node.event.full.yml
```diff
@@ -33,11 +33,16 @@ content:
...
+    type: date_range_without_time
     label: hidden
     settings:
       timezone_override: ''
+      date_format: 'Y-m-d\TH:i:s'
       separator: '-'
+      one_day: 'd. M Y'
+      one_month: 'd. - {d}. M Y'
+      several_months: 'd. M - {d}. {M} Y'
+      several_years: 'd. M Y - {d}. {M} {Y}'
```
kasperg, line 45: :+1: for taking the effort of refactoring the way we show event dates on the full event after introducing `date_range_formatter`.

This removes the need of a large chuck of code currently nested in a template. This is much better.

config/sync/views.view.events.yml
```diff
@@ -0,0 +1,229 @@
...
+        - 'config:search_api.index.events'
+        - 'search_api_list:events'
+  page_1:
+    id: page_1
+    display_title: Page
+    display_plugin: page
+    position: 1
+    display_options:
+      display_extenders: {  }
+      path: Arrangementer
```
kasperg, line 218: Please change the path to something in English e.g `events`.

The site is in english by default. We will use path aliases to localize it.

config/sync/views.view.events.yml
```diff
@@ -0,0 +1,229 @@
...
+  config:
+    - field.storage.node.field_event_date
+    - search_api.index.events
+  module:
+    - datetime_range
+    - search_api
+id: events
+label: Events
+module: views
+description: ''
```
kasperg, line 14: Consider if you can provide a description that describes the purpose of the view.

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
kasperg, line 1: :+1: The BEM structure here makes sense to me.

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
kasperg: Do we need to render the identity color box if there is no media?

kasperg (changes requested):
Nice work! I like that you have made the effort of adding Search API already. That will come in handy later on.

There are some areas here that must be addressed. We can talk about them together.

In general: Consider your commit structure.

- Some commits are large. In a6621efc8c85d4151ed16881deb03fbe39aa3430 it is hard to see what is the result of enabling the `search_api` module(s) and what is is setting up the specific events view. In this regard 95a2f2e9df1ed7918051ab2733da4d66f9105b31 has a nice size.
- Remember to add a bit about your process and more importantly *why* in your commit comments.
  - Why did you add Search API and Date Range Formatter. Why did you create a new view mode for this list etc.
  - When adding modules and configuration it is helpful to add a bit of info about the process. Are these the defaults for Search API and Search API DB or did you intentionally do something extra that is important to the task at hand?

config/sync/core.entity_view_mode.node.event_list_teaser.yml
```diff
@@ -0,0 +1,10 @@
+uuid: 8c2b4ca5-9c0e-4734-a4a3-b03b6c7252b0
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+id: node.event_list_teaser
+label: 'Event list teaser'
```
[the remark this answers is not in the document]
rasben: I agree :) List-teaser would be my suggestion.

web/themes/custom/novel/novel.theme
rasben: Yeah, I would argue this is complicated enough that it should be in a module, rather than `.theme`

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
LasseStaus: I see your point. 

Since the tag was included in the design system, i feel like it makes sense to also have it here, if not for anything else then just for the visual demonstration.  Retaining it in the template makes it easier to see the layout as a whole instead of having missing parts in my opinion 🥳

web/themes/custom/novel/novel.theme
LasseStaus: I have moved this logic from the theme file, to the dpl_event module. 

In terms of price calculation, i've added methods and refactored priceFormatter.php 

I have also added tests for the new methods. 

**Important**

I got an error with phpcs specified in the commit that i was not able to fix. 
Please remove the line specified in the commit message for the tests and try to replicate.

web/themes/custom/novel/templates/views/node--event-list-teaser.html.twig
LasseStaus: I am not sure how to accomplish it. 

I thought i could overwrite the media or field when used in a specific view mode by appplying the view mode to the end of the template --list-teaser.html.twig, but i haven't been able to get this working. 

Not sure if view mode specific overrides are not possible by defualt or if they need to be added by preprocess. I tried this also but it didn't work fully. Could also be that i need to create a view mode for the media type, but this seems strange to me. 


I largely kept the same, but added correct alt text. see more in the commit.

LasseStaus: in terms of whether we need to render the identity color box, i would assume so. 

I guess this might not ahve been a part of the original ticket, but since media library was added etc, and i figured the identity should be there aswell.

web/modules/custom/dpl_event/tests/src/Unit/PriceFormatterTest.php
```diff
@@ -38,9 +38,103 @@ public function priceProvider(): array {
...
+   * @phpcs:ignoreFile
+   * Test price range formatting.
+   *
+   *
+   * @dataProvider priceRangeProvider
+   * @param int[] $prices
+   *   Array of integers representing prices.
+   * @param string $expected
+   *   Expected formatted string.
+   */
```
kasperg, line 129:
```suggestion
  /**
   * Test price range formatting.
   *
   * @param int[] $prices
   *   Array of integers representing prices.
   * @param string $expected
   *   Expected formatted string.
   *
   * @dataProvider priceRangeProvider
   */
```

This should pass.

web/themes/custom/novel/templates/views/node--list-teaser.html.twig
```diff
@@ -0,0 +1,40 @@
...
+
+<a href="{{ content.event_link }}" class="event-list-item arrow__hover--right-small">
+  <div class="event-list-item__image-container">
+    {% if image_file %}
+      <img src="{{ file_url(image_file.fileuri) }}" alt="{{ image_field.alt }}" class="event-list-item__image" />
+    {% endif %}
+  </div>
+  <div class="event-list-item__content">
+    <div class="tag tag--fill cursor-pointer tag--small event-list-item__tag">foredrag</div>
+    <time class="event-list-item__date" datetime="{{ content.field_event_date['#items'].0.value|date('Y-m-d') }}">
```
kasperg, line 16: Please reconsider using a `<time>` tag here.

When I think about it this tag may contain a range. As I understand it the current logic will not work here.

kasperg (approved):
:+1: This should be good enough to if you feel like it. We can always take another look afterwards.

PR #595, opened 2024-01-02
Remove `dpl_identity` module and move identity settings to `.theme` file

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-100

#### Description

This pull request removes the `dpl_identity` module and moves the identity settings to the .theme file. 

This change consolidates the theme-related settings in the appearance settings and eliminates the need for the `dpl_identityà module.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-cms/assets/49920322/c5a08aeb-f145-412b-9b3b-f601da61da21


composer.lock
kasperg, line 1: Please reconsider updating third party code as a part of this PR.

I wonder why this has been included here. Updating third party code is nice and all but it does not seem to be related to the change at hand. I could not find an explanation for it either in the commit comment.

web/themes/custom/novel/novel.theme
```diff
@@ -146,4 +147,55 @@ function novel_form_system_theme_settings_alter(array &$form, FormStateInterface
...
+/**
+ * Attaches identity color styles to the page.
+ *
+ * @param array<mixed> $variables
+ *   An associative array containing the following possible keys:
+ *   - '#attached': An array of attachments for rendering.
+ * @param \Spatie\Color\Hsl $hslColor
+ *   The HSL color object.
+ */
+function attach_identity_color_styles(array &$variables, Hsl $hslColor): void {
```
kasperg, line 184:
```suggestion
function novel_attach_identity_color_styles(array &$variables, Hsl $hslColor): void {
```

In general it is a good practice to prefix functions with their module names in Drupal - even if it is not a hook implementation. This has a number of advantages:

- It signals what part of the project a particular function belongs to.
- We avoid collisions because functions in PHP files are not namespaced by default.

web/themes/custom/novel/novel.theme
```diff
@@ -146,4 +147,55 @@ function novel_form_system_theme_settings_alter(array &$form, FormStateInterface
   // Remove the Path field from the logo and favicon upload fields.
   unset($form['logo']['settings']['logo_path']);
   unset($form['favicon']['settings']['favicon_path']);
+  // Add in identity settings.
+  $form['identity']['settings'] = [
+    '#type' => 'fieldset',
+    '#title' => t('identity settings'),
```
kasperg, line 153: Please remember the translation context.

web/themes/custom/novel/novel.theme
```diff
@@ -146,4 +147,55 @@ function novel_form_system_theme_settings_alter(array &$form, FormStateInterface
...
+  ];
+  $form['identity']['settings']['identity_color'] = [
+    '#type' => 'color',
+    '#title' => t('Identity Color', [], ['context' => 'Identity settings']),
+    '#default_value' => theme_get_setting('identity_color'),
+    '#description' => t('Choose library identity color.', [], ['context' => 'Identity settings']),
```
kasperg, line 159: Consider changing the translation context to "Novel settings" or the like

Now that we have moved the implementation the context could move as well. By using the module or theme name as the context we avoid having a million different contexts.

kasperg (changes requested):
:+1: but a few things needs to be updated here.

Most importantly is the part about updating third party code.

kasperg (approved):
:+1: LGTM

PR #596, opened 2024-01-02
Add new `Files` paragraph + new `Document` Media types

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-8

#### Description

This pull request adds a new `Files` paragraph that allows for adding links to uploaded files. This could well have been built into the current `Links` paragraph, but as we would like to keep our paragraphs as simple as possible, we have chosen to do it separately.

 It also introduces a new `Document` Media type for handling PDF files.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-cms/assets/49920322/b2ece8c1-1027-47e5-a729-12131c6ddb38


web/themes/custom/novel/templates/fields/field--media--field-media-file.html.twig
```diff
@@ -0,0 +1,9 @@
+{% for item in items %}
+   <a href="{{ item.content['#file']|file_url }}" class="link-with-icon arrow__hover--right-small">
+    <div class="link-with-icon__icon ">
+      <img class="invert" src="/themes/custom/novel/assets/dpl-design-system/icons/collection/Ebook.svg" alt="" />
+    </div>
+    {{ item.content["#description"] }}
+    {{ source("/themes/custom/novel/assets/dpl-design-system/icons/arrow-ui/icon-arrow-ui-small-right.svg") }}
```
kasperg, line 7: Please look into moving the icon path into a variable.

[We have already done this for paragraphs](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/f592784bf837dbee012bbde2293b2a78676ffbb9/web/modules/custom/dpl_paragraphs/dpl_paragraphs.module#L18-L18) and [used it in a template](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/f592784bf837dbee012bbde2293b2a78676ffbb9/web/themes/custom/novel/templates/paragraphs/paragraph--links.html.twig#L5-L5), but I think [it should be possible to do this for all templates in the theme](https://drupal.stackexchange.com/a/237666).

config/sync/field.field.media.document.field_media_file.yml
```diff
@@ -0,0 +1,27 @@
...
+description: ''
+required: true
+translatable: true
+default_value: {  }
+default_value_callback: ''
+settings:
+  handler: 'default:file'
+  handler_settings: {  }
+  file_directory: '[date:custom:Y]-[date:custom:m]'
+  file_extensions: pdf
```
kasperg, line 24: Consider talking to the business about what types of files they typically upload.

Limiting the allowed file extensions to PDF seems a bit too restrictive to me.

kasperg (approved):
:+1: A few things for you to consider.

Please remember to update the example content with an uploaded file. Here is a nice (and small) example PDF you can use: https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf.

Generally I think it is very nice that we can implementing this quickly based on all the previous work.

PR #599, opened 2024-01-03
Add missing translations based on errors

Adamik10:
#### Link to issue

DDFLSBP-359

#### Description
Adding hopefully the last missing translations and moving the ones used globally to global definition.
We recently introduced some changes that throw errors when translation definitions aren't provided to our React apps, and the result in dpl-cms was a lot of errors due to missing translation texts. So here we add them.

#### Screenshot of the result
n/a

n/a


spaceo (approved):
LGTM 👍

PR #601, opened 2024-01-04
Add multiple text fields for event partners and modify rendering of event place

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-27

#### Description
This pull request adds multiple text fields for event partners and modifies the rendering of the event place field to be an unordered list.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-cms/assets/49920322/f1db1123-d5e0-4762-ba0a-f7fc5eae6a43

PR #557, opened 2023-12-14
Add example content

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-83

#### Description

This adds an article and an event with content from Figma to the example content module.

- The example content module is enabled by default on development environments
- Documentation for generating example content is updated
- Any leftovers in the example module are deleted

#### Screenshot of the result

Monosnap Jesper Stein vinder Læsernes Bogpris for Rampen | DPL CMS 2023-12-14 17-49-14

Monosnap Fernisering Moderne Dans | DPL CMS 2023-12-14 17-48-29

#### Additional comments or questions

Content for the article looks a bit weird to me when it comes to spacing and alignment...

PR #602, opened 2024-01-05
Recurring events

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-25

#### Description

This PR does introductory work to add support for Recurring Events:

- The Recurring Events module is added
- The default event series and instances are configured with the same fields as the current event node type and inheritance between the two
- New templates are added which reuse the current node templates to show that what has already been developed still works 

The purpose of this is to provide enough of a foundation to show that we have something to demo and built upon going forward.

These tasks should be addressed in future PRs:

- Automatic handling of event states should be moved from event nodes to event instances
- Any work that has been done in the meantime for events e.g. #586, #601, #596 and #557 will have to be updated to use new entities
- The current event content type should be deleted
- All templates for event nodes must be migrated properly to event instances and series
- Whatever I have forgotten :smile:


rasben (approved):
This is a bit of a rubber stamp, as this is so much stuff, and it's hard to tell what might be problematic, until I actually work with it myself.

PR #617, opened 2024-01-08
Add Taxonomy + Expandable Categories in Event Content Type

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-130

#### Description

This pull request introduces two key updates to the event content type in the 'dpl-cms' system. Firstly, it integrates taxonomy and categories, enabling users to tag events effectively. Secondly, it adds a 'Show More/Less' feature to the field categories template, improving user interaction by allowing the category list to be expanded or collapsed. 

#### Screenshot of the result
image
image

### DIFF
--- a/web/themes/custom/novel/templates/fields/field--field-categories.html.twig
+++ b/web/themes/custom/novel/templates/fields/field--field-categories.html.twig
@@ -1,5 +1,14 @@
+{{ attach_library('novel/show-more') }}
+{% set show_more_text = '...'|t %}
+{% set show_less_text = 'Show less'|t %}
+
+
 {% for item in items %}
-	<span class="tag tag--fill cursor-pointer tag--small">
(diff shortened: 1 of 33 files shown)


web/themes/custom/novel/templates/fields/field--field-categories.html.twig
```diff
@@ -0,0 +1,27 @@
+{{ attach_library('novel/show-more') }}
+{% set show_more_text = '...'|t %}
+{% set show_less_text = 'Show less'|t %}
+
+{% if items|length > 1 %}
```
rasben, line 5: I get why you're doing it, but instead of having an ìf`/`else`, why not just always print the parent, and then maybe add a `if item|length > 1, add class that JS listens for`

PR #613, opened 2024-01-08
Related/Upcoming events slider view, displayed on event page. DDFFORM-86

rasben:
- Add a field_teaser_text / field_teaser_image, that is used for card fields. The teaser text is the alternative text that is shown in replacement for a missing teaser image. This has also been done for article nodes.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-86

Depends on https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/420

PR #618, opened 2024-01-08
Introduce 'content slider' paragraph. DDFFORM-89

rasben:
Also introduces a re-usable entity reference view, that can be used for other content reference fields.

It builds ontop of https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/613 and https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/420 which really forms the whole basis for this PR.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-89


LasseStaus (changes requested):
The paragraph itself works, LGTM. 

However, the display is incorrect because the paragraph is rendered within some container with "event-paragraphs" classes. 

in `node--event--full.html.twig` all paragraphs are rendered within this `  <section class="event-paragraphs">
    {{ content.field_event_paragraphs }}
  </section>`
  
Which is incorrect in my opinion since we now can't control the layout of paragraphs that should go further. 

This is also a thing that should be changed in the design-system for `ÈventParagraphs.tsx` and the block(s) displaying the page. I think all the paragraphs should control they own position, but this is not the case currently. 

The optimal layout css can be added to [this ticket](https://reload.atlassian.net/jira/software/c/projects/DDFFORM/boards/485?selectedIssue=DDFFORM-156), but the event-paragraphs container should be removed to not have too many "refactorings" later.

rasben:
This is ready for you again, @LasseStaus

PR #623, opened 2024-01-10
Change patron menu translation function string params

spaceo:
#### Description

Change patron menu translation function string params
To use single quotes instead of double.
This is crazy, but apparently Potion won't use context if it is not defined with single quotes!!! Really!


Dresse (approved):
LGTM

PR #624, opened 2024-01-10
Don't trigger translation workflow on release branches

spaceo:
#### Description
Don't trigger translation workflow on release branches
Since it is not possible to make the full translation workflow work with dynamic branches (which release/*) are we remove that possibility again.


Dresse (approved):
LGTM

PR #632, opened 2024-01-15
Update FeesListBlock text string definitions

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-332

#### Description

A sibling to [this PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/857) where some text string definitions were changed.

#### Screenshot of the result
-
Accessibility and cypress are failing because the dpl-react package doesn't contain these translation strings. 
I will update it after the sibling PR is merged in.


Dresse (approved):
LGTM 👍

PR #640, opened 2024-01-16
Skip language import in pa11y ci test

spaceo:
#### Description

In order to have stable tests we need to be able trust the ui language and not get suprised by translated strings.


Dresse (approved):
LGTM 👍 I previously ran into the same problem.

PR #651, opened 2024-01-17
Set pointer to version 2024.3.1 of dpl-react

spaceo:
#### Description

Set pointer to version 2024.3.1 of dpl-react
Which contains materialType field changes on Work. Previously they were only implemented on manifestations.


Dresse (approved):
LGTM

PR #653, opened 2024-01-18
DDFLSBP-377 - Added a new variable to dpl_react_apps_format_app_url f…

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-377

#### Description

This PR makes it possible to pass an empty string instead of an URL string to react_apps. 
The reason for this is that we in some cases would like to be able to check if a URL field is empty or not.

It has a sister PR in dpl-react. It can be found here: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/864

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
                 "version": "2024.3.0",
                 "type": "drupal-library",
                 "dist": {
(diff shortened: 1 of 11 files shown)


spaceo (changes requested):
We have found another solution. This should probably go to the eternal afterlife of non merged PR's

spaceo (approved):
LGTM 👍

PR #654, opened 2024-01-18
Migrate sliders to use eventseries/instances. DDFFORM-161

rasben:
- Re-work the event series and instances to support 'card' display.
- Add a custom service, that generates a human-readable date string based on an event series - e.g. 'daily, 20:00-21:00' or similar.
- Present this scustom date string in event series displays
- Add a basic slider on event series, showing upcoming events, rather than just showing it as a "broken" wall of text at the top. This will need to be updated in the future, as it's just winging something that has no design.
- Introduce teaser text/image fields, also used by cards.
- Introduce `dynamic_entity_reference` module, that allows us to make reference fields that also allows searching for event instance/series content.
- Update the slider paragraph, to use `dynamic_entity_reference`
- Moving card HTML into a seperate file, to make it easier for nodes and eventseries to re-use it.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-161

https://reload.atlassian.net/browse/DDFFORM-122


kasperg (approved):
:+1: LGTM

PR #656, opened 2024-01-18
Delete 'event' node type, and clean up templates. DDFFORM-122

rasben:
The event node has been replaced by eventseries/eventinstances from the reccurring_events module.


kasperg (approved):
:+1: Looks good.

It is a bit hard to figure out when you are done but a few more things that could be cleaned up:
- References to nodes in dpl_event.module: https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/03070725536cb11af36250b921f17393b03cb22b/web/modules/custom/dpl_event/dpl_event.module#L241-L241. This is typically docblocks that people have forgotten to update.
- Example content event node: https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/03070725536cb11af36250b921f17393b03cb22b/web/modules/custom/dpl_example_content/dpl_example_content.info.yml#L15-L15 and related comments

PR #657, opened 2024-01-18
Paragraph: Card-grid (Nyhedskomponent). DDFFORM-127

rasben:
Requires this `dpl-design` PR to be merged: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/439

----

A pretty simple paragraph, following the same style as the slider. Most of the twig templating is added in the `field` template, as we want to be able to control the `items` very directly.

**This is the manual version of the paragraph.**
Up next, there'll be a seperate paragraph that adds content automatically, through a view.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-127

-----

Screenshot 2024-01-18 at 21 06 17


LasseStaus (approved):
I don't see any things in terms of card-grid that needs changing besides design system. Nice work! 


I noticed this for events -> event instance settings -> form display: "
image
"

The edit  mode for ticket categories should not be default open. 
I would expect this to not be the case, since i changed that in another PR. 

The config changes between the PR's confuse me a bit. As long as this doesnt add to the code that these should be default open it's fine.

PR #665, opened 2024-01-19
Taxonomy: Categories and tags

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-130

#### Description

This PR adds two new taxonomies: Categories and tags, adds these to articles event series and instances with field inheritance setup between the two.

As this is in place we can now render categories for these. This PR adds it for articles, event series, instances and on the list page.

To make this easier to demo an category term is added to the example content.

This replaces #662 and depends on https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/452.

#### Screenshot of the result

Article

billede

Event Instance

billede

Series

billede

List

billede

Term editing

billede

Event field

billede

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,9 +52,9 @@
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-design-system",
                 "type": "drupal-library",
-                "version": "0.0.0-dev",
+                "version": "2024.6.0-rc1",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/branch-taxonomy_categories/dist-taxonomy-categories.zip",
+                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/branch-release%2F2024-6-0/dist-release-2024-6-0.zip",
                     "type": "zip"
                 }
             }
@@ -74,7 +74,7 @@
         "brick/math": "^0.12.1",
         "composer/installers": "1.12.0",
         "cweagans/composer-patches": "1.7.3",
-        "danskernesdigitalebibliotek/dpl-design-system": "0.0.0-dev",
+        "danskernesdigitalebibliotek/dpl-design-system": "2024.6.0-rc1",
         "danskernesdigitalebibliotek/dpl-react": "2024.6.0-rc1",
         "danskernesdigitalebibliotek/fbs-client": "*",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.1",
--- a/composer.json
+++ b/composer.json
@@ -52,9 +52,9 @@
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-design-system",
                 "type": "drupal-library",
-                "version": "2024.6.0-rc1",
+                "version": "0.0.0-dev",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/branch-release%2F2024-6-0/dist-release-2024-6-0.zip",
+                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/branch-taxonomy_categories/dist-taxonomy-categories.zip",
                     "type": "zip"
                 }
             }
@@ -74,7 +74,7 @@
         "brick/math": "^0.12.1",
         "composer/installers": "1.12.0",
         "cweagans/composer-patches": "1.7.3",
-        "danskernesdigitalebibliotek/dpl-design-system": "2024.6.0-rc1",
+        "danskernesdigitalebibliotek/dpl-design-system": "0.0.0-dev",
         "danskernesdigitalebibliotek/dpl-react": "2024.6.0-rc1",
         "danskernesdigitalebibliotek/fbs-client": "*",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.1",
--- a/composer.json
+++ b/composer.json
@@ -115,7 +115,6 @@
         "drupal/restui": "^1.21",
         "drupal/search_api": "^1.31",
         "drupal/twig_tweak": "^3.2.1",
-        "drupal/twigsuggest": "^1.0@RC",
         "drupal/upgrade_status": "^4.0",
         "drupal/uuid_url": "^1.3",
         "drupal/varnish_purge": "^2.2",
--- a/composer.json
+++ b/composer.json
@@ -115,6 +115,7 @@
         "drupal/restui": "^1.21",
         "drupal/search_api": "^1.31",
         "drupal/twig_tweak": "^3.2.1",
+        "drupal/twigsuggest": "^1.0@RC",
         "drupal/upgrade_status": "^4.0",
         "drupal/uuid_url": "^1.3",
         "drupal/varnish_purge": "^2.2",
--- a/config/sync/core.entity_form_display.node.article.default.yml
+++ b/config/sync/core.entity_form_display.node.article.default.yml
@@ -29,7 +29,7 @@ third_party_settings:
       label: 'Meta tags'
       region: content
       parent_name: ''
-      weight: 6
+      weight: 8
       format_type: details_sidebar
       format_settings:
         classes: ''
@@ -46,7 +46,7 @@ third_party_settings:
       label: 'Teaser card'
       region: content
       parent_name: ''
-      weight: 7
+      weight: 9
       format_type: details_sidebar
       format_settings:
         classes: ''
@@ -71,13 +71,13 @@ content:
     third_party_settings: {  }
   field_categories:
     type: options_select
-    weight: 26
+    weight: 2
     region: content
     settings: {  }
     third_party_settings: {  }
   field_override_author:
     type: string_textfield
-    weight: 5
+    weight: 7
     region: content
     settings:
       size: 60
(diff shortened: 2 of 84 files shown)


web/themes/custom/novel/templates/fields/field--eventinstance--event-categories--list-teaser.html.twig
```diff
@@ -0,0 +1,3 @@
+{% include 'field--field-categories.html.twig' with {
+  item_attributes: create_attribute().addClass('event-list-item__tag')
```
rasben, line 2: Doesnt `attributes` already exist on the items?

config/sync/core.extension.yml
```diff
@@ -101,6 +102,7 @@ module:
   views: 10
   paragraphs: 11
   paragraphs_ee: 15
+  twigsuggest: 100
```
rasben, line 105: Eh, I think this is a bit overkill.
I think it'd be better to just make a really simple `dpl_theme_suggestions` (or whatever) module, that adds the suggestions we need, that we can always expand on

Twigsuggest is only a "yellow" release candidate, and the code for it seems a bit.. complicated.

rasben (changes requested):
A few small comments, along with:

- Please make the tag field a dropdown, as it's easier for editors to understand

kasperg:
> Please make the tag field a dropdown, as it's easier for editors to understand

I have done so, but I do not think a multiselect HTML element (default for multivalue fields) makes things much better. Do we need another module here?

Monosnap Edit default Event Fernisering Moderne Dans | DPL CMS | Logged in 2024-01-21 21-32-27

config/sync/core.extension.yml
kasperg: :+1: I have replaced this with a hook implementation in the theme.

web/themes/custom/novel/templates/fields/field--field-categories.html.twig
```diff
@@ -0,0 +1,29 @@
+{{ attach_library('novel/show-more') }}
+{% set show_more_text = '...'|t %}
+{% set show_less_text = 'Show less'|t %}
+
+{% set item_attributes = item_attributes ?? create_attribute() %}
```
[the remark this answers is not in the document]
kasperg: Done in [6824ffc](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/665/commits/6824ffc00bd4df88c00862e2232d37f7b789af7f).

web/themes/custom/novel/templates/fields/field--eventinstance--event-categories--list-teaser.html.twig
kasperg: You are totally right. I replaced this with proper `item.attributes` in [b294c0d](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/665/commits/b294c0d39fc5346b7f04f1b35c81e5976004d408). Much better!

rasben:
>I have done so in [4f577fa](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/665/commits/4f577faa0d5f9a17ad86a4ea7d18c9f54ace8b1a), but I do not think a multiselect HTML element (default for multivalue fields) makes things much better. Do we need another module here?

Hmm.. maybe you could set it as checkboxes instead? I dont know how many categories we're planning on having :)

rasben (dismissed):
Approved, but yeah, you're right that a multi `select` isnt too admin friendly.
You dont have to fix it in this PR though :) But, as i said, either look at just making it checkboxes, or include the jquery_ui_multiselect module.

We use that module on BUPL, however, notice that you'll have to use the dev version:

```
 "drupal/jquery_ui_multiselect_widget": "1.x-dev@dev",

File: config/sync/user.role.editor.yml [hunk 4]
```diff
@@ -30,6 +32,7 @@ permissions:
   - 'create article content'
   - 'create campaign content'
   - 'create image media'
+  - 'create terms in categories'
```
LasseStaus, line 35: I'm unsure which roles are given to who. But should regular editors be able to create these? I thought it was only for admins / "webmaster".

composer.json
```diff
@@ -115,6 +115,7 @@
         "drupal/restui": "^1.21",
         "drupal/search_api": "^1.31",
         "drupal/twig_tweak": "^3.2.1",
+        "drupal/twigsuggest": "^1.0@RC",
```
LasseStaus: This is nice. 

It also makes {{base_path}} available in twig templates

web/themes/custom/novel/templates/fields/field--field-tags.html.twig
```diff
@@ -0,0 +1,4 @@
+{% include '@novel/components/horizontal-term-line.html.twig' with {
+  'title': "Tags"|trans({}, {"context": "Tags"}),
```
LasseStaus, line 2: Im unsure if this should be called tags, or categories / other.

web/themes/custom/novel/templates/fields/field--node--field-tags--article--full.html.twig
```diff
@@ -0,0 +1,5 @@
+{% include '@novel/components/horizontal-term-line.html.twig' with {
+  'title': "Tags"|trans({}, {"context": "Tags"}),
```
LasseStaus, line 2: Im unsure if this should be called tags, or categories / other.

composer.json
```diff
@@ -115,7 +115,6 @@
         "drupal/restui": "^1.21",
         "drupal/search_api": "^1.31",
         "drupal/twig_tweak": "^3.2.1",
-        "drupal/twigsuggest": "^1.0@RC",
```
LasseStaus: Maybe the old commit enabling this could be removed.

config/sync/core.entity_form_display.node.article.default.yml
```diff
@@ -125,6 +126,16 @@ content:
...
+  field_tags:
+    type: entity_reference_autocomplete_tags
+    weight: 27
+    region: content
+    settings:
+      match_operator: CONTAINS
+      match_limit: 10
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
```
LasseStaus: Just marking this here. 

To add the _next_ tag after the first one, a comma is necessary. Is there a simple way to make the input automatically add this comma when a tag is selected clicked.

web/themes/custom/novel/templates/fields/field--field-tags.html.twig
```diff
@@ -0,0 +1,10 @@
+<div class="text-small-caption horizontal-term-line">
+  <h2 class="text-label-bold">
+    {{ "Tags"|trans({}, {"context": "Tags"}) }}
```
LasseStaus: Im unsure if this should be named tags instead of categories / other

composer.json
```diff
@@ -52,9 +52,9 @@
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-design-system",
                 "type": "drupal-library",
-                "version": "2024.6.0-rc1",
+                "version": "0.0.0-dev",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/branch-release%2F2024-6-0/dist-release-2024-6-0.zip",
+                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-design-system/releases/download/branch-taxonomy_categories/dist-taxonomy-categories.zip",
```
LasseStaus, line 57: Remember to revert 🪂

LasseStaus (approved):
Looks good when i'm testing it (locally). 

I would insist we change the default recur type away from "consecutive event". I suggest single event as default. 

Consider changing the naming of "Tags" to "Categories" in the `field--field-tags.html.twig` template and other relevant places.

If i create a new event under ´admin/content/events´ with 'Add event', i get this with for a single event:
image

Not sure if i am creating the event incorrectly.

config/sync/core.entity_form_display.node.article.default.yml
kasperg: Good idea. Perhaps.

composer.json
kasperg: Yeah, but I think it still has some value in documenting the process to [`novel_theme_suggestions_alter()` in case someone comes along looking for an alternate at some point](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/574600bda1fac037093222f816870bffc6b475bd/web/themes/custom/novel/novel.theme#L234-L234).

composer.json
kasperg: I also like it but we decided against (for now) in [20e8df9](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/665/commits/20e8df9caccaf6388cbb60a08a4376ebd0c80727).

config/sync/user.role.editor.yml
kasperg: I understand. We have a "lesser" role called Mediator which I intentionally did not add this permission to.

kasperg:
> I would insist we change the default recur type away from "consecutive event". I suggest single event as default.

I think it is outside the scope of this PR but [I raised it with the business who requested this recur type to be kept available](https://reload.zulipchat.com/#narrow/stream/412046-DDF---Formidling.2B/topic/Tags.2Femneord/near/418095473).

PR #666, opened 2024-01-19
Remove Add Event link from events list

kasperg:
#### Description

We use the /events path for our events list. The same path is also
used by the recurring_events module with the
entity.eventinstance.collection route. Consequently we remove any
action which recurring event or other modules configure to appear
here.

See cms/web/modules/contrib/recurring_events/recurring_events.routing.yml

#### Screenshot of the result

Before:

Monosnap Events | DPL CMS 2024-01-19 21-16-19

### DIFF
--- a/web/modules/custom/dpl_admin/dpl_admin.module
+++ b/web/modules/custom/dpl_admin/dpl_admin.module
@@ -41,6 +41,22 @@ function dpl_admin_local_tasks_alter(array &$local_tasks): void {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


rasben (changes requested):
Looks good, but please fix the things that phpcs is complaining about :)
(pro-tip: you can use `task dev:format`)

kasperg:
Sorry about that. I have fixed the issues.

rasben (approved):

PR #668, opened 2024-01-21
Optimize github action ressource usage.

rasben:
- Only run the most ressource heavy jobs on non-draft PRs. Stuff like Chromatic deployment only makes sense once the pull request has actually been marked as ready for review, as other times, you might push several times to the branch
- Add a concurrency checker for all jobs, cancelling on-going jobs if a new push to the same branch has been detected.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,9 +1,16 @@
 # These workflows takes care of various CI tests.
-on: pull_request
 name: CI Tests
+
+on: pull_request
 env:
   PHP_VERSION: 8.1
   COMPOSER_VERSION: v2
+
+# Detect if this action is already running, and cancel it.
+# This most likely happened because a second push has been made to a branch.
+concurrency:
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   ValidateComposer:
@@ -105,6 +112,12 @@ jobs:
 
   LightHouse:
     name: Test site performance using Lighthouse
+
+    # Don't run action if the pull request is a draft.
+    # This is an expensive job to run, so we only want to run it when the
+    # PR is actually ready to be reviewed.
+    if: ${{ github.event.pull_request.draft == false }}
+
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-source.yml
```diff
@@ -2,6 +2,13 @@
 name: Publish source
 on:
   create
+
+# Detect if this action is already running, and cancel it.
+# This most likely happened because a second push has been made to a branch.
+concurrency:
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
spaceo, line 11: I think it as a bad idea to use it on the translation system. Maybe I am wrong, but I think it could have unforeseen consequences . So I would rather leave it without the concurrency functionality.

.github/workflows/translate-source.yml
```diff
@@ -5,6 +5,12 @@ on:
     branches:
       - develop
 
+# Detect if this action is already running, and cancel it.
+# This most likely happened because a second push has been made to a branch.
+concurrency:
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
spaceo, line 12: Same comment as previous

spaceo (changes requested):
Looks good but one thing I think wee need to change.

.github/workflows/ci-tests.yml
```diff
@@ -105,6 +112,12 @@ jobs:
 
   LightHouse:
     name: Test site performance using Lighthouse
+
+    # Don't run action if the pull request is a draft.
+    # This is an expensive job to run, so we only want to run it when the
+    # PR is actually ready to be reviewed.
+    if: ${{ github.event.pull_request.draft == false }}
+
```
kasperg, line 120: Personally I am not fan of this change.

My workflow is usually that I initially create pull requests on GitHub in the draft state to indicate that it is not ready for review (or merge-able), see that al checks are green and then assign it. Now I will have to create the PRs that are ready for review before I can see if all checks are green. 

Alternately I can create them as draft, see that they are green, then mark them as ready for review, then I have to see that they are still green. *Then* I can assign them.

Perhaps not a huge problem - but different and potentially more time consuming.

I

kasperg (commented):
Generally I am not a fan of the changes here. As I see it they introduce complexity without much gain.

I am willing to try it out though if it makes a difference to someone else.

PR #675, opened 2024-01-22
Set start date as a default date for end date input fields. DDFFORM-171

rasben:
Screenshot 2024-01-22 at 19 06 45
Screenshot 2024-01-22 at 19 07 57


web/modules/custom/dpl_admin/assets/dpl_admin.js
```diff
@@ -0,0 +1,39 @@
...
+  attach(context, settings) {
+    const that = this;
+
+    const dateTimeFields = context.querySelectorAll(
+      '[type="date"][name$="[value][date]"]:not(.is-dpl-admin-range-initialized),' +
+        '[type="time"][name$="[value][time]"]:not(.is-dpl-admin-range-initialized)'
+    );
+
+    dateTimeFields.forEach((input) => {
+      input.classList.add("is-dpl-admin-range-initialized");
```
kasperg, line 12: Consider using a data attribute instead of a CSS class to mark state.

kasperg (approved):
:+1: Well done.

I see you not adding jQuery. This is in line with what we do elsewhere on the site. I do think it is less important here as the Drupal admin will add jQuery on it's own anyway.

PR #677, opened 2024-01-22
Replace Reviewdog with CS2PR

kasperg:
#### Description

Reviewdog will filter checkstyle results based one files changed in a pull request and modify the return code and thus success/failure of a run accordingly. This can cause problems e.g. if errors occur many lines of code away from a specific change. Such errors will pop op locally where we do not used Reviewdog. This causes causing frustration for the next developer.

To avoid this we drop Reviewdog entirely. Instead we use the CS2PR tool to convert checkstyle results from PHPCS to GitHub pull request annotations.

As the tool is only relevant in GitHub Actions we only install it there by adding it as a tool in setup-php-action.

This is the same approach as suggested by setup-php-action: https://github.com/marketplace/actions/setup-php-action#tools-with-checkstyle-support

#### Screenshot of the result

After this change errors checks fail and current errors are reported:

billede

[99ae723](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/677/commits/99ae7238d848419630220e3603c70cb4e89957a6) fixes the issues.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -35,16 +35,15 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: ${{ env.PHP_VERSION }}
-        tools: composer:${{ env.COMPOSER_VERSION }}
+        tools: composer:${{ env.COMPOSER_VERSION }}, cs2pr
         coverage: none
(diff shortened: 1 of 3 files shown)


Dresse (approved):
LGTM

PR #682, opened 2024-01-23
Allow self-signed certificates when checking Lagoon environments

kasperg:
#### Link to issue

This is a temporary fix for https://reload.atlassian.net/browse/DDFDRIFT-31

#### Description

Replace nev7n/wait_for_response with iFaxity/wait-on-action. This
action supports disabling strict SSL. This way our check will pass
even if Lagoon has not (yet) deployed a proper Let's Encrypt
certificate.

Users will still get a warning in their browsers but that is
preferable to GitHub Action checks randomly failing.

Dropping the action is not helpful as this helps us understand if the
resulting environment is broken and e.gf. returns error code 500 for
some reason.

wait-on-action configuration is very close to wait_for_response.
It does support setting a response code but that should not matter
for us.

To disable strict SSL check we have to provide a config file. To keep
things contained we create this as a part of the workflow.

To see that this is working check out the environment for this PR.

### DIFF
--- a/.github/workflows/lagoon.yml
+++ b/.github/workflows/lagoon.yml
@@ -41,11 +41,15 @@ jobs:
           ref: ${{ github.head_ref }}
           logs: ${{ steps.environment.outputs.logs }}
           debug: ${{ runner.debug && 'true' || 'false' }}
+      - name: Generate wait-on config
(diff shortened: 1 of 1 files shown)


spaceo (approved):
LGTM 👍

PR #691, opened 2024-01-24
Feature/ddflsbp 355/error when creating users not already linked to a library

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-355

#### Description

This PR fixes an issue where users not already associated with a library (agencyId) were failing the user registration process.

It solves a few issues:

1. OpenId requires a sub, which used to be based on the uniqueId attribute in the userinfo array. New users do not have this attribute. Instead we are now using the CPR attribute if present, if not we are using the uniqueId attribute. If neither of those are present, we are throwing an error. 

2. A part of the dpl_login_openid_connect_pre_authorize hook is used for checking if a user has any blocked status. This is done by calling FBS using the getPatronInformation call. However this returns a 403 error for new users. 
To fix this, we now look in the userinfo to see if the user has any agencyId associated, if not, it means that this user is new and we then skip the check for blocked status.

3. One of the strings used in the /user/registration/create form was undefined. We have moved that into the global scope as it is also used in the user menu.

4. Added logging during the user registration and login flow. This was originally part of DDFLSBP-391 but was moved to this PR instead. Logging is added to the dpl_login_openid_connect_pre_authorize and dpl_login_openid_connect_post_authorize hooks.

5. The PR also expands and updates the cypress login test to be able to handle 4 test cases:
 - User creation with both CPR and uniqueId attributes.
 - User creation with only CPR attribute.
 - User creation with only uniqueId attribute.
 - User creation without both CPR and uniqueId attributes.


web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -71,41 +78,48 @@ function dpl_login_openid_connect_pre_authorize(AccountInterface|FALSE $account,
...
-  catch (ApiException $e) {
-    $logger->error('Authorization blocked. Unable to get patron information: @message', [
-      '@message' => $e->getMessage(),
-    ]);
-    return FALSE;
+  // Check if the user is associated with any agencies.
+  // If that is the case, the user is new, and we skip
+  // checking for blocked status.
+  if (!empty($context['userinfo']['attributes']['agencies'])) {
+    try {
```
spaceo, line 85: Please flip the logic so you do not have to indent the whole block of code.
So this should do it, I think:
```php
if (empty($context['userinfo']['attributes']['agencies'])) {
  return TRUE;
}
```

web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -27,21 +29,26 @@ use function Safe\sprintf as sprintf;
...
+ *
+ * @throws Exception
  */
 function dpl_login_openid_connect_userinfo_alter(array &$userinfo, array $context): void {
 
-  // If we cannot resolve uinque id we cannot continue.
-  if (!$id = $userinfo['attributes']['uniqueId'] ?? FALSE) {
-    $userinfo = [];
-    return;
+  // If we cannot resolve cpr we cannot continue.
```
spaceo, line 37: I guess it should be:
```
// If we cannot resolve both cpr and uniqueId we cannot continue.

File: web/modules/custom/dpl_react_apps/dpl_react_apps.module [hunk 3]
```diff
@@ -189,6 +189,7 @@ function dpl_react_apps_texts(): array {
     'patron-page-change-pincode-body' => t('Change current pin by entering a new pin and saving', [], ['context' => 'Global']),
     'patron-page-change-pincode-header' => t('PINCODE', [], ['context' => 'Global']),
     'patron-page-confirm-pincode-label' => t('Confirm new pin', [], ['context' => 'Global']),
+    'patron-page-phone-input-message' => t('The phone number must be 6 to 15 characters in length and should be comprised solely of numbers or begin with a +', [], ['context' => 'Global']),
```
spaceo, line 192: Is this a deliberate change?

spaceo (commented):
Looks good :)
I have some comments I would like you to look at.

web/modules/custom/dpl_login/dpl_login.module
Dresse: Yes, you are right. 
I have updated the comment now.

web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -140,6 +161,56 @@ function dpl_login_openid_connect_userinfo_save(UserInterface $account, array $c
...
+ * @return \Psr\Log\LoggerInterface
+ *   Logger object.
+ */
+function _dpl_login_get_user_login_logger(): LoggerInterface {
+  $logger_key = "dpl_login";
+
+  /** @var \Drupal\Core\GeneratedUrl $url */
+  $url = Url::fromRoute('dpl_patron_reg.create')->toString(TRUE);
+
+  if (($_SESSION['openid_connect_destination'] ?? "") === $url->getGeneratedUrl()) {
```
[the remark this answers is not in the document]
Dresse: I agree. I have added a comment that describes the logic.

web/modules/custom/dpl_react_apps/dpl_react_apps.module
Dresse: Yes. I moved the string to where the other registration form strings are stored. 
Before this, we got an error as react couldn't find it.

spaceo (approved):
LGTM 👍

PR #697, opened 2024-01-28
Add task that enables us to set openid settings

spaceo:
Add task that enables us to set openid settings (and runs cron)
In that way we do not have to do the tedious repeated configuration in the admin area every time we spin up a new version of the site locally.

It depends on variables set in .env file. See .env.example to see what is needed.


Dresse (approved):
LGTM

Very nice to have this added as a task!

PR #700, opened 2024-01-29
DDFLSBP-417 - Updated Library Token retriever function to be able to …

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-417

#### Description

This PR adds the possibility to do a "Force" retrieval of Library tokens.  This way we do not have to wait for the old token to expire before setting the new one using the new settings in OpenID Connect configuration.

This is helpful when you want to switch between different Library Agencies. This is especially helpful when debugging locally or when the client wants to do testing on demo environments. 

The PR adds hook_form_FORM_ID_alter and a custom submit handler for openid_connect_admin_settings form. The submit handler will now do a "Force" retrieval of a new token whenever a new Agency ID value is set.

### DIFF
--- a/web/modules/custom/dpl_login/dpl_login.module
+++ b/web/modules/custom/dpl_login/dpl_login.module
@@ -211,9 +211,8 @@ function dpl_login_openid_connect_admin_settings_submit(array &$form, FormStateI
   if ($form_state->getValue('agency_id') != $agency_id) {
     /** @var Drupal\dpl_library_token\LibraryTokenHandler $handler */
     $handler = Drupal::service('dpl_library_token.handler');
-    $result = $handler->retrieveAndStoreToken(TRUE);
 
-    if ($result === FALSE) {
+    if (!$handler->retrieveAndStoreToken(TRUE)) {
       $messenger->addMessage(t('Unable to retrieve token from Adgangsplatformen. Please check if the provided Agency ID is correct.', [], ['context' => 'openId connect settings']), 'error');
     }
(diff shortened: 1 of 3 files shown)


web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -179,3 +180,41 @@ function dpl_login_preprocess_html(array &$variables): void {
...
+  // If agency_id is different from the already stored, we want to
+  // force a new token to be created that belongs to the new agency.
+  if ($form_state->getValue('agency_id') != $agency_id) {
+    /** @var Drupal\dpl_library_token\LibraryTokenHandler $handler */
+    $handler = Drupal::service('dpl_library_token.handler');
+    $result = $handler->retrieveAndStoreToken(TRUE);
+
+    if ($result === FALSE) {
+      $messenger->addMessage(t('Unable to retrieve token from Adgangsplatformen. Please check if the provided Agency ID is correct.', [], ['context' => 'openId connect settings']), 'error');
+    }
```
spaceo, line 218:
```suggestion

    if (!$handler->retrieveAndStoreToken(TRUE)) {
      $messenger->addMessage(t('Unable to retrieve token from Adgangsplatformen. Please check if the provided Agency ID is correct.', [], ['context' => 'openId connect settings']), 'error');
    }
```

spaceo (approved):
LGTM
I have one suggestion, but otherwise 👍

PR #701, opened 2024-01-30
Use release 2024.5.0 of dpl-design-system and dpl-react packages

Adamik10:
Use release 2024.5.0 of dpl-design-system and dpl-react packages


clausbruun (commented):
LGTM

PR #706, opened 2024-01-30
Nav-grid paragraph (Navigationskomponent) - DDFFORM-164

rasben:
A new manual paragraph, with the same concept as `card-grid`, in terms of how to deal with fields, title, etc.

https://reload.atlassian.net/browse/DDFFORM-164

See more info in the design system: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/463


rasben:
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/463

config/sync/language/da/core.date_format.short.yml
```diff
@@ -1 +1,2 @@
 label: 'Standard kort dato'
+pattern: 'm/d/Y - H:i'
```
LasseStaus, line 2: Is this supposed to be here or a mistake? 

This is a change i keep seeing repeated in config exports.

config/sync/language/da/views.view.files.yml
```diff
@@ -26,8 +26,6 @@ display:
           label: Ændringsdato
         count:
           label: 'Brugt i'
-          alter:
-            path: 'admin/content/files/usage/{{ fid }}'
```
LasseStaus, line 30: This is also a thing i keep seeing repeated in config. 
I think you mentioned before you wanted to talk about this, but i dont think we ever did.

config/sync/language/da/views.view.user_admin_people.yml
```diff
@@ -19,9 +19,6 @@ display:
           label: Roller
         created:
           label: 'Medlem i'
-          settings:
-            future_format: '@interval'
-            past_format: '@interval'
```
LasseStaus, line 24: Same as others. 

"This is also a thing i keep seeing repeated in config.
I think you mentioned before you wanted to talk about this, but i dont think we ever did."

web/themes/custom/novel/templates/components/nav-teaser.html.twig
```diff
@@ -0,0 +1,19 @@
+<article {{ attributes.addClass('nav-teaser') }}>
+  <a href="{{ url }}">
+    <h1 class="nav-teaser__title">{{ title }}</h1>
+
+    {% if subtitle %}
+      <p class="nav-teaser__subtitle">{{ subtitle }}</p>
+    {% endif %}
+
+    <svg
```
LasseStaus, line 9: In other components we are using source.
We shouln't use svg's like this directly. 

An exmaple is eventinstance--list-teaser 

`  {{ source(directory ~ '/assets/dpl-design-system/icons/arrow-ui/icon-arrow-ui-small-right.svg') }}`

LasseStaus (changes requested):
I have some comments on some config changes i keep seeing in our PRs. 

I think you should change the SVG implementation. 


In general, i would argue this is too big be to be one commit. It would help a great deal for readability if you seperated it more

config/sync/language/da/core.date_format.short.yml
rasben: I dunno why it happens - either way, I've just manually removed it now.

config/sync/language/da/views.view.user_admin_people.yml
rasben: I dunno why it happens - either way, I've just manually removed it now.

web/themes/custom/novel/templates/components/nav-teaser.html.twig
rasben: 👍

LasseStaus (approved):
LGTM 👍 🥳 

Would be nice with a comment on my request for smaller commits.

rasben:
>Would be nice with a comment on my request for smaller commits.

I dont see how I can really split it up, as it only has business value when it's all connected.

PR #708, opened 2024-01-31
Feature/ddflspb 418/update branches when changing agency

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-418

#### Description

This PR changes the way we set Library Tokens. The main problem why the branches did not update when fetching a new Library Token, was that we actually did not fetch a new Library Token as we were using the function `setWithExpireIfNotExists`.

1. Instead of using `setWithExpireIfNotExists` we now use `setWithExpire`. This is done as we are already checking if a token is set elsewhere earlier in the flow, so we will only get to here if we want to bypass the token checker and force fetch a new token.

2. Instead of using a custom submit hook, we are now using a custom validate hook to check if the Agency ID provided on form_save is different to the already saved on. This way we avoid fetching a new Token unnecessary.

### DIFF
--- a/web/modules/custom/dpl_login/dpl_login.module
+++ b/web/modules/custom/dpl_login/dpl_login.module
@@ -203,12 +203,11 @@ function dpl_login_openid_connect_admin_settings_validate(array &$form, FormStat
 
   /** @var \Drupal\dpl_login\Adgangsplatformen\Config $adgangsplatformen_config */
   $adgangsplatformen_config = \Drupal::service('dpl_login.adgangsplatformen.config');
-  $agency_id = $adgangsplatformen_config->getAgencyId();
-  $values = $form_state->getValues();
+  $stored_agency_id = $adgangsplatformen_config->getAgencyId();
 
   // If agency_id is different from the already stored, we want to
   // force a new token to be created that belongs to the new agency.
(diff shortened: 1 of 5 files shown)


web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -199,21 +198,22 @@ function dpl_login_form_openid_connect_admin_settings_alter(array &$form, FormSt
...
 
   /** @var \Drupal\dpl_login\Adgangsplatformen\Config $adgangsplatformen_config */
   $adgangsplatformen_config = \Drupal::service('dpl_login.adgangsplatformen.config');
   $agency_id = $adgangsplatformen_config->getAgencyId();
+  $values = $form_state->getValues();
 
   // If agency_id is different from the already stored, we want to
   // force a new token to be created that belongs to the new agency.
-  if ($form_state->getValue('agency_id') != $agency_id) {
+  if ($values['clients']['adgangsplatformen']['settings']['agency_id'] != $agency_id) {
```
spaceo, line 211: What is the difference here?

Dresse: The difference is that if I am just calling `$form_state->getValue('agency_id')` it is always returning NULL. For some reason I have to explicitly go through the array to get the agency_id value.

spaceo (approved):
LGTM 👍

PR #713, opened 2024-02-01
Make sure that event paragraphs work same way as node paragraphs.

rasben:
#### Link to issue

#### Description

#### Screenshot of the result


LasseStaus (approved):
LGTM in general. 

In the CSS refactor we included more example content for the stories. This could be done here aswell, i there is a task for updating the content now. 

Maybe add a link/connection to the CSS changes in the PR / commit for later, even tho this should behave this way regardless.

PR #717, opened 2024-02-02
Logo svg

rasben:
Enable `dblog` module, to have access to system logs for debugging.

-----

So, it turns out that there is a bug in Drupal core, that completely stops you from uploading a SVG logo.
I've added a patch for fixing that.

Ontop of that, it turns out that Drupal crashes if you try to render an SVG with an image style. I've added a simple check, and if it is an SVG, we'll just render it as an original picture.

It alsoooo turns out that if no `width`/`height` values are set on the `img`, it just doesnt show up.
I've fixed that in the design system:

https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/471

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -175,7 +175,7 @@ tasks:
   dev:enable-dev-tools:
     desc: Enable dev modules and settings, which are not to be used in Prod. They are config-ignored
     cmds:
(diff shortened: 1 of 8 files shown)


kasperbirch1 (approved):
LGTM 👍

PR #722, opened 2024-02-02
Create 'limited' WYSIWYG format, and use it for event description.

rasben:
Some times, we want to have WYSIWYG fields for basic formatting, such as bold, links etc., but not advanced stuff like lists, tables and headings.
This adds a simplified WYSIWYG format, and adds it to the even description, as requested by the client.

DDFFORM-227

https://reload.atlassian.net/browse/DDFFORM-227


LasseStaus (approved):
i see the allowed tags has {limited} but is this set correctly? 

Spinning it up i get 
image

I guess the box shouldn't be there, and limited should just be seleted by default. 

Besides that it seems good.

PR #723, opened 2024-02-05
Set event reccurance as 'custom' by default. DDFFORM-217

rasben:
- Set event reccurance as 'custom' by default. DDFFORM-217
- Remove unused hook, left over after node => eventseries reformatting.

------

https://reload.atlassian.net/browse/DDFFORM-217


LasseStaus (approved):
LGTM. Nice cleanup with hook.

PR #724, opened 2024-02-05
update event list item

LasseStaus:
- Rewrite rendering for tag on eventinstance-list-teaser.

#### Description

Adjust the eventinstance--list-teaser to keep styling intact after <Tag> has been modified in the design system. 

#### Screenshot of the result

image


rasben (changes requested):
Instead of doing this directly in the eventinstance template, i think it would be better if you overwrite the field template (i guess it's called something like `field--event-categories.html.twig`?

That way, it'll work everywhere :)

PR #736, opened 2024-02-06
Ddfform 184 kategoriafgraensning pa arrangementsliste

rasben:


kasperg (approved):
:+1: The code looks excellent.

I have not tested it in practice.

PR #741, opened 2024-02-08
Add 1 hour by default to end date, when selecting start date.

rasben:
We've made some code that auto-fills the end date fields in a daterange input.
However, it is not valid to select an end date that has the same TIME as the start. To get around that, we'll bump time fields with 1 hour, by default.

DDFFORM-265


kasperbirch1 (approved):
LGTM 👍

PR #742, opened 2024-02-08
Re-factor header links to match design system. DDFFORM-220

rasben:
See https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/493


kasperbirch1 (approved):
LGTM 👍

PR #743, opened 2024-02-08
Admin ux

rasben:
- Re-order the order of fields in event form. DDFFORM-187
- Give admins right to edit novel theme settings. DDFFORM-188


kasperbirch1 (approved):
LGTM 👍

PR #744, opened 2024-02-08
Add validation against setting start and end date identical. DDFFORM-265

rasben:
The reccuring dates module has a bug, where if you set the start and end date to the same value, it completely crashes every list that the event series/instance shows up.
We'll add a custom check, to make sure this is not the case.

### DIFF
--- a/web/modules/custom/dpl_admin/dpl_admin.module
+++ b/web/modules/custom/dpl_admin/dpl_admin.module
@@ -5,6 +5,7 @@
  * Functionality for admin / editor UX.
  */
 
+use Drupal\Core\Datetime\DrupalDateTime;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Markup;
 use Drupal\Core\Url;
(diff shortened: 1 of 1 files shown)


kasperg:
The problem is not with the Recurring Events module but with the Date Range Formatter module that we use.

The stack trace provides a hint of this by showing what class the error occurs in. [The problem is that the date format used by the module if start and end date is the same `single` has been removed](https://www.drupal.org/project/date_range_formatter/issues/3309324#comment-15372390) and then [reintroduced again](https://www.drupal.org/project/date_range_formatter/issues/3151725).

As I see it we should be able to update from v4.0.1 to 4.0.2 of the module and recheck the configuration of our field formatters to fix this.

kasperg (changes requested):
The problem is not with the Recurring Events module but with the Date Range Formatter module that we use.

The stack trace provides a hint of this by showing what class the error occurs in. [The problem is that the date format used by the module if start and end date is the same `single` has been removed](https://www.drupal.org/project/date_range_formatter/issues/3309324#comment-15372390) and then [reintroduced again](https://www.drupal.org/project/date_range_formatter/issues/3151725).

As I see it we should be able to update from v4.0.1 to 4.0.2 of the module and recheck the configuration of our field formatters to fix this.

PR #749, opened 2024-02-12
Patch date_range_formatter: Avoid crash on identical start/end date.

rasben:
Patch date_range_formatter: Avoid crash on identical start/end date.

If an editor chooses a start and end date on the exact same day/time,
the actual display of the event instance will fail with a WSOD.
This both updates the module to the latest version, and adds
a patch. Logically, the patch should not be necessary, but no matter
what i do, the error still happens unless the patch is also added.

DDFFORM-265

------

I originally did a fix in https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/744 but that was cause i thought it was a direct issue with reccuring dates.
It turns out it was a problem with date_range_formatter, as @kasperg pointed out :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -235,6 +235,9 @@
                 "2259567: Support SVG files for theme logo setting": "https://www.drupal.org/files/issues/2023-05-08/2259567-168.patch",
                 "Always assume chmod succeeds": "patches/core-304fd8b-chmod-true.patch"
             },
+            "drupal/date_range_formatter": {
+                "3309324: Fails when start and end date are identical": "https://www.drupal.org/files/issues/2023-12-22/date_range_formatter_3309324_1.patch"
+            },
             "drupal/default_content": {
                 "3200212: Overwrite files on import": "https://git.drupalcode.org/project/default_content/-/merge_requests/8.patch"
             },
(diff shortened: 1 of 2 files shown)


kasperg (changes requested):
In #744 I wrote:

> The situation is a bit hairy, but as I see it we should be able to update from v4.0.1 to 4.0.2 of the module and recheck the configuration of our field formatters to fix this.

This PR updates the module *and* introduces a patch.

If there is a reason why it is not sufficient to update the module then I would expect some reasoning in the commit comment or in the pull request message.

rasben:
@kasperg :

1) I dont know why you experienced so many changes in the `composer.lock` before. All I did was update the module, and add the patch. Either way, I've re-done the same thing now, and the lock file looks cleaner.

2) I totally get what you mean, that the patch AND the update should not be necessary - but, no matter what i do, it only works with both.

kasperg (approved):
>  totally get what you mean, that the patch AND the update should not be necessary - but, no matter what i do, it only works with both.

I tried to update the events view and all display modes and even though I could get some configuration changes through I was still failing. We will just have to roll with the patch for now.

PR #757, opened 2024-02-13
DDFLSBP-399 - Added first version of form_node_confirm_form_alter tha…

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-399

#### Description

This PR adds a hook that checks if the node being deleted is referenced in configuration. If this is the case, it disables the "Delete" button and adds a informative text that helps locate where the node is being referenced in configuration. This also work when trying to delete multiple nodes.

#### Screenshot of the result
Single node deletion:
Screenshot 2024-02-21 at 01 52 17

Multiple node deletion:
Screenshot 2024-02-21 at 01 52 52

#### Other notes

The PR fails a cypress test, which needs to be resolved. I might need help solving it.


web/modules/custom/dpl_fees/dpl_fees.module
```diff
@@ -21,6 +21,18 @@ function dpl_fees_dpl_react_apps_data(array &$data): void {
...
+/**
+ * Implements hook_dpl_protected_nodes_get_protected_nodes().
+ */
+function dpl_fees_dpl_protected_nodes_get_protected_nodes(): array {
+  // Get the configuration form.
+  $form = \Drupal::formBuilder()->getForm('Drupal\dpl_fees\Form\FeesListSettingsForm');
+  // Get the configuration path.
+  $config_path = Url::fromRoute('dpl_fees.settings', [], ['absolute' => TRUE])->toString();
+
+  return dpl_protected_nodes_find_linkit_elements($form, $config_path);
```
spaceo, line 33: Wouldn't be an idea to change the parameters of `dpl_protected_nodes_find_linkit_elements()` to `$form_class_path`and `$config_route`. Since you are repeating the same pattern in the `hook_dpl_protected_nodes_get_protected_nodes` hooks?
Then the `dpl_protected_nodes_find_linkit_elements()` function could resolve the context by itself.

web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module
```diff
@@ -0,0 +1,112 @@
...
+ * Checks if node is protected and alters the delete confirm form accordingly.
+ *
+ * @param mixed[] $nodes
+ *   The node id to check.
+ * @param mixed[] $form
+ *   The form array to alter.
+ *
+ * @return mixed[]
+ *   The altered form array.
+ */
```
spaceo, line 52: @kasperg Made me aware of that since we are type hinting the function parameters and defining the return types of the functions  we do not need to replicate the documention of those in the dockblock. Does it make sense?

web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module
```diff
@@ -0,0 +1,112 @@
...
+ *   An associative array containing the structure of the form.
+ * @param array<mixed> $elements
+ *   An associative array containing the structure of the form.
+ *
+ * @return array<mixed>
+ *   An array containing linkit form element information.
+ *
+ * @throws \Safe\Exceptions\PcreException
+ */
+function dpl_protected_nodes_find_linkit_elements(array $form, string $config_path, array &$elements = []): array {
```
spaceo, line 92: See previous comment about dockblocks

web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module
```diff
@@ -0,0 +1,112 @@
...
+  // Pattern to match node url e.g. /node/1.
+  $pattern = '/^\/node\/\d+$/';
```
spaceo, line 94: Since you are only using `$pattern` once I think you should just write the pattern as a part of the `preg_match`function.

web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module
```diff
@@ -0,0 +1,112 @@
...
+          'node_id' => basename($value['#default_value']),
+          'config_title' => $value['#title'],
+          'config_path' => $config_path,
+        ];
+      }
+      else {
+        dpl_protected_nodes_find_linkit_elements($value, $config_path, $elements);
+      }
+    }
+  }
```
spaceo, line 109: I think it can be a bit tidier with less indentation.
Note: I did NOT test this :) : 
```php
foreach ($form as $key => $value) {
  // If the value is not an array, continue to the next iteration.
  if (!is_array($value)) {
    continue;
  }

  $field_value = $value['#default_value'] ?? '';
  $field_type = $value['#type'] ?? '';
  // If the value is a linkit element, add it to the elements array.
  if ($field_type === 'linkit' && preg_match('/^\/node\/\d+$/', $field_value)) {
    $elements[$key] = [
      'node_id' => basename($value['#default_value']),
      'config_title' => $value['#title'],
      'config_path' => $config_path,
    ];
  }
  else {
    // Otherwise try to find linkit elements.
    dpl_protected_nodes_find_linkit_elements($value, $config_path, $elements);
  }
}

**Review by spaceo — APPROVED** [review]
Looks good. I have some suggestions

File: web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module (hunk 2 above)
Comment (Dresse) [reply web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module:52]: Hmm, yes and no. I thought it was standard to add docblocks for all functions. If you don't and you omit @params, @return etc. then phpstan is complaining about it? Or am i misinterpreting what you meant?

File: web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module (hunk 3 above)
Comment (Dresse) [reply web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module:92]: Hmm, yes and no. I thought it was standard to add docblocks for all functions. If you don't and you omit @params, @return etc. then phpstan is complaining about it? Or am i misinterpreting what you meant?

File: web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module (hunk 5 above)
Comment (Dresse) [reply web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module:109]: I tested it out and it is working. I think you are right, it looks cleaner now. 👍

File: web/modules/custom/dpl_fees/dpl_fees.module (hunk 1 above)
Comment (Dresse) [reply web/modules/custom/dpl_fees/dpl_fees.module:33]: Just to make sure what you mean, you would like it to be something like this:

`
/**
 * Implements hook_dpl_protected_nodes_get_protected_nodes().
 */
function dpl_fees_dpl_protected_nodes_get_protected_nodes(): array {
  // Get the configuration form.
  $form = \Drupal::formBuilder()->getForm('Drupal\dpl_fees\Form\FeesListSettingsForm');
  // Get the configuration path.
  $config_path = Url::fromRoute('dpl_fees.settings', [], ['absolute' => TRUE])->toString();

  return dpl_protected_nodes_find_linkit_elements('Drupal\dpl_fees\Form\FeesListSettingsForm', 'dpl_fees.settings');
  `

File: web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module (hunk 2 above)
Comment (spaceo) [reply web/modules/custom/dpl_protected_nodes/dpl_protected_nodes.module:52]: No. We have a conflict between accepting no params/return in docblocks and a phpstan rule that wants a definition of the array type. I guess we just let params/returns with `mixed[]` be as they are...

File: web/modules/custom/dpl_fees/dpl_fees.module (hunk 1 above)
Comment (Dresse) [reply web/modules/custom/dpl_fees/dpl_fees.module:33]: I'll leave it as it is for now, as it will take some some time to figure out a better logic when calling the function recursively.

Comment (Dresse) [reply web/modules/custom/dpl_fees/dpl_fees.module]: Scratch my last comment. I found an easy way to solve this, which I think is quite nice. 
It will also make it possible to add more form/config paths from the same module if needed.

File: web/modules/custom/dpl_library_agency/dpl_library_agency.module [hunk 6]
```diff
@@ -4,3 +4,10 @@
  * @file
  * Contains dpl_library_agency.module.
  */
+
+/**
+ * Implements hook_dpl_protected_nodes_get_protected_nodes().
+ */
+function dpl_library_agency_dpl_protected_nodes_get_protected_nodes(): array {
+  return dpl_protected_nodes_get_context('Drupal\dpl_library_agency\Form\GeneralSettingsForm', 'dpl_library_agency.general_settings');
```
spaceo, line 12: Is it possible to refer to the class like this (without quotes)?:
```
Drupal\dpl_library_agency\Form\GeneralSettingsForm::class
```
In that way we can directly see if the class exists instead of using a string as indicator.

web/modules/custom/dpl_patron_page/dpl_patron_page.module
```diff
@@ -6,3 +6,10 @@
  *
  * Display patron page (user information page).
  */
+
+/**
+ * Implements hook_dpl_protected_nodes_get_protected_nodes().
+ */
+function dpl_patron_page_dpl_protected_nodes_get_protected_nodes(): array {
+  return dpl_protected_nodes_get_context('Drupal\dpl_patron_page\Form\PatronPageSettingsForm', 'dpl_patron_page.settings');
```
spaceo, line 14: Same comment as [previous](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/757/files#r1512384297)

web/modules/custom/dpl_patron_reg/dpl_patron_reg.module
```diff
@@ -4,3 +4,10 @@
  * @file
  * Dpl_patron_reg drupal module file.
  */
+
+/**
+ * Implements hook_dpl_protected_nodes_get_protected_nodes().
+ */
+function dpl_patron_reg_dpl_protected_nodes_get_protected_nodes(): array {
+  return dpl_protected_nodes_get_context('Drupal\dpl_patron_reg\Form\PatronRegSettingsForm', 'dpl_patron_reg.settings');
```
spaceo, line 12: Same comment as [previous](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/757/files#r1512384297)

spaceo (commented):
We are close now.
I have one comment regarding the `$form_class_path` argument sent to `dpl_protected_nodes_get_context()`...

web/modules/custom/dpl_patron_reg/dpl_patron_reg.module
Dresse: Implemented suggestion.

web/modules/custom/dpl_patron_page/dpl_patron_page.module
Dresse: Implemented suggestion.

web/modules/custom/dpl_library_agency/dpl_library_agency.module
Dresse: Implemented suggestion.

spaceo (approved):
LGTM 👍

PR #761, opened 2024-02-14
Add extra example content to content types

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-162

#### Description

Add extra example content to content types" in English

#### Screenshot of the result
image
image
image
image
image


web/modules/custom/dpl_example_content/content/eventseries/f6aafa27-a538-44a3-ae1f-02f8efbfb068.yml
```diff
@@ -0,0 +1,252 @@
...
+      end_value: '2030-01-01T13:49:30'
+      time: '08:00 am'
+      duration: 900
+      end_time: '08:00 am'
+      duration_or_end_time: duration
+      days: monday
+  monthly_recurring_date:
+    -
+      value: '2024-01-01T13:49:30'
+      end_value: '2025-01-01T13:49:30'
```
kasperg, line 71: Please reconsider this values

It is nice to have a recurring event to generate some instances to populate `/events`. Setting the year 2024 is not helpful. We risk over time that the dates pass and thus no longer shown on the list.

 I suggest setting a value in the far future where the recurring event will interleave with the existing single events.

kasperg (changes requested):
:+1: Nice with a recurring event but please reconsider the date.

kasperg (approved):
:+1: LGTM

PR #765, opened 2024-02-14
Add the "dpl-react-app-container--page" to twig templates

spaceo:
#### Description

We have to add the "dpl-react-app-container--page" to the templates in order to prevent layout jumping: the footer jumps up and down...


Dresse (approved):
LGTM

PR #767, opened 2024-02-15
Singular plural buttons distinction

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-172

#### Description
In this PR we adjust two text string definitions for when the user is deleting their queued reservations from the dashboard. There was no distinction between singular and plural form depending on how many reservations were deleted before, which is now fixed. 

#### Screenshot of the result
-

[Sibling PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/950)

### DIFF
--- a/web/modules/custom/dpl_react_apps/dpl_react_apps.module
+++ b/web/modules/custom/dpl_react_apps/dpl_react_apps.module
@@ -108,7 +108,13 @@ function dpl_react_apps_texts(): array {
         t('@count reservations were deleted', [], ['context' => 'Global']),
       ],
     ]),
-    'delete-reservation-modal-success-title' => t('You have deleted your reservation', [], ['context' => 'Global']),
+    'delete-reservation-modal-success-title' => json_encode([
+      'type' => 'plural',
+      'text' => [
+        t('Reservation deleted', [], ['context' => 'Global']),
+        t('Reservations deleted', [], ['context' => 'Global']),
+      ],
+    ]),
     'delete-reservation-modal-header' => json_encode([
       'type' => 'plural',
       'text' => [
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_react_apps/dpl_react_apps.module
```diff
@@ -213,7 +219,13 @@ function dpl_react_apps_texts(): array {
     'publizon-podcast' => t('Podcast', [], ['context' => 'Global']),
     'ready-for-loan-counter-label' => t('Ready', [], ['context' => 'Global']),
     'ready-for-loan' => t('Ready for pickup', [], ['context' => 'Global']),
-    'remove-all-reservations' => t('Remove reservations (@amount)', [], ['context' => 'Global']),
+    'remove-all-reservations' => json_encode([
+      'type' => 'plural',
+      'text' => [
+        t('Remove reservation (@amount)', [], ['context' => 'Global']),
```
clausbruun, line 225: Since the amount will always be "1" in the singular form, perhaps we should consider omitting the amount?
```suggestion
        t('Remove reservation', [], ['context' => 'Global']),
```

clausbruun (approved):
LGTM 🎉 However, I have a small suggestion that you might want to consider.

web/modules/custom/dpl_react_apps/dpl_react_apps.module
Adamik10: I see your point, but we provide this amount variable from the react apps when calling the `t()` function, so I think I'll just keep it in :)

clausbruun: Cool :)

PR #771, opened 2024-02-16
Fix display of materials in recommendation apps with custom content

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-280

#### Description

Currently materials are not shown when editors add a recommendation paragraph and provide a custom title and description. This occurs on Lagoon environments but not locally. No React app is rendered at the expected location at `{{ react_app }}` in `paragraph--recommendation--recommended-material.html.twig`.

It appears as if `dpl_paragraphs_preprocess_paragraph__recommendation()` is not run here.

To address this the handling of recommendations is refactored to remove the additional theme suggestion and instead implement the variation directly in the `paragraph--recommendation.html.twig` template.

Tests on Lagoon shows that this works.

#### Screenshot of the result

billede

I do not have an explanation for why this is the case or why it works locally but not on Lagoon. I have tried the following things to recreate the issue locally but without success:

- Bypassing Varnish and accessing the page through NGINX
- Changing the [`LAGOON_ENVIRONMENT_TYPE` ](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/839cddb25a3bc9d06f6ca54299d313616698d219/docker-compose.yml#L33-L33) to [not disable caching](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/839cddb25a3bc9d06f6ca54299d313616698d219/web/sites/default/local.settings.php#L11-L11)


web/themes/custom/novel/templates/paragraphs/paragraph--recommendation.html.twig
```diff
@@ -1 +1,14 @@
-{{ react_app }}
+{% if content.field_recommendation_title.0 or content.field_recommendation_description.0 %}
+  <div class="recommendation {{ react_app['#position_image_right'] ? 'recommendation--reversed"' : '' }}" >
+    {{ react_app }}
+    <a href="{{ react_app['#material_url'] }}" class="recommendation__texts arrow__hover--right-small">
+      <h3 class="recommendation__title">{{ content.field_recommendation_title }}</h3>
+      <p class="recommendation__description">
+        {{ content.field_recommendation_description }}
+      </p>
+      {{ source(directory ~ '/assets/dpl-design-system/icons/arrow-ui/icon-arrow-ui-small-right.svg') }}
```
rasben, line 9: I'm pretty sure that we have a variable specifically for the icon directory :)

rasben (approved):
A small comment, nothing critical

PR #776, opened 2024-02-16
Dialog UI fixes. DDFFORM-270 DDFFORM-299

rasben:
This is particularly visible when looking at a "node preview" page, and trying to click on any frontend link.


LasseStaus (approved):
Nice one LGTM!

PR #777, opened 2024-02-18
Event API spec

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-305

#### Description

This is the first stab at providing a specification and some documentation which will allow third party systems to retrieve all events and update selected fields on some of them.

No logic has been implemented yet. This is format and documentation only. The purpose is to provide something that third parties can comment on.

We currently have no defined strategy for versioning the API. Given what is supported by the Drupal REST module my suggestion is that we add a version element to paths. Note that this should be added to other REST resources as well.


rasben (approved):
Looks good, but could use a swagger

kasperg:
We have gotten some feedback from external parties now so this can be closed.

PR #782, opened 2024-02-20
Dont show event instance overview on event series page. DDFFORM-310

rasben:
Before this commit, we displayed the event instances on the bottom of the event series view, as a slider.
However - this confuses the users a lot, and the reason we use a slider is cause nothing else was ready at the time. We'll disable it for now, and find some more intuitive way of displaying it in the future.


LasseStaus (approved):
LGTM 👍

PR #785, opened 2024-02-20
Ddfform 297 stack visning af event serier

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-297

#### Description
This Pull Request introduces a new display mode named Stacked Event along with the logic. When multiple events from the same series occur consecutively without any intervening events, we aim to avoid repetitive display of the teaser card. Therefore, we have implemented a 'Stacked Event' link that will only show the date and time of these events.

#### Screenshot of the result
image

### DIFF
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -312,6 +312,39 @@ function dpl_event_preprocess_eventinstance__list_teaser(array &$variables): voi
 }
 
 /**
+ * Implements hook_preprocess_eventinstance__stacked_event().
+ *
+ * Preprocesses variables for the 'list_teaser' view mode of eventinstance.
+ * - Generating the URL for the full view of the event instance.
+ * - Processing start and end dates from the 'date' field and formatting them
+ *   in the site's timezone.
+ */
+function dpl_event_preprocess_eventinstance__stacked_event(array &$variables): void {
+  $eventInstance = $variables['eventinstance'];
+  /** @var \Drupal\Core\Datetime\DateFormatterInterface $date_formatter */
+  $date_formatter = \Drupal::service('date.formatter');
+
+  // Extract the event url.
+  $variables['eventinstance_url'] = Url::fromRoute('entity.eventinstance.canonical', ['eventinstance' => $eventInstance->id()])->toString();
+
+  // Extract start_time, end_time and datetime attributes.
+  if (!$eventInstance->get('date')->isEmpty()) {
+    $date_field = $eventInstance->get('date')->first();
+
+    $start_time = $date_field->start_date;
+    $end_time = $date_field->end_date;
+
+    if ($start_time instanceof DrupalDateTime) {
+      $variables['start_time'] = $date_formatter->format($start_time->getTimestamp(), 'custom', 'H:i');
+      $variables['datetime_attribute'] = $date_formatter->format($start_time->getTimestamp(), DATE_ATOM);
+    }
+    if ($end_time instanceof DrupalDateTime) {
+      $variables['end_time'] = $date_formatter->format($end_time->getTimestamp(), 'custom', 'H:i');
+    }
+  }
+}
+
+/**
  * Implements hook_form_FORM_ID_alter().
  *
  * Altering the form for EDITING (not creating) an event series.
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -294,19 +294,19 @@ function dpl_event_preprocess_eventinstance__list_teaser(array &$variables): voi
     }
   }
 
-  // Extract start_date, end_date and datetime attributes.
+  // Extract start_time, end_time and datetime attributes.
   if (!$eventInstance->get('date')->isEmpty()) {
     $date_field = $eventInstance->get('date')->first();
 
-    $start_date = $date_field->start_date;
-    $end_date = $date_field->end_date;
+    $start_time = $date_field->start_date;
+    $end_time = $date_field->end_date;
 
-    if ($start_date instanceof DrupalDateTime) {
-      $variables['start_date'] = $date_formatter->format($start_date->getTimestamp(), 'custom', 'H:i');
-      $variables['datetime_attribute'] = $date_formatter->format($start_date->getTimestamp(), DATE_ATOM);
+    if ($start_time instanceof DrupalDateTime) {
+      $variables['start_time'] = $date_formatter->format($start_time->getTimestamp(), 'custom', 'H:i');
+      $variables['datetime_attribute'] = $date_formatter->format($start_time->getTimestamp(), DATE_ATOM);
     }
-    if ($end_date instanceof DrupalDateTime) {
-      $variables['end_date'] = $date_formatter->format($end_date->getTimestamp(), 'custom', 'H:i');
+    if ($end_time instanceof DrupalDateTime) {
+      $variables['end_time'] = $date_formatter->format($end_time->getTimestamp(), 'custom', 'H:i');
     }
   }
 }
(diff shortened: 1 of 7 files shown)


web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -294,19 +294,52 @@ function dpl_event_preprocess_eventinstance__list_teaser(array &$variables): voi
...
+
+    $start_time = $date_field->start_date;
+    $end_time = $date_field->end_date;
+
+    if ($start_time instanceof DrupalDateTime) {
+      $variables['start_time'] = $date_formatter->format($start_time->getTimestamp(), 'custom', 'H:i');
+      $variables['datetime_attribute'] = $date_formatter->format($start_time->getTimestamp(), DATE_ATOM);
+    }
+    if ($end_time instanceof DrupalDateTime) {
+      $variables['end_time'] = $date_formatter->format($end_time->getTimestamp(), 'custom', 'H:i');
```
kasperg, line 342: Please check if you can move extraction of url and dates to a new preprocess function which is used across view modes: `dpl_event_preprocess_eventinstance()`.

The conent seems to be partly duplicated from what is going on in `dpl_event_preprocess_eventinstance__list_teaser()`.

web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -294,19 +294,52 @@ function dpl_event_preprocess_eventinstance__list_teaser(array &$variables): voi
...
+ * - Processing start and end dates from the 'date' field and formatting them
+ *   in the site's timezone.
+ */
+function dpl_event_preprocess_eventinstance__stacked_event(array &$variables): void {
+  $eventInstance = $variables['eventinstance'];
+  /** @var \Drupal\Core\Datetime\DateFormatterInterface $date_formatter */
+  $date_formatter = \Drupal::service('date.formatter');
+
+  // Extract the event url.
+  $variables['eventinstance_url'] = Url::fromRoute('entity.eventinstance.canonical', ['eventinstance' => $eventInstance->id()])->toString();
```
kasperg, line 328: This is an example of a comment that adds very little value.

web/themes/custom/novel/novel.theme
```diff
@@ -276,3 +277,43 @@ function novel_theme_suggestions_select_alter(array &$suggestions, array $variab
...
+ */
+function novel_preprocess_views_view_unformatted(&$variables): void {
+  $rows = &$variables['rows'];
+
+  foreach ($rows as $index => &$row) {
+    if (isset($row['content']['#eventinstance'])) {
+      /** @var \Drupal\recurring_events\Entity\EventInstance $eventInstance */
+      $eventInstance = $row['content']['#eventinstance'];
+
+      if ($eventInstance instanceof EventInstance) {
```
kasperg, line 304:
```suggestion
      $eventInstance = $row['content']['#eventinstance'];

      if ($eventInstance instanceof EventInstance) {
```

The type hint should not be necessary/valuable when a subsequent line actually ensures that the value is of the expected type.

web/themes/custom/novel/templates/views/eventinstance--stacked-event.html.twig
```diff
@@ -0,0 +1,8 @@
+
+<a href="{{ eventinstance_url }}" class="event-list-item-stacked arrow__hover--right-small">
+  <time class="event-list-item-stacked__content" datetime="{{ datetime_attribute }}">
+    <span class="event-list-item-stacked__date">{{ content.date }}</span>
+    <span class="event-list-item-stacked__time">{{ start_time }} - {{ end_time }}</span>
+  </time>
+  {{ source(directory ~ '/assets/dpl-design-system/icons/arrow-ui/icon-arrow-ui-small-right.svg') }}
```
kasperg, line 7: Please use the `baseIconPath` variable like [elsewhere](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/f7f7dce01cfa4a4d4dd10f50e1b7d9cfca0ed009/web/themes/custom/novel/templates/components/nav-spot.html.twig#L28-L28).

web/themes/custom/novel/novel.theme
```diff
@@ -276,3 +277,43 @@ function novel_theme_suggestions_select_alter(array &$suggestions, array $variab
...
+ * Preprocesses rows in a Drupal view to handle the display of event instances.
+ *
+ * This function iterates over each row in the view, checking for the presence
+ * of an EventInstance object.
+ * When an EventInstance is found, it checks if it belongs to the same
+ * event series as the previous row.
+ * If so, it marks the current row as 'stacked' and adjusts its view
+ * mode accordingly.
+ * This is useful for grouping events from the same series in the view,
+ * enhancing visual continuity and clarity.
```
kasperg, line 291: This smells like a CoPilot commenting :smile:. It is not really helpful except for the last line. It tries to get close to why we are doing this.

web/themes/custom/novel/novel.theme
```diff
@@ -276,3 +277,43 @@ function novel_theme_suggestions_select_alter(array &$suggestions, array $variab
...
+    if (isset($row['content']['#eventinstance'])) {
+      /** @var \Drupal\recurring_events\Entity\EventInstance $eventInstance */
+      $eventInstance = $row['content']['#eventinstance'];
+
+      if ($eventInstance instanceof EventInstance) {
+        $eventSeriesId = $eventInstance->getEventSeries()->id();
+
+        // Check if not the first row and if event_series_id matches
+        // the previous row.
+        $isStacked = $index > 0 && $eventSeriesId == $rows[$index - 1]['content']['#eventinstance']->getEventSeries()->id();
```
kasperg, line 309:
```suggestion
      $eventInstance = $row['content']['#eventinstance'];
      $previousInstance = ($index > 0) ? $rows[$index - 1]['content']['#eventinstance'] : NULL;
      if ($eventInstance instanceof EventInstance && $previousInstance instanceof EventInstance) {
        $isStacked = $eventInstance->getEventSeries()->getId() == $previousInstance->getEventSeries()->id();
```

Consider restructuring the code. In my opinion this version above (totally untested!) is preferable:

1. It adds proper type checking for the previous row
2. It eliminates unnecessary variables
3. It is shorter without sacrificing clarity

kasperg (changes requested):
:+1: Nice but I think there is room for improvement.

web/themes/custom/novel/templates/views/eventinstance--stacked-event.html.twig
```diff
@@ -0,0 +1,8 @@
+
+<a href="{{ eventinstance_url }}" class="event-list-item-stacked arrow__hover--right-small">
+  <time class="event-list-item-stacked__content" datetime="{{ datetime_attribute }}">
+    <span class="event-list-item-stacked__date">{{ content.date }}</span>
+    <span class="event-list-item-stacked__time">{{ start_time }} - {{ end_time }}</span>
+  </time>
```
kasperg, line 6: Consider adding a more explicit title.

Currently the stacked event is a separate list item in the DOM but the content of the link within has no information about what the item is about. It is simply a date.

One approach to this would be to add [a visually hidden element](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/3308fb2d6640c71e5b44ded179bf405360b82b61/src/styles/scss/tools/mixins.misc.scss#L11-L11) containing the event instance title.

web/themes/custom/novel/novel.theme
kasperbirch1: There is something about this suggested change that breaks the code 😅

kasperg: Without more info it is hard to tell whats wrong. I guess we will just have to leave it as is.

kasperg (approved):
:+1: LGTM

PR #790, opened 2024-02-21
Make sure that a uregistered user is using right token when logging out

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-403

#### Description

When an unregistered user logs out we need to make sure that the unregeistered user token is being used.


Dresse (approved):
LGTM, but tests needs to be fixed. I approve for now.

PR #792, opened 2024-02-22
Refactor and streamline the hero elements.

rasben:
Up until now, the same classes and components were being used almost identically, across several elements.
This PR creates a unified way of adding heros, both with and OA without links.

Depends on https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/518


kasperbirch1 (approved):
LGTM 👍

PR #797, opened 2024-02-22
Move library token creation to submit handler

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-418

#### Description

When token creation runs in the validate handler the wrong client is being used. By moving it to the submit handler we ensure that the new client configuration has been saved before the token creation.


Dresse (approved):
LGTM

PR #798, opened 2024-02-22
New node CT: "Branch". DDFFORM-224

rasben:
- Create new node type, branch, that defines the physical (or virtual) branches of a library.
- Setup permissions, so editors and above can handle them.
- Setup a reference field on articles and pages.
- Setup a basic full shown page, with the new hero.

See it in action here:

https://varnish.pr-784.dpl-cms.dplplat01.dpl.reload.dk/node/1

The varnish environment wont let me upload images for some reason, but, it should work. It works locally, atleast.


kasperbirch1 (approved):
LGTM 👍

PR #804, opened 2024-02-23
Use release/2024-9-1 version of design system & react packages

Adamik10:
Use release/2024-9-1 version of design system & react packages


Dresse (approved):
LGTM

PR #816, opened 2024-02-27
Allow editor to display a page title / subtitle. DDFFORM-374

rasben:


web/themes/custom/novel/templates/layout/node--page--full.html.twig
```diff
@@ -1,3 +1,14 @@
...
+  {% if node.hasField('field_display_titles') and
+        node.get('field_display_titles').getString() == '1' %}
+    <header class="article-header">
+      <h1 class="article-header__title">
+      {{ content.field_hero_title.0|default(label) }}
+      </h1>
+
+      {{ content.field_subtitle }}
+    </header>
+  {% endif %}
```
kasperg, line 11: Please consider the approach with rendering the title conditionally.

Accessibility requires us to have a proper tree of heading levels. If we do not render the page title as a `<h1>` we need some other component (paragraph?) to have that `<h1>` heading. We do not support this at the moment and in my opinion we do not want to add that option/configurability to Paragraphs either.

If we do not want to display the title then I suggest we render it but [hide it visually using CSS](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/6e5f558e015c59f3f54bc632b9031d0c24f29a6c/src/styles/scss/tools/mixins.misc.scss#L1).

config/sync/core.entity_form_display.node.page.default.yml
```diff
@@ -18,19 +21,34 @@ mode: default
...
+    settings:
+      display_label: true
+    third_party_settings: {  }
+  field_hero_title:
+    type: string_textarea
+    weight: 2
+    region: content
+    settings:
+      rows: 2
+      placeholder: 'If empty, the regular page title will be shown..'
```
kasperg, line 47:
```suggestion
      placeholder: 'If empty, the regular page title will be shown.'
```

config/sync/core.entity_view_display.node.page.default.yml
```diff
@@ -4,7 +4,10 @@ status: true
 dependencies:
   config:
     - field.field.node.page.field_branch
+    - field.field.node.page.field_display_titles
+    - field.field.node.page.field_hero_title
     - field.field.node.page.field_paragraphs
+    - field.field.node.page.field_subtitle
```
kasperg, line 10: I wonder why you went with the approach with adding a separate title field and a subtitle field.

It might be good but I cannot tell whether this is what was actually requested. Personally I had expected a single toggle field.

kasperg (changes requested):
:+1: but please give this a second round of thought regarding how this is rendered.

config/sync/core.entity_view_display.node.page.default.yml
rasben: I'm not sure what you mean by "seperate"

However, maybe this answers your question:

The standard label field on a node, is still used by the `<title>` tag, and the admin content search.
For example, it might make sense to have the frontpage named "frontpage", but you will probably not want to show that on the actual page.

As for the subtitle - the design has the subtitle as part of it, and we already had a "field_subtitle" used on articles, so i figured i'd re-use it, so we can also use it in teasers and what not

Screenshot 2024-02-27 at 13 44 18

config/sync/core.entity_form_display.node.page.default.yml
rasben: It was actually on purpose, as i generally add ".." in placeholder fields, but it's not a must for me, so i'll just change it

web/themes/custom/novel/templates/layout/node--page--full.html.twig
rasben: Fair enough, I've updated it now:

https://github.com/danskernesdigitalebibliotek/dpl-cms/compare/5180e26e55e26b3137744919ca38a2b8d316e8ab..51bf549a32a7966ca8122c1d84f57a2b5fd1a685

config/sync/core.entity_form_display.node.page.default.yml
Dresse: I was wondering what this means. Doesn't the logic in the template either show the field_hero_title or nothing at all?

kasperg: > [...] i generally add ".." in placeholder fields

OK. I would expect either a `.` or a `...`

config/sync/core.entity_view_display.node.page.default.yml
kasperg: > The standard label field on a node, is still used by the <title> tag, and the admin content search.
For example, it might make sense to have the frontpage named "frontpage", but you will probably not want to show that on the actual page.

Separate as in separate from the standard title field.

I can understand that people might want to separate the administration title from the displayed title but the fact that with this change the administration title is still used in the `<title>` tag means that it is still not totally hidden to users.

To me it would be more consistent to tell users to use titles that they actually want to show to end users whether it is (partially) hidden from view or not.

Having alternate titles will cause complexity in other situations such e.g. user facing views.

> As for the subtitle - the design has the subtitle as part of it, and we already had a "field_subtitle" used on articles, so i figured i'd re-use it, so we can also use it in teasers and what not

Personally I had imagined the subtitle being a paragraph. Perhaps that not good enough.

I understand your point about subtitles in lists but with this approach I think that is somewhat confusing that the only way you can get a subtitle into lists is by enabling the "Display titles" option.

kasperg (approved):

PR #809, opened 2024-02-26
Branch list and example content. DDFFORM-236

rasben:
Depends on https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/527

Can see in action here: (although, with messed up styling until the other PR has been merged)

https://varnish.pr-809.dpl-cms.dplplat01.dpl.reload.dk/branches

PR #819, opened 2024-02-27
Refactor / Generalize event-item-* / event-list-page. DDFFORM-326

rasben:
In the future, we'll want to use these styles for other lists, not limited to events. This updates the class names and structure, to match what is upcoming in the design-system.

----

Originally, a lot of this was done in https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/809 but I decided to pull it out in its own seperate PR.
This is basically a preparation update - it doesnt make any differences for how things are displayed nor functionality, but it cleans up the items, and makes them reusable.

-----

Relies on https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/527 being merged, until then it will look like a mess.


web/modules/custom/dpl_event/dpl_event.module
kasperg, line 1: :+1: for this refactor. Using the actual field should yield a better result.

web/themes/custom/novel/templates/components/list-teaser.html.twig
```diff
@@ -0,0 +1,49 @@
...
+  </div>
+  <div class="content-list-item__content">
+    {{ categories }}
+    {% if date %}
+      <div class="content-list-item__date">
+        {{ date }}
+      </div>
+    {% endif %}
+
+    <h3 class="content-list-item__title">{{ title }}</h3>
```
kasperg, line 20: Please make this a `<h2>`.

With regard to accessibility and with the latest changes to the design system this should be right heading level.

web/themes/custom/novel/templates/components/list-teaser-stacked.html.twig
```diff
@@ -0,0 +1,12 @@
+<a href="{{ url }}" class="content-list-item-stacked arrow__hover--right-small">
+  <h3 class="hide-visually">{{ title }}</h3>
```
kasperg, line 2: Please also make this a `<h2>`.

web/themes/custom/novel/templates/views/views-view--content-list-page.html.twig
kasperg, line 1: As mentioned in #809 I wonder if this should be in `components` since it is a template that is purely used for inclusion.

web/themes/custom/novel/templates/components/list-teaser.html.twig
kasperg, line 1: Consider renaming this `list-item`.

It would translate nicely into the BEM block name. This also goes for the stacked version.

web/themes/custom/novel/templates/components/list-teaser-stacked.html.twig
```diff
@@ -0,0 +1,12 @@
+<a href="{{ url }}" class="content-list-item-stacked arrow__hover--right-small">
```
kasperg, line 1: This uses a different approach to adding attributes compared to `list-teaser`.

I wonder if it is intentional.

kasperg (approved):
:+1: This is good.

I think a few things needs changing but I trust you to proceed and merge as you see fit.

PR #820, opened 2024-02-28
Content lock

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-294

#### Description

This adds and configured the content lock module to support locking of content and preventing simultaneous editing. Everyone can break a lock to prevent situations where editors are stuck due to someone forgetting to release locks.

Locks are displayed in the admin view to surface information about who edits what easily.

Please see the individual commits for additional commentary.

#### Screenshot of the result

billede

Monosnap Content | DPL CMS | Logged in 2024-02-28 09-42-42

- Even though events also support locking information about locks is not shown in this view as the view also does not show the current author.
- Should we consider locking entities like media, taxonomy and menu items?

### DIFF
--- a/web/modules/custom/dpl_admin/dpl_admin.module
+++ b/web/modules/custom/dpl_admin/dpl_admin.module
@@ -64,35 +64,36 @@ function dpl_admin_menu_local_actions_alter(&$local_actions) : void {
 
 /**
  * Implements template_preprocess_views_view_field().
- *
- * Preprocesses the output of a field in the "event_admin" view.
- * Altering the output to generate a unique indicator with a link to the
- * eventseries.
- *
- * {@inheritDoc}
  */
 function dpl_admin_preprocess_views_view_field(array &$variables): void {
-  if (!isset($variables['view'], $variables['field'], $variables['output'])) {
+  if (
+    !isset($variables['view'], $variables['field']) ||
+    !$variables['view'] instanceof ViewExecutable ||
+    !$variables['field'] instanceof EntityField
+  ) {
     return;
   }
 
-  $view = $variables['view'];
-  $field = $variables['field'];
-  $output = $variables['output'];
-
-  if (!$view instanceof ViewExecutable || $view->id() !== 'event_admin') {
-    return;
+  if ($variables['view']->id() == 'event_admin' &&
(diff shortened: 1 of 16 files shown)


web/modules/custom/dpl_admin/dpl_admin.module
```diff
@@ -103,6 +109,36 @@ function _dpl_admin_preprocess_views_view_field_event_admin_eventseries_id(array
   $variables['output'] = Markup::create($markup);
 }
 
+/**
+ * Preprocess the author field in the "content" admin view.
```
LasseStaus, line 113: I couldn't see any visual indication that an Eventseries/instance was locked, when it had been locked. 

If content-lock is also applied to events, i think this should also be the case for the event admin view. 

Content from the "content" admin view will surely be edited more, but i think if there is a visual indication for that view, the same should apply if content-lock as used elsewhere.

LasseStaus (approved):
This is nice, and works as described in the ticket when i test it with different users.  🚀 

In regards to your comments:

- I think it's unfortunate that "locked by X" is not shown for events, when it is also automatically locked the same way as other content. I saw your comment about the view.

- I don't think this should be enabled for anything else, unless we see it is needed later on. 

Personal takes that are potentially out of scope: 

- I don't understand the "unlock" button being shown for the editor currently editing, since i don't know when that would be pressed. Instead it would make sense to me the "save" said "Save and unlock" instead. 

- I would prefer if "break lock" was button instead of the status message, or both were present. I think generally that all these status message displays are a bit suboptimal for handling this. 

 - When testing this, i was thinking whether it would be better to actively activate a lock, rather than it automatically being applied. I can see a lot of unncessary locking and unlocking with this approach, even tho this is what is specified.

kasperg:
> I don't understand the "unlock" button being shown for the editor currently editing, since i don't know when that would be pressed. Instead it would make sense to me the "save" said "Save and unlock" instead.

The appropriate situation usage of the unlock button is when you navigated to the edit form but decide not to change anything anyway. In this situation it would not be appropriate to click "Save".

The current implementation allows editors to navigate away from an editing page but retain the lock.

> I was thinking whether it would be better to actively activate a lock, rather than it automatically being applied

I was thinking the same thing. At least in this regard it makes sense to allow everybody to break locks.

PR #824, opened 2024-02-28
Remove required for Alt text on images

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-301

#### Description
Images that are only for decorative purposes should be allowed to not have an alt text.

I tried to make some dynamic toggle behavior here. but with no success: 
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/823

#### Screenshot of the result
image

image


kasperg (approved):
:+1: LGTM

PR #825, opened 2024-02-28
Add opening hours link to General Settings Form and display it in the header - Ddfform 370

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-370

#### Description
This pull request introduces a feature to include a link to the opening hours in the General Settings Form, which is then displayed in the website's header

I have also updated the design-system 
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/539

#### Screenshot of the result
https://github.com/danskernesdigitalebibliotek/dpl-cms/assets/49920322/81f6c864-d2d6-4028-8fcd-e197a73acd52

### DIFF
--- a/web/modules/custom/dpl_library_agency/src/GeneralSettings.php
+++ b/web/modules/custom/dpl_library_agency/src/GeneralSettings.php
@@ -23,6 +23,7 @@ class GeneralSettings extends DplReactConfigBase {
   const EREOLEN_MY_PAGE_URL = 'https://ereolen.dk/user/me';
   const EREOLEN_HOMEPAGE_URL = 'https://ereolen.dk';
   const FBI_PROFILE = 'next';
+  const OPENING_HOURS_URL = '';
 
   /**
    * Gets the configuration key for general settings.
--- a/web/modules/custom/dpl_library_agency/src/GeneralSettings.php
+++ b/web/modules/custom/dpl_library_agency/src/GeneralSettings.php
@@ -23,7 +23,7 @@ class GeneralSettings extends DplReactConfigBase {
   const EREOLEN_MY_PAGE_URL = 'https://ereolen.dk/user/me';
   const EREOLEN_HOMEPAGE_URL = 'https://ereolen.dk';
   const FBI_PROFILE = 'next';
-  const OPENING_HOURS_URL = '';
+  const OPENING_HOURS_URL = '/branches';
 
   /**
    * Gets the configuration key for general settings.
--- a/web/themes/custom/novel/novel.theme
+++ b/web/themes/custom/novel/novel.theme
@@ -7,7 +7,6 @@
 
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
-use Drupal\dpl_library_agency\GeneralSettings;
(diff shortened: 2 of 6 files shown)


web/themes/custom/novel/novel.theme
```diff
@@ -122,6 +123,9 @@ function novel_preprocess_dpl_react_app__loan_list(array &$variables): void {
 function novel_preprocess(array &$variables): void {
   $directory = $variables['directory'] ?? 'web/themes/custom/novel';
   $variables['baseIconPath'] = "/{$directory}/assets/dpl-design-system/icons";
+
+  $config = \Drupal::config('dpl_library_agency.general_settings');
+  $variables['opening_hours_url'] = $config->get('opening_hours_url') ?? GeneralSettings::OPENING_HOURS_URL;
```
kasperg, line 128: Please use another process function for this. 

Using *_preprocess() will add the variable to _all_ templates. That is probably not what you intended. Also consider moving the preprocess function to the `dpl_library_agency` module as that is the module responsible for the configuration.

web/modules/custom/dpl_library_agency/src/GeneralSettings.php
```diff
@@ -23,6 +23,7 @@ class GeneralSettings extends DplReactConfigBase {
   const EREOLEN_MY_PAGE_URL = 'https://ereolen.dk/user/me';
   const EREOLEN_HOMEPAGE_URL = 'https://ereolen.dk';
   const FBI_PROFILE = 'next';
+  const OPENING_HOURS_URL = '';
```
kasperg, line 26: Consider making `/branches` the default value.

#809 will create a list of branches which in turn will link to the individual branch with the corresponding opening hours. I think that would be a sensible default value.

web/themes/custom/novel/templates/layout/header.html.twig
```diff
@@ -38,7 +38,9 @@
         </div>
         <div class="header__clock-items">
             <img loading="lazy" width="58" height="58" src="/themes/custom/novel/assets/dpl-design-system/icons/basic/icon-watch-static.svg" class="mb-8" alt="clock icon"/>
-            <span class="text-small-caption">Åbningstider</span>
+            {% if opening_hours_url %}
+              <a href="{{ opening_hours_url }}" class="link-tag text-small-caption">{{ "Open hours"|t }}</a>
+            {% endif %}
```
kasperg, line 43: I wonder if it is a good idea to toggle the text based on the value being present.

kasperg, line 43:
```suggestion
            {% if opening_hours_url %}
              <a href="{{ opening_hours_url }}" class="link-tag text-small-caption">{{ "Opening hours"|t }}</a>
            {% endif %}
```

:+1: for remembering to translate the "Åbningstider" string but I do not think you got the English text right.

kasperg, line 43: Please consider making the entire element a link instead of just the text.

Note that this may introduce corresponding changes in the design system. So be it.

kasperg (changes requested):
:+1: Nice. A few elements need a second look.

kasperg (approved):

PR #835, opened 2024-03-01
Patch paragraphs_ee, to have arrow buttons.

rasben:
Someone else ended up making a better patch than me :(

Screenshot 2024-03-01 at 08 36 34

It's not perfect, visually, but it's better than the current solution.


kasperbirch1 (approved):
LGTM 👍

PR #836, opened 2024-03-01
Move static generation cmd to drush

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-459

#### Description

Move static generation cmd to drush
So we are able to detect if we have already run the cmd in the cmd itself.


Dresse (approved):
LGTM

PR #837, opened 2024-03-01
Changed langcode of generated content from en to da

Dresse:
Changed langcode of generated content from en to da


spaceo (approved):
LGTM 👍 
I have no overview over if you catched them all, but I trust it :)

PR #843, opened 2024-03-04
Move event description from WYSIWYG to plain textarea. DDFFORM-298

rasben:
- Add warning on existing WYSIWYG field, and hide it from editors
- Setup basic field_description text area field
- Create deploy hook, moving data from WYSIWYG to textarea
- Delete existing data in WYSIWYG
- In a future deploy, we'll delete the WYSIWYG field.

### DIFF
--- a/web/modules/custom/dpl_event/dpl_event.deploy.php
+++ b/web/modules/custom/dpl_event/dpl_event.deploy.php
@@ -1,0 +1,56 @@
+<?php
+
+/**
+ * @file
+ * Event deploy hooks.
+ *
+ * These get run AFTER config-import.
+ */
+
+use Drupal\Core\Entity\FieldableEntityInterface;
+
+/**
+ * Pre-populate data to new non-WYSIWYG field field_description.
+ *
+ * We also empty out the old field. The field has been set to be hidden
+ * from the editors, and we'll delete it in a future deploy.
+ */
+function dpl_event_deploy_port_description_fields() : string {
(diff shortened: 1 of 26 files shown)


web/modules/custom/dpl_event/dpl_event.deploy.php
```diff
@@ -0,0 +1,56 @@
...
+  $message = _dpl_event_port_wysiwyg('eventseries', 'field_event_description', 'field_description');
+  $message .= _dpl_event_port_wysiwyg('eventinstance', 'field_event_description', 'field_description');
+
+  return $message;
+}
+
+/**
+ * A helper function, for porting WYSIWYG fields to plain textareas.
+ */
+function _dpl_event_port_wysiwyg(string $entity_type, string $source_field, string $target_field): string {
```
kasperg, line 28: Such reusable helper functions are nice but placing it in the `dpl_event` module makes it hard to reuse in other modules.

In the future we should consider moving it to somewhere shareable - perhaps in the profile?

kasperg (approved):
:+1: Nicely done!

Do note that I do not think this ticket can be closed without providing the second PR which will actually delete the obsolete field.

web/modules/custom/dpl_event/dpl_event.deploy.php
rasben: Right now, i made it as a helper function as i needed to do it on two fields.

If this is something we end up needing again, we can always move it out of dpl_event

PR #844, opened 2024-03-04
Simple sorting of branches: Promote branch to top of lists. DDFFORM-237

rasben:
This is done by a simple checkbox, essentially making two lists, both sorted by ABC:
Promoted and non-promoted.


kasperg (approved):
:+1: The code here looks fine but a comment during our meeting with the business again mentioned a request to be able to change the individually.

I suggest that you verify whether this approach is acceptable before we merge it.

PR #847, opened 2024-03-04
DDFFORM-367 - Refactor event URL title based on ticket price.

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-367

#### Description
This pull request refactors the event URL title based on the ticket price. If the tickets are free ('0.00' kr.), it changes the title to 'Get Tickets' instead of 'Buy Tickets'.

I have assumed that if there is only one of the ticket types that is free, we will switch to get tickets

#### Screenshot of the result
image
image
image

### DIFF
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -174,7 +174,26 @@ function dpl_event_preprocess_field(array &$variables): void {
   // After determining the 'event_is_active' flag, check if the event is active.
   if ($variables['event_is_active']) {
     // If active, set a translatable string as the title.
-    $variables['items'][0]['content']['#title'] = t('Buy tickets');
+    $price_formatter = \Drupal::service('dpl_event.price_formatter');
+    $hasFreeTickets = FALSE;
+
+    if ($entity->hasField('event_ticket_categories')) {
+      $ticket_categories_fields = $entity->get('event_ticket_categories')
+        ->referencedEntities();
+
+      if ($ticket_categories_fields) {
+        foreach ($ticket_categories_fields as $ticket_category) {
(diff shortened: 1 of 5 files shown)


web/modules/custom/dpl_example_content/content/eventseries/2ce9d9b0-7f10-4a17-9a5e-383d36c1ed93.yml
```diff
@@ -118,6 +118,11 @@ default:
   field_event_image:
     -
       entity: a0176f8c-bb7e-4f5a-bf23-b48cd8bab05f
+  field_event_link:
+    -
+      uri: 'https://www.google.com/'
```
kasperg, line 123: Consider using a more appropriate link like https://place2book.com/ or https://billetto.dk/ for these links.

I know we have been tossing around google.com in other cases but now that we store in content that is reused I think it makes sense to use values that make a little more sense.

web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -173,16 +173,52 @@ function dpl_event_preprocess_field(array &$variables): void {
...
+/**
+ * Check if the event has free tickets.
+ *
+ * @param array<\Drupal\paragraphs\Entity\Paragraph> $ticket_categories
+ *   An array of ticket categories.
+ *
+ * @return bool
+ *   True if the event has free tickets, false otherwise.
+ */
+function dpl_event_has_free_tickets(array $ticket_categories): bool {
```
kasperg, line 208: Please consider the signature for your function.

Instead of padding an array of ticket categories you might as well pass an actual event entity. This would also make it easier for others to reuse the function in other cases.

kasperg (approved):
:+1: Looks good. I have a few comments for your consideration.

PR #848, opened 2024-03-04
Pull event address from possible branch node. DDFFORM-402

rasben:
Pull event address fallback from possible branch node. DDFFORM-402

Making a custom field formatter, that looks up if a branch node
is associated with the field, and sets the branch address as fallback
if nothing exists on the event already.


kasperg (changes requested):
Please reconsider the logic in this PR-

Bascally I think you reversed the business intention. If an address is shown for the event then it should be shown. Otherwise it should fall back to the address of the related branch.

With the current approach it is not possible to have an event which is both related to a branch and with a custom address.

rasben:
@kasperg You were right - it was the other way around.

I've updated the PR now, along with adding example content.
I've also made sure that the branch is displayed on the event view.

config/sync/field.field.eventinstance.default.field_branch.yml
```diff
@@ -11,7 +11,7 @@ field_name: field_branch
 entity_type: eventinstance
 bundle: default
 label: Branch
-description: ''
+description: 'When connecting a branch, it will also overwrite the event address with the address from the branch.'
```
kasperg, line 14: Please reconsider these descriptions.

I think they are leftovers from the previous implementation where the logic was reversed.

This also applies for the three other descriptions.

phpstan-baseline.neon
```diff
@@ -5,6 +5,11 @@ parameters:
 			count: 1
 			path: web/modules/custom/dpl_dashboard/src/DplDashboardSettings.php
 
+		-
+			message: "#^Method Drupal\\\\dpl_event\\\\Plugin\\\\Field\\\\FieldFormatter\\\\BranchAddressFormatter\\:\\:viewElements\\(\\) return type has no value type specified in iterable type array\\.$#"
+			count: 1
+			path: web/modules/custom/dpl_event/src/Plugin/Field/FieldFormatter/BranchAddressFormatter.php
+
```
kasperg, line 12: Please move this to an `ignoreErrors` entry in `phpstan.neon.dist`.

I this it is perfectly acceptable that you ignore such an error related to Drupals render arrays but ignoring it here will leave the next guy with the same hard-to-fix error when implementing a field formatter.

I think something like this should do the trick:

```
    # Field formatters return render arrays which we cannot provide more detailed typing of.
    - '#Method Drupal\\.*\\Plugin\\Field\\FieldFormatter\\.*\:\:viewElements\(\) return type has no value type specified in iterable type array\.#'
```

web/modules/custom/dpl_event/src/Plugin/Field/FieldFormatter/BranchAddressFormatter.php
```diff
@@ -0,0 +1,76 @@
...
+ * )
+ */
+class BranchAddressFormatter extends AddressDefaultFormatter {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsSummary() {
+    $summary = [];
+    $summary[] = $this->t('Use the address from associated branch as fallback.', [], ['context' => 'DPL admin UX']);
```
kasperg, line 31: Please reconsider the context.

If in doubt use the module name.

web/modules/custom/dpl_event/src/Plugin/Field/FieldFormatter/BranchAddressFormatter.php
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @FieldFormatter(
+ *   id = "dpl_branch_address",
+ *   label = @Translation("DPL: Branch address fallback"),
+ *   field_types = {
+ *     "address"
+ *   }
+ * )
+ */
+class BranchAddressFormatter extends AddressDefaultFormatter {
```
kasperg, line 24: :+1: Well done implementing this as a field formatter.

Personally I like this approach compared to an endless amount of preprocessing.

web/modules/custom/dpl_example_content/content/eventseries/2ce9d9b0-7f10-4a17-9a5e-383d36c1ed93.yml
```diff
@@ -100,34 +101,37 @@ default:
         href: '/user/login'
   event_instances:
     -
-      entity: 21850bf7-ffc9-483c-bcd5-489d420552f9
+      entity: 744dba3f-0fbc-4f58-81ca-7d99a862f4f7
+  field_branch:
+    -
+      entity: 852d781f-82c9-40ac-bc8d-d0cefe07706d
```
kasperg, line 107: :+1: for remembering to create/update example content with new features as we add them.

kasperg (approved):
:+1: LGTM

A few things needs a little cleanup but I do not think the changes need another review before merging.

PR #851, opened 2024-03-04
Make paragraphs cloneable. DDFFORM-398

rasben:
This is to not "re-use" paragraphs when using templates. If this is not done, you might make a template clone, and edit paragraphs and see that it does the changes on both the child, and the parent.

----

This has already been reviewed and merged into the release branch on formidling, but this is critical enough to get it fasttracked into the pilot sites.

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/846


Dresse (approved):
LGTM

PR #853, opened 2024-03-05
Delete deprecated field_event_description field. DDFFORM-298

rasben:
This includes deleting the data using an install hook, and exporting the changes to config.
If we dont delete the field using an install hook first, the `drush cim` might fail, if some site, for whatever reason, already has data in the field.

This is part two of https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/843


web/modules/custom/dpl_event/dpl_event.install
```diff
@@ -0,0 +1,39 @@
...
+      $message .= t('All field_event_description storage of @type has been deleted.', ['@type' => $type], ['context' => 'DPL admin UX'])->render();
+    }
+
+    $config = FieldConfig::loadByName($type, 'default', 'field_event_description');
+
+    if ($config instanceof FieldConfig) {
+      $config->delete();
+      $message .= t('All field_event_description config of @type has been deleted.', ['@type' => $type], ['context' => 'DPL admin UX'])->render();
+    }
+  }
```
kasperg, line 36: I wonder if there a reason we do not do the deletion of the field content as a part of #843.

I have not does this type of work very much.

web/modules/custom/dpl_event/dpl_event.install
```diff
@@ -0,0 +1,39 @@
...
+      $storage->delete();
+
+      $message .= t('All field_event_description storage of @type has been deleted.', ['@type' => $type], ['context' => 'DPL admin UX'])->render();
+    }
+
+    $config = FieldConfig::loadByName($type, 'default', 'field_event_description');
+
+    if ($config instanceof FieldConfig) {
+      $config->delete();
+      $message .= t('All field_event_description config of @type has been deleted.', ['@type' => $type], ['context' => 'DPL admin UX'])->render();
```
kasperg, line 34: Please use a proper context.

It is nice that you use translations for these texts but if you do so a context that matches the string.

kasperg (changes requested):
:+1: One thing I think we forgot in #843 is to [setup field inheritance for the description](https://danskernesdigitalebibliotek.github.io/dpl-docs/dpl-cms/event-field-config/). 

Here I think you also forgot to delete the existing field inheritance for the field that we now delete.

Monosnap Field inheritance entities | DPL CMS | Logged in 2024-03-05 12-29-26

web/modules/custom/dpl_event/dpl_event.install
rasben: I could also have done that, but, I wanted to make sure that the data was migrated first, and then the old field deleted in this PR.
No matter what, we needed to do two deploys regardless, as I also had to export the config.

rasben:
Good point about the inheriterance, i've fixed it in a seperate commit in this PR.

web/modules/custom/dpl_event/dpl_event.install
kasperg: :+1: You are right. It makes sense to keep the old data a bit in case of errors.

kasperg (commented):
:+1: LGTM

I am not going to mark this as approved yet to ensure the we do not merge it prematurely.

kasperg (approved):

PR #855, opened 2024-03-05
Make sure that editors and local administrators actually can edit menus.

rasben:
There is already a sub-permission that limits which menus they can edit - but this permission is apparantly also necessary to show the link of the overview page.


spaceo (approved):
LGTM 👍

PR #858, opened 2024-03-05
Move the "ticket" link to above ticket categories.

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-364

#### Description
This pull request relocates the link field within the admin interface. The previous placement of the link on the left side was counterintuitive, given its close association with ticket categories

#### Screenshot of the result
Skærmbillede 2024-03-05 kl  16 12 19

### DIFF
--- a/config/sync/core.entity_form_display.eventinstance.default.default.yml
+++ b/config/sync/core.entity_form_display.eventinstance.default.default.yml
@@ -98,7 +98,7 @@ content:
     third_party_settings: {  }
   field_event_address:
     type: address_default
-    weight: 7
+    weight: 6
     region: content
     settings: {  }
     third_party_settings: {  }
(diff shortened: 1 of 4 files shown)


LasseStaus (approved):
Nice with the link being closer to the ticket categories. 👍  

I have some thoughts/suggestions you can consider. 

Maybe we could explain more what kind of "link" it is somehow. Extended description or a different titel


I notice that the event state dropdown is located in the sidebar for an event_series, but located in the main area in an event_instance. I'm not sure which one is correct, or if the different placements is intended. 

Paragraphs could be moved to the bottom, so it's not placed in the middle of other event-related categories. 

Will leave the changes up to you, as they are not described in the ticket, but i think it could be done as part of "cleaning-up-on-the-go"

PR #861, opened 2024-03-06
Refactor media placeholder, making it a seperate component. DDFFORM-360

rasben:
Before now, we used various scss mixins to re-use the same placeholder component, that shows up when no image is available. Instead, we'll make a media-container component that can be re-used along with it's own CSS classes.

Please see the design PR, as it will have most of the communication:

https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/542


LasseStaus (approved):
This seems correct to me based on the design changes 👍 🚀

LasseStaus (approved):
LGTM, works when testing. Nice one 👍 

There was a ticket about some images being rendered in a smaller scale than they should, and appearing pixelated. 

I looked but couldn't find the ticket. Perhabs it is already fixed. 

Consider if this has any implications in that regard that you could think of, otherwise 🚀

PR #863, opened 2024-03-07
Ddfform 407 - Remove duplicate markup for date in Event

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-407

#### Description
This pull request removes the duplicate markup for the date in the Event template file. The date is already included and should be retained in `article-header-info.twig`.

#### Screenshot of the result
image


web/themes/custom/novel/templates/components/article-header-info.html.twig
kasperg, line 1: :+1: for the very appropriate rename and keeping it in a separate commit.

kasperg (approved):
:+1: LGTM

PR #864, opened 2024-03-07
Added a branch to each event series in default content

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-65

#### Description
This pull request adds a new branch for each event series within the default content.


kasperg (changes requested):
:+1: I appreciate the intention but as long as branch is not a required field then please leave it unset for one of the events.

Events not having a branch associated is at least an edge case and potentially not something that should be possible.

As long as it is not required then having at least one example that shows behavior without a branch would be good.

kasperg (approved):
:+1: LGTM

PR #865, opened 2024-03-07
Ddfform 393 - Restrict card grid to article selection + tag filtering on automatic

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-393

#### Description
In this pull request, I have updated the card grid to exclusively support `articles`. 

Additionally, the filtering mechanism has been shifted to focus on tags rather than content type.

It's important to note that the `field_filter_content_types` field has been removed. To prevent any issues on sites already using this field, a script has been added to `dpl_filter_paragraphs.install` for its removal. This change is essential to ensure smooth functionality and prevent problems during `drush cim` executions on sites that may have existing data in this field.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-cms/assets/49920322/d04a1d69-9cbc-485a-80b4-432785bb04f5


kasperg (approved):
:+1: LGTM

PR #866, opened 2024-03-08
Automatic unpublishing of events

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-240

#### Description

This PR adds support for automatic unpublishing of events a configurable time period after they have occured.

At the high level this consists of:

1. A new configuration form for handling this is added
2. A new job schedule for unpublishing events is implemented

In the process of implementing this some things were refactored along the way. I recommend reading the commits one by one reading the additional commentary there.

Since Benjamin is on vacation I have also requested a review from @spaceo and @Dresse. If one of you can take a look today I would appreciate it. If this is accepted by @LasseStaus or @kasperbirch1 before then I will merge it at the end of the day.

### DIFF
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -12,6 +12,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
 use Drupal\dpl_event\EventState;
 use Drupal\dpl_event\Workflows\OccurredSchedule;
+use Drupal\dpl_event\Workflows\UnpublishSchedule;
 use Drupal\drupal_typed\DrupalTyped;
 use Drupal\recurring_events\Entity\EventInstance;
 use Drupal\recurring_events\Entity\EventSeries;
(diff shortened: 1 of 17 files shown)


web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -11,112 +11,61 @@ use Drupal\Core\Entity\FieldableEntityInterface;
...
- * Callback to be executed for scheduled jobs.
- */
-function dpl_event_set_occurred_callback(JobSchedule $job): void {
-  $event = \Drupal::entityTypeManager()->getStorage('eventinstance')->load($job->getId());
-  if (!$event || !$event instanceof EventInstance) {
-    return;
+  $updatedEvent = new EventWrapper($entity);
+  if (isset($entity->original) && $entity->original instanceof EventInstance) {
+    // If there are no changes to the date then skip updating schedules.
+    if ($updatedEvent->hasSameDate($entity->original)) {
```
kasperbirch1, line 62:
```suggestion
  $updated_event = new EventWrapper($entity);
  if (isset($entity->original) && $entity->original instanceof EventInstance) {
    // If there are no changes to the date then skip updating schedules.
    if ($updated_event->hasSameDate($entity->original)) {
```

Remember to use `snake_case` in functional code.

kasperbirch1 (approved):
Commits and code are well documented. 👍  

I have tested the functionality and it works. Good job. 😄 

* Maybe it's worth mentioning that it only happens after the cron job has run.

PR #867, opened 2024-03-08
DDFFORM 369

LasseStaus:
#### Link to issue

[DDFFORM-369](https://reload.atlassian.net/browse/DDFFORM-369)

This is connected to [react PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1006)

#### Description

This PR extends the dpl_fbi_work_id field with a property for selecting a material_type. This entails adding a widget for using this display for editors. 

The PR also updates the work_id field for  Recommendation & material grid manual, aswell as updating their preprocess to handle a material_type and pass it to the react apps. 

#### Screenshot of the result

There's no visual change. 

The update function for the storage on work_id was done before changing the fields used on recommendation & materialGridManual paragraphs. 

If the change on the fields makes this update hook redundant, please comment.


web/modules/custom/dpl_fbi/src/Plugin/Field/FieldType/WorkIdItem.php
```diff
@@ -47,6 +47,10 @@ public static function propertyDefinitions(FieldStorageDefinitionInterface $fiel
       ->setLabel(t('Text value'))
       ->setRequired(TRUE);
 
+    $properties['material_type'] = DataDefinition::create('string')
+      ->setLabel(t('The material type'))
+      ->setRequired(TRUE);
```
kasperg, line 52: Consider if `setRequired()` should be `TRUE` here.

In general I do not think it should be required for editors to provide a material type on work fields.
I also wonder how this works in conjuction with us setting `material_type` as not null.

web/modules/custom/dpl_fbi/src/Plugin/Field/FieldType/WorkIdItem.php
```diff
@@ -82,6 +86,12 @@ public static function schema(FieldStorageDefinitionInterface $field_definition)
         'description' => 'Work id.',
         'length' => 255,
       ],
+      'material_type' => [
+        'type' => 'varchar',
+        'length' => 255,
+        'not null' => TRUE,
+        'description' => 'The material type (e.g., bog, e-bog).',
```
kasperg, line 93: I appreciate the examples here.

web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdWithTypeWidget.php
```diff
@@ -0,0 +1,83 @@
...
+    // value of an availability label displayed in React. All strings are
+    // intentionally not translatable. They are already in Danish and wouldn't
+    // be translated anyway, and this keeps the code simpler.
+    $element['material_type'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Material Type'),
+      '#default_value' => $items[$delta]->material_type ?? '',
+      '#options' => [
+              // @codingStandardsIgnoreStart
+              '' => '- None -',
```
kasperg, line 56: This is OK. Note that [Form API also has specific properties to set the empty value](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Select.php/class/Select/10).

web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdWithTypeWidget.php
```diff
@@ -0,0 +1,83 @@
...
+'),
+    ];
+
+    // All of these options are derived from the 'display' material types
+    // available from FBI. The list is not exhaustive. If you change any of
+    // these, remember that they are appended to some URL constructions for
+    // paragraphs. A value, such as 'Musik (online)' must be equivalent to the
+    // value of an availability label displayed in React. All strings are
+    // intentionally not translatable. They are already in Danish and wouldn't
+    // be translated anyway, and this keeps the code simpler.
```
kasperg, line 49: Excellent commenting with references to *why* the code is with references to external systems and specifics that the developer should be aware of.

web/modules/custom/dpl_fbi/dpl_fbi.install
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Adds 'material_type' column to 'dpl_fbi_work_id' field storage.
+ *
+ * This update function alters the schema for custom field storage
+ * to include a new column for material type. It handles both the
+ * base table and the revision table for the field.
+ *
+ * The function is taken from drupal stackexchange:
+ * https://drupal.stackexchange.com/questions/255495/add-column-to-custom-field-schema
```
kasperg, line 15: We are beginning to build a collection of functions that will help us with update functions. As some point it would be nice to collect these in a helper module/package.

See also #843.

web/modules/custom/dpl_fbi/dpl_fbi.install
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @throws \Exception
+ *   Throws exception if the field type cannot be found or if the expected
+ *   field storage is not an instance of FieldStorageDefinitionInterface.
+ */
+function dpl_fbi_update_10201() :void {
```
kasperg, line 21: I wonder why you picked this version name for the update hook.

config/sync/field.field.paragraph.material_grid_manual.field_work_id.yml
kasperg, line 1: This will cause problems when upgrading existing sites. Drupal will not let you delete a field that has data. You have to empty all existing data before.

Also I wonder if this is necessary at all. As far as I remember the cardinality of a field can vary between where it is used.

web/modules/custom/dpl_paragraphs/dpl_paragraphs.module
```diff
@@ -148,16 +147,31 @@ function dpl_paragraphs_preprocess_paragraph__recommendation(array &$variables):
...
+      // or description, otherwise the URL is constructed in the React app.
+      $route_parameters = ['wid' => $work_id];
+      $query_parameters = [];
+
+      if (!empty($material_type)) {
+        $query_parameters['type'] = $material_type;
+      }
+
+      $material_url_object = Url::fromRoute('dpl_react_apps.work', $route_parameters,
+        ['absolute' => TRUE, 'query' => $query_parameters]);
```
kasperg, line 165: :+1: Appreciate the commenting here.

kasperg (changes requested):
:+1: Looks good. My main question here is in regard to the field update which we need to look into.

web/modules/custom/dpl_fbi/dpl_fbi.install
LasseStaus: I thought the standard was to use the drupal version you are on. So this would be drupal 10.2.x. 
I realise didn't check that we are actually on 10.1.7 now, which would make the hook 10101. 

If this isn't correct though let me know! 👀  🚀

kasperg: It is not. You can read more about the naming scheme here: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_N/10

kasperg (approved):

PR #868, opened 2024-03-08
Add error message aria/alt text for the close button

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-482

#### Description

This makes sure that a text is provided for the alt/aria label of the close button and button image.


Dresse (approved):
LGTM

PR #870, opened 2024-03-08
Drop any existing database before restoring from dump

kasperg:
#### Description

A database dump may not drop all existing tables if they were not known at the time of the backup. Consequently they will be left in the database after the restore.

This can cause problems if the tables are created by modules. (Re)installing these modules will fail after restoring as their table already exists.

Use Drush to drop the entire database before restoring to ensure we are in a clean state.


Dresse (approved):
LGTM

PR #872, opened 2024-03-11
DDFLSBP-465 - Add custom template for user creation page

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-465

#### Description

This PR adds a new custom template for the user registration page (`user/registration/create`)
The template is a stripped down version of the normal page template. It removes the header and footer sections of the page.

This PR also has a sister PR: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1019

#### Screenshot of the result

Screenshot 2024-03-11 at 12 10 59

This PR has a sister PR in dpl-react: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1019

### DIFF
--- a/web/themes/custom/novel/templates/layout/page--user--registration--create.html.twig
+++ b/web/themes/custom/novel/templates/layout/page--user--registration--create.html.twig
@@ -1,0 +1,52 @@
+{#
+/**
+ * @file
+ * Theme override to display a user/registration/create page.
+ *
+ * The doctype, html, head and body tags are not in this template. Instead they
+ * can be found in the html.html.twig template in this directory.
+ *
+ * Available variables:
+ *
+ * General utility variables:
+ * - base_path: The base URL path of the Drupal installation. Will usually be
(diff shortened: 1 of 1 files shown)


web/themes/custom/novel/templates/layout/page--user--registration--create.html.twig
```diff
@@ -0,0 +1,52 @@
+{#
+/**
+ * @file
+ * Theme override to display a user/registration/create page.
```
spaceo, line 4: Great that you write that the doctype, html, head and body tags have been left out of this template.
But I think also it makes sense to write why.

Dresse: Actually this is not something I have added. It originates from the default page.html.twig file. 

But maybe I should instead change the description and say that this template is a variant of the page.html.twig template, but without the header and footer?

spaceo (approved):
LGTM 👍

PR #873, opened 2024-03-11
Add link-tag class to user registration links

clausbruun:
#### Description

Adds missing `link-tag` class to user registration info links


Dresse (approved):
LGTM

PR #875, opened 2024-03-11
Make work id fields use new widget

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-369

#### Description

Even though it is currently the only available widget for the field we still need to declare that it is the one to use.


LasseStaus (approved):
Tested and it works. 👍  

Assuming this was not supposed to be a draft.

PR #547, opened 2023-12-11
Lighthouse fixes

rasben:
Lighthouse keeps failing (below 0.9) on PRs.
It appears that a few issues have been building up.

This aims to fix them:

- Setting height and width on existing images (icons)
- After i ran `drush cex`, some missing config translation came out. I guess it was missing from another PR

PR #878, opened 2024-03-11
Ddfform 404 - Add multiple branches related to events

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-404

#### Description
This pull request introduces support for handling multiple branches in relation to events. 

To avoid the complexity of adding a new class with preprocessing, I've implemented the changes directly within the design-system repository. 

You can review the design-system changes here:
[DPL Design System PR #547](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/547/files)

#### Screenshot of the result
image

image


web/themes/custom/novel/templates/fields/field--eventinstance--branch--full.html.twig
```diff
@@ -0,0 +1,3 @@
+{% for item in items %}
+    {{ item.content }}
+{% endfor %}
```
kasperg, line 3: Consider refactoring this.

This is [the default field template](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/e1edb7e35beacb93e78eef0541c8ccebad68d6f1/web/themes/custom/novel/templates/fields/field.html.twig#L1-L1).

I understand that we need to do something because [the default template for branches on event templates overrides this](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/e1edb7e35beacb93e78eef0541c8ccebad68d6f1/web/themes/custom/novel/templates/fields/field--eventinstance--branch.html.twig#L1-L1) but I suggest that we make the current branch field template more specific instead of double override we are doing now.

config/sync/field.storage.eventinstance.field_branch.yml
```diff
@@ -13,7 +13,7 @@ settings:
   target_type: node
 module: core
 locked: false
-cardinality: 1
```
kasperg, line 16: :+1: for remembering to update the cardinality on both series and instances.

config/sync/field.storage.eventseries.field_branch.yml
```diff
@@ -13,7 +13,7 @@ settings:
   target_type: node
 module: core
 locked: false
-cardinality: 1
+cardinality: -1
```
kasperg, line 16: Now that we have updated the cardinality then please remember to update one of the example events with multiple branches - or create a new one.

web/themes/custom/novel/templates/layout/eventinstance--full.html.twig
```diff
@@ -19,6 +19,10 @@
   {
     label: "Address"|trans,
     value: content.event_address.0 ? content.event_address : null
+  },
+  {
+    label: "branch"|trans,
```
kasperg, line 24:
```suggestion
    label: "Branch"|trans,
```

web/themes/custom/novel/templates/layout/eventinstance--full.html.twig
```diff
@@ -19,6 +19,10 @@
...
+    value: content.branch.0 ? content.branch : null
```
kasperg, line 25: Please discuss this addition with the business.

I think I understand your reasoning but there is some duplication between the address field (which we have not figured out how to handle with multiple branches), the event partners and the branches.

As I see it we are not showing the branch because it will duplicate the address.

kasperg (changes requested):
:+1: but I think there are changes here that need a second thought.

I suggest that you discuss this with the business tomorrow.

web/themes/custom/novel/templates/layout/eventinstance--full.html.twig
kasperbirch1: I have removed it for now in:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/878/commits/1a2067e9ea3c576c06cd9d2a5482ab34f4c12c57

PR #879, opened 2024-03-11
Prevent warning when displaying period for past event

kasperg:
#### Description

Do not try to load event instances if no ids for future events can be found.

This will cause an empty value to be passed to `EventInstance::load()` which in turn will issue a PHP warning.

#### Screenshot of the result

Before this change the following error was raised:

Monosnap | DPL CMS 2024-03-11 20-47-19


LasseStaus (approved):
Looks good.

PR #888, opened 2024-03-12
Remove field-ticket-categories.html.twig template.

LasseStaus:
#### Link to issue

[DDFFORM-366](https://reload.atlassian.net/browse/DDFFORM-366)

#### Description

This PR ensures that ticket categories will always be shown on event series. 

Before, we only wanted to show the price, if there were multiple ticket categories with different prices. This is no longer the case. 

#### Screenshot of the result

image


kasperg (approved):
:+1: LGTM

PR #892, opened 2024-03-14
Add support for TwentyThree video platform

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-427

#### Description

Some public libraries use the TwentyThree video platform. This platform supports oEmbed which we currently use for using remote video.

However it is not supported out of the box - partly because it is not listed on the official list of oEmbed providers on https://oembed.com/providers.json. In practice it is not possible to get the platform listed here either. It is a whitelabel software-as-a-service platform which can be exposed using custom domains. This makes it impossible to add a finite list of url schemes and endpoints supported by TwentyThree.

We could in theory use the oEmvbed Providers Drupal module to allow libraries to define their TwentyThree platform with its custom domain as an oEmbed Provider.
https://www.drupal.org/project/oembed_providers

This has two drawbacks:

1. It is somewhat technical
2. It does not integrate with the Core Media module in adding the new providers to the list of supported oEmbed providers for remote video. Instead one has to create a new media type for the provider.

Instead we introduce a new module, media_twentythree which integrates with TwentyThree which should be easy for editors to handle:

1. It adds TwentyThree to the list of supported providers for remote video
2. Through oEmbed discovery it allows editors to add videos from whatever TwentyThree site using whatever url.

#### Screenshot of the result

Medie element fra TwentyThree
billede

Indsættelse af medie fra TwentyThree
billede

Redigering af side med medie fra TwentyThree
billede

Visning af side med medie fra TwentyThree
billede

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -82,6 +82,7 @@ module:
   media: 0
   media_library: 0
   media_library_edit: 0
+  media_twentythree: 0
   menu_link_content: 0
   menu_ui: 0
   metatag: 0
--- a/web/modules/custom/media_twentythree/README.md
+++ b/web/modules/custom/media_twentythree/README.md
@@ -1,0 +1,30 @@
+# Media: TwentyThree
+
+[TwentyThree](https://www.twentythree.com/) is a video platform. [It supports oEmbed](https://www.twentythree.com/help/manage-oembed)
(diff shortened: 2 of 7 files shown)


kasperbirch1 (approved):
LGTM 👍  Great that you have written a good and explanatory `readme.md`

PR #900, opened 2024-03-15
Add Opening Hours Taxonomy and Color Field

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-408

#### Description
This pull request introduces a taxonomy for various types of Opening Hours, such as 'Åbent' and 'Telefontid.' 
These terms are intended for future use in the API from `dpl_opening_hours` module.

Additionally, I have added a color field to the 'Opening Hours Categories' terms. This enhancement allows users to select a color that will be visible in the event display within the full calendar view.

#### Screenshot of the result
image

image

image

Added color palette
image

### DIFF
--- a/config/sync/language/da/core.date_format.short.yml
+++ b/config/sync/language/da/core.date_format.short.yml
@@ -1,1 +1,2 @@
 label: 'Standard kort dato'
+pattern: 'm/d/Y - H:i'
--- a/web/modules/custom/dpl_example_content/content/media/618e176a-a45a-4c36-a197-664230aa0a34.yml
+++ b/web/modules/custom/dpl_example_content/content/media/618e176a-a45a-4c36-a197-664230aa0a34.yml
@@ -33,6 +33,10 @@ default:
       attributes:
         rel: canonical
         href: '/user/login'
+  path:
+    -
+      alias: ''
+      langcode: da
   field_byline:
     -
       value: 'Af Ahmad Odeh / Unsplash'
--- a/composer.json
+++ b/composer.json
@@ -82,6 +82,7 @@
         "drupal/address": "^1.0",
         "drupal/admin_toolbar": "^3.4",
         "drupal/administerusersbyrole": "^3.4",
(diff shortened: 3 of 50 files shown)


web/modules/custom/dpl_example_content/content/media/618e176a-a45a-4c36-a197-664230aa0a34.yml
```diff
@@ -33,6 +33,10 @@ default:
       attributes:
         rel: canonical
         href: '/user/login'
+  path:
+    -
+      alias: ''
+      langcode: da
```
LasseStaus, line 39: It seems this config is added to most example contents we have as part of this commit. 
Just highlighting since i wasn't sure if this was intentional. I guess its' just something drupal picked up on as part of updating the content?

LasseStaus (approved):
LGTM, nice one! 

I think it would be nice to add an opt-in option to use the colors, rather than forcing them to use them. 

As i see it, the colors are something that could help the editors distinguish _if_ they want to. If this instead was an opt-in, then if no color was selected, we would just use the default fullCalender color.

config/sync/language/da/core.date_format.short.yml
```diff
@@ -1 +1,2 @@
 label: 'Standard kort dato'
+pattern: 'm/d/Y - H:i'
```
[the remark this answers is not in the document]
kasperbirch1: It's coming from the `dce drush cex`

It is not relevant for this pull request, but I think it's a good idea to keep it. Otherwise, it will just be the same for the next person that exports the config files

web/modules/custom/dpl_example_content/content/media/618e176a-a45a-4c36-a197-664230aa0a34.yml
kasperbirch1: It's coming from the `dce drush default-content:export-module dpl_example_content` and we don't want to cherry-pick what it does

PR #901, opened 2024-03-15
Remove unused group-Modal-Return-Library translation text

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-463

#### Description
Removes unused group-Modal-Return-Library translation text

#### Screenshot of the result
-

[Sibling PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1024)


Dresse (approved):
LGTM

I can see that the Drupal config check is failing. I don't know why that is, but I know how you can fix it.
It happens because it expects the Drupal module names to be in alphabetical order, but for some reason it is not. I think if you move the module and place it in the correct spot, it will fix the error 😄

Adamik10:
@Dresse it was failing to begin with without any changes of mine. But I do have a separate PR to fix it 🎉 
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/903

PR #903, opened 2024-03-15
Export latest drupal config

Adamik10:
Export latest drupal config changes to satiate our beloved drupal config CI test.

Note: This might wanna get merged into develop?


config/sync/language/da/block.block.claro_breadcrumbs.yml
```diff
@@ -1,2 +1,2 @@
 settings:
-  label: Brødkrummer
+  label: Breadcrumbs
```
kasperg, line 2: This replaces a Danish text "Brødkrummer" with an English text "Breadcrumbs" in a Danish translation file. That does not seem intentional - nor correct. The same goes for the rest of this PR.

This needs to be investigated further. Perhaps it is related to #849?

kasperg (changes requested):
Please reconsider the changes in this PR. I do not think it should merged as is.

Adamik10:
Follow further in #910

PR #904, opened 2024-03-17
DDFLSBP-503 - Created a new Cypress test suite that checks if static exists

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-503

#### Description

This PR sets up a new Cypress test suite for checking if static pages exists and that they contain an header title.
I have created 1 new test for each static page. The reason for this is, that it would then be easier to expand each test later on if needed.


Adamik10 (approved):
LGTM 💯

clausbruun (approved):
LGTM 🎉

PR #905, opened 2024-03-18
Feature/ddflsbp 474/fix cypress tests on user registration pages

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-474

#### Description

This PR moves the header specific javascript assets into its own librarie, which is then attached to the header template. 
The reason for this is, that the header javascripts where failing on pages where the header region is not included, this in turn broke our Cypress tests.

This PR also updates the "Adgangsplatformen" Cypress tests to check both the registration and cancellation of user registration flow. It also adds a new test for checking if the header and footer regions are removed from the user registration page.

This PR also has a sister PR thats adds new data-cy properties to user registration buttons. That PR needs to be merged before this one.
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1030


cypress/e2e/adgangsplatformen.cy.ts
```diff
@@ -59,8 +59,11 @@ describe("Adgangsplatformen", () => {
       "The website encountered an unexpected error. Please try again later."
     );
   });
-  // TODO: The test should be re-added when the new user registration is implemented.
-  it.skip("can register a new user and expose the right tokens for the react apps", () => {
+
+  // When a user comes back from authentication with MitID, the user should
```
Adamik10, line 63: Great that you document things!!! 👏

Adamik10 (approved):
Good stuff! :) Let's hope that combined with the other PR it all works!! 🤞 ✨

Dresse:
> Good stuff! :) Let's hope that combined with the other PR it all works!! 🤞 ✨

After merging the sister PR, the Cypress CI tests are now all green 🎉 

I'll merge this PR

PR #908, opened 2024-03-18
Event facets

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-65

#### Description

This PR adds support for filtering events based on branches. This introduces the use of Facets to support filtering.

Please go through the individual commits for additional commentary.

#### Screenshot of the result

Monosnap Events | DPL CMS 2024-03-18 14-04-20

This PR depends on https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/559.

### DIFF
--- a/web/themes/custom/novel/templates/views/views-view--events.html.twig
+++ b/web/themes/custom/novel/templates/views/views-view--events.html.twig
@@ -1,1 +1,5 @@
 {% extends '@novel/views/views-view--content-list-page.html.twig' %}
+
+{% block facets %}
+  {{ drupal_block("facet_block:event_categories") }}
+{% endblock %}
--- a/web/themes/custom/novel/templates/views/views-view--events.html.twig
+++ b/web/themes/custom/novel/templates/views/views-view--events.html.twig
@@ -1,7 +1,10 @@
 {% extends '@novel/views/views-view--content-list-page.html.twig' %}
 
 {% block facets %}
-  {{ drupal_block("facet_block:event_categories") }}
-
-  {{ drupal_block("facet_block:branch") }}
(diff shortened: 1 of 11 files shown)


kasperbirch1 (approved):
Well done! That you got it solved! I have tested it and it works, but there is a minor CSS error in terms of spacing between dropdowns on my local setup / Brave browser.

image

web/themes/custom/novel/templates/views/views-view--events.html.twig
```diff
@@ -1,7 +1,13 @@
...
-  {{ drupal_block("facet_block:branch") }}
+  <div class="content-list-page__filter">
+    {{ drupal_block("facet_block:event_categories") }}
+  </div>
+  <div class="content-list-page__filter">
+    {{ drupal_block("facet_block:branch") }}
+  </div>
+  <div class="content-list-page__filter">
+    {{ drupal_block("facet_block:date") }}
+  </div>
```
LasseStaus, line 12: Is this ment to be here? I see no mention of date filtering and as far as i can see this is not a facet block on `/admin/config/search/facets`

It's causing a lot of warnings on `/events` when logged in.

LasseStaus (changes requested):
See comment about the date facet block. 🚀

web/themes/custom/novel/templates/views/views-view--events.html.twig
kasperg: You are quite right. I messed up the juggling between the date facet and the category and branch.

LasseStaus (approved):
Seems good to me! 🎈

PR #910, opened 2024-03-19
Export latest drupal config

Adamik10:
Export latest drupal config changes to satiate our beloved drupal config CI test.


Dresse (approved):
LGTM

.gitignore
```diff
@@ -59,3 +59,6 @@ docker-compose.override.yml
 
 # Module upload folder
 /web/modules/local
+
+# Danish translations folder
+/config/sync/language/da
```
kasperg, line 64: Consider deleting the contents of this folder if we do not care about the contents anymore because [translations now live in translations on GitHub Pages](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/gh-pages/translations/da.config.po#L15-L17).

PR #912, opened 2024-03-20
Ddflsbp 357 remove redirect looping on blocked users

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-357

#### Description

An blocked user went into an redirect loop until the user received an error message. The reason for this is that the destination query on the adgangsformen login url is set to /user/me/dashboard. This works well for unblocked users as they are allowed access to their own dashboard. But anonymous users also gets redirected to /user/me/dashboard, but as this page is protected behind the patron redirect module, the user gets send back to adgangsplatforment, which remembers the user and once again sending the user to /user/me/dashboard page. This loop continues until it breaks with a error message.

This PR seeks to prevent this behaviour by adding:
- A new patch for openid_connect which makes it possible to set a new redirect destination URL in session. This is done by moving the redirect destination logic to after the pre_authorization hook has been run.
- The PR also expands the blocked user check so it now adds a new redirect destination URL for blocked users. I have set the redirect blocked users URL to be whatever is set to be the frontpage.


web/modules/custom/dpl_login/src/Controller/DplLoginController.php
```diff
@@ -110,14 +110,42 @@ public function logout(): TrustedRedirectResponse|RedirectResponse {
...
+  /**
+   * Check if a user token has been set and decide redirection.
+   *
+   * Check if a user token has been set and either allow redirecting
+   * to the original path or redirect to frontpage.
+   *
+   * @param \Symfony\Component\HttpFoundation\Request $request
+   *   Symfony request object.
+   *
+   * @return \Drupal\Core\Routing\TrustedRedirectResponse|\Symfony\Component\HttpFoundation\RedirectResponse
```
clausbruun, line 136: Vil den ikke altid returnere `TrustedRedirectResponse` ?
```suggestion
   * @return \Drupal\Core\Routing\TrustedRedirectResponse
```

clausbruun (approved):
LGTM, blot en enkelt kommentar.

web/modules/custom/dpl_login/src/Controller/DplLoginController.php
Dresse: Jo du har ret. Har fjernet den nu og opdateret beskrivelsen.

PR #917, opened 2024-03-21
WIP: Ddfform 465 billede kreditering viser ikke pa event instanser

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-465

#### Description
This pull request should be put on hold until we can collectively decide on a strategy for managing different image formats. 

The problem with the hero image not showing credits is that the byline field is not showing. We need to consider that applying hero_wide across different sections, such as list teasers and the map grid, leads to design inconsistencies, and we should also review its impact in other areas.

My idea is to add a new image format `hero_wide_with_crediting` to handle how we can get credit on some images and not others. This needs to be decided together if this is the right way to handle it.

#### Screenshot of the result
image

image

image

### DIFF
--- a/config/sync/core.entity_view_display.media.image.hero_wide.yml
+++ b/config/sync/core.entity_view_display.media.image.hero_wide.yml
@@ -4,6 +4,7 @@ status: true
 dependencies:
   config:
     - core.entity_view_mode.media.hero_wide
+    - field.field.media.image.field_byline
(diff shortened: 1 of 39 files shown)


rasben:
I've closed this, as there was a ton of conflicts, and it was easier to add to my own PR:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/861

PR #918, opened 2024-03-22
Event date facet (DDFFORM-67)

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-67

#### Description

#### Screenshot of the result

This PR depends on the related changes in the design system: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/562


LasseStaus (changes requested):
I think there are some issues with this.

Since most of this logic comes from the design system, i will add most of my thoughts in the design system.

So some things i have tested here in the cms, will be highlighted in the design PR. 


One thing for the CMS tho, is that if i select a date range, and reload the page, the CSS is not rendered for the page when testing.

kasperg:
Closed to make room for more pull requests.

PR #920, opened 2024-03-22
Make sure that contrib modules translations are being scanned as well

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-514

#### Description

Labels and descriptions are exposed to library admins from contrib modules so the need to be translated as well.

### DIFF
--- a/task/Taskfile.translations.yml
+++ b/task/Taskfile.translations.yml
@@ -14,6 +14,8 @@ tasks:
     cmds:
       # Make sure that the potion module is active
       - task dev:cli -- drush en -y potion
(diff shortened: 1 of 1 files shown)


Dresse (approved):
LGTM - However, won't we get a ton of new strings to translate? Many of the strings might not even be things DDF editors will ever have access to?

spaceo:
> LGTM - However, won't we get a ton of new strings to translate? Many of the strings might not even be things DDF editors will ever have access to?

Yes... and yes :)

PR #890, opened 2024-03-13
Opening Hours

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-410

#### Description

This adds

- A hook_schema() for creating a database table for storing opening hours instances.
- A value object for storing instances with a proper types as we are going to get.
- A repository for managing persistance and mapping between database rows and value objects.
- An updated OpenAPI specification
- Generated API code based on the specification
- REST resources for handling API requests to manage opening hours instances based on the repository and the specification
- A few updates to the opening hours category taxonomy

PR #921, opened 2024-03-25
Initial model for managing opening hours instances and persistance

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-410

#### Description

This adds

- A hook_schema() for creating a database table for storing opening hours instances.
- A value object for storing instances with a proper types as we are going to get.
- A repository for managing persistance and mapping between database rows and value objects.

If mapping fails due to data issues we prefer to handle errors within the repository, log them and return whatever values we have to the caller.

Another option for handling opening hours would be to create an entity for storing each instance, but after some consideration we opted for a simple table based approach.

So there is no actual UI or the like for this yet so there is nothing to test or the like. This is mostly a matter of talking data structures, code patterns, programmatic APIs and the like.


kasperg:
This has now been merged in #890

PR #925, opened 2024-03-26
Add selective scanning of contrib modules

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-514

#### Description

Some modules breaks the translation scanning from Potion. So we needed to ignore the failing modules.

In the end it was done with a white list of modules because newly added modules might break the Github workflow and it might not be obvious for everybody what caused the workflow to break.


dev-scripts/translate-source/scan-translations.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+# Get the directory of the script
+THIS_DIR=$(dirname "$0")
+
+
+# First import translations from custom modules.
+drush potion:generate $LANGUAGE modules/custom/ $PO_DIR --recursive --default-write-mode=override
+
+# Define an array of modules to exclude
```
Dresse, line 29:
```suggestion
# Define an array of modules to exclude
```
# Define an array of modules to include

dev-scripts/translate-source/scan-translations.sh
```diff
@@ -0,0 +1,47 @@
...
+done < ${THIS_DIR}/scanned_modules.txt
+
+for module in "${include_modules[@]}"; do
+    # Prepend the base path to the directory
+    dir="web/modules/contrib/$module"
+
+    if [ -d "$dir" ]; then
+        echo "Processing $module:"
+        dir=$(echo $dir | sed 's/^web\///')
+        drush potion:generate da $dir profiles/dpl_cms/translations --recursive --default-write-mode=merge
```
Dresse, line 42: Should the default-write-mode be "merge" or "override"? I can see that it was set to "override" when all contrib modules was being scanned.

Dresse (changes requested):
Looks good overall. Have 1 suggested change and a question regarding --default-write-mode

dev-scripts/translate-source/scan-translations.sh
spaceo: The [first operation](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/925/files#diff-55569087b6a5448bba7209331df93ef27c9f2ac4e2c5388ff1018fa8ada49354R27), that covers the custom modules, **overrides**, which basically means starting on a clean slate.
The following two opererations scans contrib modules and themes and they **merge**, which means they add to the translations from the first operation.
So it is setup as intended.

Dresse: Ah okay, it makes sense. 👍 

However shouldn't this also be the way it was implemented before? From your explanation it seems that in the previous implementation we would always override the the scanning of the contrib modules when starting scanning the custom modules? Maybe I am missing something 😄

Dresse (approved):
LGTM

dev-scripts/translate-source/scan-translations.sh
spaceo: The former change that added the scanning of contrib modules was accidentally set to **override** which was a mistake.
The idea is that the first operation **overrides** and the rest **merges**.
Does it make sense?

Dresse: Yes that makes sense 👍 Just wasn't sure if the former was on purpose or a mistake 😄

PR #928, opened 2024-03-27
Display term hiarchy on /articles. DDFFORM-512

rasben:


kasperg (approved):

kasperg (commented):
This shows the hierarchy on `/articles` but what about the category dropdown on `/events`?

Monosnap Events | DPL CMS 2024-04-02 09-14-49

If we are working actively with term hierarchy then perhaps consider making a child term in one of our categories in the example content. That makes it easier for us to work with.

rasben:
I've added now, what we discussed in regards to /events

config/sync/facets.facet.event_categories.yml
```diff
@@ -19,13 +19,13 @@ field_identifier: event_categories
...
-use_hierarchy: false
+use_hierarchy: true
 keep_hierarchy_parents_active: false
 hierarchy:
   type: taxonomy
   config: {  }
-expand_hierarchy: false
-enable_parent_when_child_gets_disabled: true
+expand_hierarchy: true
+enable_parent_when_child_gets_disabled: false
```
kasperg, line 28: This config name is somewhat a mindfuck...

kasperg (approved):
:+1: LGTM

config/sync/facets.facet.event_categories.yml
rasben: indeed

PR #935, opened 2024-04-02
Footer update

rasben:
- Override faulty button styling from reset.scss. DDFFORM-487
- Expand footer settings, to include secondary footer. DDFFORM-406

------

Test it by going here: https://varnish.pr-935.dpl-cms.dplplat01.dpl.reload.dk/admin/structure/footer

And yeah, the form is getting a bit bloated. I think we should take this discussion with DDF/in the demo.

### DIFF
--- a/web/modules/custom/dpl_admin/assets/dpl_admin.css
+++ b/web/modules/custom/dpl_admin/assets/dpl_admin.css
@@ -2,6 +2,11 @@
  * This file is loaded into admin pages.
  */
 
+/* Overriding wrong styling from reset.scss */
+.button:focus {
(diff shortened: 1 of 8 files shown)


LasseStaus (approved):
LGTM, works when testing 👍  

I don't necessarily think that the form is too bloated, but we can discuss it. 

I would remove the default "blank" column. Even tho it won't be displayed if blank i don't think it is generally intuitive. And it would also "unbloat" the form a bit. I think there is a setting for it. 

The button for saving is translated into "indsend". Maybe this is already the case, but just check that this is not unintended. I Assume it should say "Gem" in danish, like the rest of the save buttons. 

🚀

PR #941, opened 2024-04-04
Fix remaining search/material local/global refactoring

spaceo:
#### Description

Not all keys were refactored. This should fix the problem.


Dresse (approved):
LGTM

PR #942, opened 2024-04-04
DDFLSBP-470-add-login-statistics

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-470

#### Description

I suggest not going through the individuel commits, as I have been doing some debugging in GitHub, as the cypress tests were failing. Instead have a look at the complete PR files instead.

This PR makes sure to save a new variable in Session that describes which kind of login method a user has been logged in with when logging in through adgangsplatformen. 

We also add a new a new tracking event that sends the login method info on each page request. 

Lastly it expands the mapp-tracking.cy.ts cypress test. We make sure that the user is logged out of Drupal after setting the initial settings in dpl_mapp module. 

It is difficult to test if the logging is happening as expected, so we want to get this PR out and get feedback from the client in regards to wether of not this change meets the expected requirements.


Adamik10 (approved):
Good job looking into this. Let's hope 🤞 this works.

PR #945, opened 2024-04-05
Decrease minimal react app container height to be same as screen height

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-532

#### Description
The containers for React apps' min-height was set to double the size of screen height, which creates a lot of unnecessary whitespace on short pages. 

#### Screenshot of the result
-

-


spaceo (approved):
LGTM 👍

PR #947, opened 2024-04-05
Ddflsbp 513 oversatte tekster lober ikke tilbage til sitet

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-513

#### Description
For various, undocumented, reasons our translation workflow does not like multiline strings separated by: `\r\n`. And it also does not accept `<br>` tags - it should be `<br />`.

So this PR make sure that:
* Existing custom t strings are using `<br />`tags instead of `<br>`
* Existing config exported descriptions are using `<br />` tags instead of `\r\n`
* Future config exported descriptions will make sure that `\r\n` are converted to `<br />` tags


Dresse (approved):
LGTM

PR #950, opened 2024-04-08
Update event specification

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-476

#### Description

This PR updates our API specification for exposing events based on feedback. It consists of two changes:

1. API versioning
2. Moving capacity to a per event count instead of a per category count

Note that the API versioning currently only applies to the event resource as it will be consumed externally. Other APIs should follow shortly.


spaceo (approved):
Documentation makes sense since we all agreed (as far as I remember) in versioning the services via the URI.
Nice with e addition of what we considering breaking and what not.

LGTM 👍

PR #955, opened 2024-04-08
Make sure header doesnt break out zoom on mobile. DDFFORM-517

rasben:
We cant really reflect this in the design system, as we have no story that combines the header, page etc.

This relies on https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/574 being merged first.

Screenshot 2024-04-08 at 15 16 51

### DIFF
--- a/web/themes/custom/novel/templates/layout/page.html.twig
+++ b/web/themes/custom/novel/templates/layout/page.html.twig
@@ -43,7 +43,8 @@
  */
 #}
 
-<div>
+{# Overflow hidden is necessary, to avoid the header breaking out on mobile. #}
+<div class="overflow-hidden">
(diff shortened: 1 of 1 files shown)


web/themes/custom/novel/templates/layout/page.html.twig
```diff
@@ -43,7 +43,8 @@
  */
 #}
 
-<div>
+{# Overflow hidden is necessary, to avoid the header breaking out on mobile. #}
+<div class="overflow-hidden">
```
kasperbirch1, line 47: Would we like an extra div here? Shouldn't it be solved in the design system?

kasperbirch1 (approved):

PR #958, opened 2024-04-08
Displaying the DPL CMS version in the secondary admin toolbar. DDFFORM-553

rasben:
**Displaying the DPL CMS version in the secondary admin toolbar. DDFFORM-553**

Which includes these two also:

**Create a version service, for pulling CMS version number. DDFFORM-553**

This is necessary, as we want to be able to display this data in
other places, rather than just the module requirements.
By moving it into a service, we can use it all over the site.

**Stop Gin theme hiding secondary toolbar links when there is space.**

Screenshot 2024-04-08 at 15 58 57

https://reload.atlassian.net/browse/DDFFORM-553


web/modules/custom/dpl_admin/dpl_admin.install
```diff
@@ -5,64 +5,30 @@
...
 
-    $requirements['dpl_admin_version'] = [
-      'title' => t('CMS release version'),
-      'value' => t('<code>@version</code>', ['@version' => $version]),
-      'severity' => REQUIREMENT_INFO,
-    ];
-  }
+  $requirements['dpl_admin_version'] = [
+    'title' => t('CMS release version'),
+    'value' => t('<code>@version</code>', ['@version' => $version]),
```
kasperg, line 29: Consider if these texts should be translateable and have a context.

I know it is outside the scope of this PR - so for now it is just a consideration.

kasperg (approved):
:+1: Nice! There are a few things that needs to be changed before this can be merged (context for translations) but you can fix this on your own.

PR #966, opened 2024-04-09
DDFLSBP-523/webform paragraph

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-523

#### Description

This PR is meant to show the minimum implementation of a contact webform through a paragraph type. More functionality and configuration will be added as part of other PR's.

This PR does the following:
* Adds the webform module through composer. 
* Enables webform and its webform_ui submodule.
* Adds default webform configuration.
* Adds a new paragraph type called "Webform" which is used for showing a webform.

#### Screenshot of the result
The below screenshot is showing hot an un-styled webform paragraph looks when embedded on a page.

Screenshot 2024-04-09 at 13 01 57

As part of this PR I have also researched:
* Is it possible to add a paragraph that is referencing a webform entity using an entity-reference field - Check
* Is it possible to add the paragraph to existing content (page, articles etc.) - Check
* Is the webform shown in the frontend on content where a webform paragraph has been added - Check
* Is it possible to submit the webform - Check
* Does it cause any troubles when having more thant 1 instance of the same webform - Check (it does not)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -135,6 +135,7 @@
         "drupal/varnish_purge": "^2.2",
         "drupal/view_unpublished": "^1.1",
         "drupal/views_infinite_scroll": "^2.0",
+        "drupal/webform": "^6.2",
         "drush/drush": "^12",
         "mnsami/composer-custom-directory-installer": "^2.0",
         "myclabs/php-enum": "^1.8",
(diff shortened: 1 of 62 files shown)


spaceo (approved):
Somehow I have the feeling that it could be a good idea to make a separate release for this. But I don't have any heavy weigthing reasons so maybe just ignore this rambling.
LGTM 👍

Dresse:
> Somehow I have the feeling that it could be a good idea to make a separate release for this. But I don't have any heavy weigthing reasons so maybe just ignore this rambling. LGTM 👍

Are you thinking a separate release from release/brahma-14? Or a release branch that contain all webform development, which are then merged into release/brahma-14 when ready?

PR #969, opened 2024-04-09
Ddfform 553 vis cms version som statisk info i redaktormenuen

rasben:
This has already been reviewed, but was mistakenly merged into develop:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/958


kasperg (approved):
:+1: LGTM

PR #985, opened 2024-04-11
Replace /articles with search API view, along with facets. DDFFORM-472

rasben:
**Replace /articles with search API view, along with facets. DDFFORM-472**

Up until now, we've used exposed filters on /articles, and a search API view on /events, with facets.
This replaces the /articles view with the same kind of facet-view that /events uses.

https://reload.atlassian.net/browse/DDFFORM-472

----

Also a little extra thing: Making sure that /events only shows published events.

Test it by looking at  https://varnish.pr-985.dpl-cms.dplplat01.dpl.reload.dk/articles :)


Dresse (approved):
LGTM - Only skimmed through the facet/search configurations.

PR #986, opened 2024-04-11
Teaser tweaks - list and slider teasers.

rasben:
**Introduce list-teaser viewmode for media, without caption. DDFFORM-465**

This is to avoid that the caption shows up, which looks weird in teaser lists - and also breaks the design.

------

**Dont display created date on page teaser. DDFFORM-524**

After preference from DDF.

https://reload.atlassian.net/browse/DDFFORM-524

https://reload.atlassian.net/browse/DDFFORM-465

### DIFF
--- a/config/sync/core.entity_view_display.eventinstance.default.list_teaser.yml
+++ b/config/sync/core.entity_view_display.eventinstance.default.list_teaser.yml
@@ -77,7 +77,7 @@ content:
     weight: 11
     region: content
   event_description:
-    type: text_default
+    type: basic_string
     label: hidden
     settings: {  }
     third_party_settings: {  }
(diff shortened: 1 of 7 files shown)


config/sync/core.entity_view_display.eventinstance.default.list_teaser.yml
```diff
@@ -77,7 +77,7 @@ content:
     weight: 11
     region: content
   event_description:
-    type: text_default
+    type: basic_string
```
Dresse, line 80: Is this change intentional? Didn't see it mentioned in the PR description.

Dresse (approved):
LGTM, but I have a small comment/question.

config/sync/core.entity_view_display.eventinstance.default.list_teaser.yml
rasben: Yes and no - It's not something i did actively, but, it has come from saving the view display, and apparantly this config was missing :) I dont see it as a problem, but good that you're awake!

PR #987, opened 2024-04-11
Delete unused Solr DB server.

rasben:
We have two Solr servers:
"DB" and "DB search".

I believe this happened because of some previous work, where several developers had to use Search API, but then never was cleaned up.
The "DB" server does nothing, and is removed now.


Dresse (approved):
LGTM

PR #992, opened 2024-04-12
Enable select2_multicheck, and replace all select fields.

rasben:
This has a much better UX for editors, can be used both for simple selects and for multiselects.
See
https://www.drupal.org/project/select2_multicheck
https://www.drupal.org/project/select2

Select2 needed me to make some repository updates to composer.json, to include the select2 library.

Screenshot 2024-04-12 at 11 38 38
Screenshot 2024-04-12 at 11 38 42

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,10 @@
             "url": "https://packages.drupal.org/8"
         },
         {
+            "type": "composer",
+            "url": "https://asset-packagist.org"
+        },
+        {
             "type": "package",
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-react",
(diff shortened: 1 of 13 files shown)


kasperg (approved):
:+1: LGTM

I wonder if you have any thoughts regarding why you went with Select2 instead of alternatives like Chosen or JQ Multiselect, which we also considered?

rasben:
>I wonder if you have any thoughts regarding why you went with Select2 instead of alternatives like Chosen or JQ Multiselect, which we also considered?

I figured an ADR is a bit overkill for this, but, yes, I did consider the two others:

- jquery_ui_multiselect worked fine on the other project I worked on, but Select2 is much 'prettier', and without a dependency on jQuery
- I cant for the live of me remember why i didnt like `chosen` - it might've been that it also relies on jQuery, though, and that I think Select2 is still a nicer experience

PR #993, opened 2024-04-12
DDFLSBP 509/catch correct exception when trying to get patron

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-509

#### Description

As part of testing the logout functionality, we encountered an PHP error for unregistered users. It turned out to be because we were not catching the correct exception. 
This PR adds ApiException instead of InvalidArgumentException as well as changing the exception logging message to match the other implementation of the function.


spaceo (approved):
LGTM 👍

PR #994, opened 2024-04-12
Replace breadcrumb update complexity with cache tags

kasperg:


rasben:
This does not work.
Screenshot 2024-04-13 at 10 59 23

rasben (approved):

PR #995, opened 2024-04-13
DDFLSBP-534/users with unregistered user tokens should redirect to user registration page

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-534

#### Description

This PR adds logic to handle redirection when logging in without being registered. We check if an unregistered-user or normal user token is set. If the normal user token is set, we redirect the user to where the user came from. If the unregistered-user token is set instead, we redirect to the user/registration/create page so the user can finish the registration process.


web/modules/custom/dpl_login/src/Controller/DplLoginController.php
```diff
@@ -149,14 +155,26 @@ public function loginHandler(Request $request): TrustedRedirectResponse {
...
+      return $tokenInfo;
     }
     if ($access_token = $this->userTokensProvider->getAccessToken()) {
-      return $access_token;
+      $tokenInfo = [
+        'token' => $access_token,
+        'type' => 'user',
+      ];
+      return $tokenInfo;
     }
```
spaceo, line 178: I think we can do a more pretty solution here.
As far as I see it we could do one of two things:
* Return the whole acces token object in this function and set the return type to the interface
* Implement a property on the token classes that says which type it is

I think I am most fan of the first solution. Then we can assert the type of the token by looking at which class the token is.

spaceo (changes requested):
Looks good. There is one point that can be done in a better way, I think.

spaceo:
This should be merged first if you agree on the changes:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1025

spaceo (commented):
This should be merged first if you agree on the changes:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1025

spaceo (approved):
LGTM 👍

PR #997, opened 2024-04-15
Add `reserve-with-material-type-text` translation

kasperbirch1:
#### Description

Add `reserve-with-material-type-text` translation


spaceo (approved):
LGTM 👍

PR #1004, opened 2024-04-16
Remove mnsami/composer-custom-directory-installer

kasperg:
#### Description

This was originally added when adding the Swagger UI module.

We have now added the oomphinc/composer-installers-extender package which adds broader support for custom package types. Consequently we no longer need mnsami/composer-custom-directory-installer to place the Swagger UI library in web/libraries.

Remove it to reduce exposure and complexity in our dependency chain.

#### Screenshot of the result

This can be tested by checking that Swagger UI is shown.


kasperg:
This did not work.

PR #1005, opened 2024-04-17
Webform - DDFLSBP-524 and DDFLSBP-527

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-524
https://reload.atlassian.net/browse/DDFLSBP-527

#### Description

**New custom module "dpl_webform_configuration":**
This PR introduces a new custom module "dpl_webform_configuration" which is used for handling custom logic and configuration for webforms. This PR adds a new hook that sets the "results_disabled" to true on all new webforms. This is done to make sure that the editor don't create forms that save submissions to the database. 

**New contrib module "[honeypot](https://www.drupal.org/project/honeypot)":**
The PR also introduces a new contrib module called "honeypot". This module is added to prevent spam bots from submitting form submissions on all webforms on the site. It does so by adding a hidden field on all webforms which the spam bots will try to fill out. If the field is filled out, it means that a spam bot is making the form submission, and it will not be handled.

**New webform select element "Category" and new vocabulary "Webform email categories":**
The PR adds a new taxonomy vocabulary type called "Webform email categories". Each new term consists of a name and an email. A new form select element called "Category" is added to the contact webform. This new select element get its input from the terms in the "Webform email categories" vocabulary. 
The idea is that editors can add new categories to the list, which can then be chosen when filling out the contact webform. We then want to be able to send a notification email to a specific email address based on what category is chosen. The logic for this has not yet been implemented.

I used some time to fiddle around with trying to add webform handlers to all new created webforms. I managed to add handlers, but was unable to configure them correctly. In the end I figured it was taking too much time away from the actual task of just adding a simple contact form POC. 

Regarding the choice of spam controlling module, I looked into other modules such as "reCaptcha". I ended up going with the "honeypot" module as it works with the webform module almost "out of the box", you only need to enable it in webform settings. Another consideration was that it does not have an impact on the end users, as the functionality is hidden from them and does not require any action from their end when submitting webforms.

The reason for basing the category select options on a new taxonomy, is because the editors are accustomed to work with taxonomy terms compared to the difficulty navigating the webform settings. Another consideration is also to limit the editors access to the webform functionality as much as possible.


web/modules/custom/dpl_webform_configuration/dpl_webform_configuration.module
```diff
@@ -0,0 +1,16 @@
...
+ * Primary module hooks for dpl_webform_configuration module.
+ */
+
+use Drupal\webform\Entity\Webform;
+
+/**
+ * Implements hook_ENTITY_TYPE_create() for webform.
+ */
+function dpl_webform_configuration_webform_create(Webform $webform): void {
+  // Disable saving of results for all new webforms.
```
spaceo, line 14: Great with comment. I think you should add why we are disabling the saving.
Its always good to write the WHY in comments and in git commit massges.
I keep forgetting it myself 😆

spaceo (approved):
It is approved, but please look at my comment.

web/modules/custom/dpl_webform_configuration/dpl_webform_configuration.module
Dresse: I agree! I have added some more informative commenting now.

I am also trying to be better to do good commit messages, but practice takes time :D

PR #1007, opened 2024-04-17
Add missing order digital copy feedback text

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-517

#### Description
This PR adds a missing translation, and moves order digital copy feedback translations into global translations. In react apps, they are accessed through a helper file, which apparently causes the translations to fail in the CMS, so now they are global.

#### Screenshot of the result
-

-


spaceo (approved):
LGTM 👍

PR #1009, opened 2024-04-18
Set danish names for view pages, such as /articles. DDFFORM-502

rasben:
Ideally, we'd want to actually translate the path, but this is not an option in Drupal. We could create a URL alias, but then /articles would still be active.
I found a module named https://www.drupal.org/project/tvp that seemed to solve the problem, but in reality, it still relies on us creating a path alias, which has the same problem.

https://reload.atlassian.net/browse/DDFFORM-502

### DIFF
--- a/config/sync/views.view.articles.yml
+++ b/config/sync/views.view.articles.yml
@@ -273,7 +273,7 @@ display:
     position: 1
     display_options:
       display_extenders: {  }
-      path: articles
+      path: artikler
     cache_metadata:
       max-age: -1
       contexts:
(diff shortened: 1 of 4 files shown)


kasperg (commented):
> Ideally, we'd want to actually translate the path, but this is not an option in Drupal. We could create a URL alias, but then /articles would still be active.

I do not think this is the right approach.

In the current state no system paths are actually redirected to path aliases. Try accessing something like `node/11` or the like. It will happily show the node even though there is an alias.

Monosnap 3 gode til godnat 3-6 år | DPL CMS — Privat browsing 2024-04-18 10-39-58

As I see it adding and configuring the Redirect module as done in #1011 will fix this. This approah will work for whatever paths the url aliases are for.

With that in mind I think the proper way to address this is by providing path aliases to these paths as suggested in #1012.

PR #1010, opened 2024-04-18
Update media library views, to use 'contains' for search. DDFFORM-547

rasben:
Otherwise, the editor has to write the exact name of the media, to able to find it.

https://reload.atlassian.net/browse/DDFFORM-547

Screenshot 2024-04-18 at 09 41 37


kasperg (approved):
:+1: LGTM

PR #1011, opened 2024-04-18
Editor experience: URL aliases and redirects.

rasben:
**Install + Setup `redirect` module. DDFFORM-568**

This allows editors to create redirects, but also makes
sure that when URL aliases are updated, redirects are automatically
added from the old.

----

**Add help text, and toolbar link for redirects. DDFFORM-568**

This will make it easier for editors to understand and navigate.

----

**Allow editors and above to edit URL aliases. DDFFORM-567**

This does NOT give them access to /admin/config/search/path, but
it does give them access to override URL alias on specific nodes.
This is the intended need functionality.

Screenshot 2024-04-18 at 10 26 22

https://reload.atlassian.net/browse/DDFFORM-567

https://reload.atlassian.net/browse/DDFFORM-568


kasperg:
@rasben Note that the tests fail because with the Redirect module enabled `/user/login` now redirects to`/` while the tests try to check the path,

DPL CMS -- supports login (failed)

config/sync/core.extension.yml
```diff
@@ -116,6 +116,7 @@ module:
   purge_queuer_coretags: 0
   purge_tokens: 0
   recurring_events: 0
+  redirect: 0
```
kasperg, line 119: The Redirect module contains a submodule especially for creating redirects for 404s. I think we should discuss enabling this with the business. I think it would be helpful to administrators in the migration process after going live.

Monosnap Extend | DPL CMS | Logged in 2024-04-22 06-36-47

web/modules/custom/dpl_admin/dpl_admin.links.menu.yml
```diff
@@ -3,8 +3,18 @@ dpl_admin.add_content:
...
+dpl_admin.redirect:
+  title: "Redirects"
+  parent: system.admin
+  route_name: redirect.list
+  weight: 10
+dpl_admin.redirect_add:
+  title: "Add redirect"
+  parent: dpl_admin.redirect
+  route_name: redirect.add
+  weight: 1
```
kasperg, line 15: I think we need to discuss our strategy for adding items to the menu bar.

It might be me but I think all these top level custom menu entries are getting a bit out of hand.
As I see it they clutter up the menu contents for everyone but editors - and while they are many we have quite a few local administrators, administrators as well. They will have to access the Drupal menu tree anyway. While having direct access it nice these items also represent functionality that I do not think will be used on a day to day basis.

kasperg (approved):
:+1: LGTM

PR #1012, opened 2024-04-18
Add path aliases for views pages

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-502

#### Description

We have a few views which are displayed as pages: Articles, events and branches. These currently have system paths e.g. /events.

Add a path alias for each of these pages to use a user friendly url for each of these as static content.

Path aliases are entities so we can continue our use of the default content module to handle this.

The Redirect module will ensure that attempts to access the system urls will redirect to the aliases.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-cms/assets/73966/eefddf77-1b6f-43e1-8f31-e6fcd7fe81f6

This is an alternate take on #1009

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -58,6 +58,7 @@ module:
   dpl_something_similar: 0
   dpl_static_content: 0
   dpl_static_content_20240404_opening_hours: 0
+  dpl_static_content_20240418_views_aliases: 0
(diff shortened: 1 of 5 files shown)


rasben (approved):
👍 But - what is it with the naming of the module? Why add the date?

`dpl_static_content_**20240418**_views_aliases`

PR #1018, opened 2024-04-19
Eventinstance reference field

rasben:


kasperg:
Merged as a part of #1021

PR #1020, opened 2024-04-19
Remove access to performance settings + Custom cache clear. DDFFORM-607

rasben:
- Override the route subscriber, to avoid editors having access to /admin/config/development/performance despite having the correct permission ('configure basic settings')
- Create a custom 'cache clear' admin page, along with a description of the consequences.
- Include a link to the cache clear page in the gin secondary toolbar.

https://reload.atlassian.net/browse/DDFFORM-607


web/modules/custom/dpl_cache_settings/src/Form/CacheSettingsForm.php
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * When submitting the form, we want to clear the caches.
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state): void {
+    if ($this->moduleHandler->moduleExists('views')) {
+      views_invalidate_cache();
+    }
+    drupal_flush_all_caches();
```
kasperg, line 78: I wonder why we need to call `views_invalidate_cache()` explicitly here.

When looking at all the project code I cannot see how this would be called when using the default `ClearCacheForm`. If that is the case I would appreciate a comment explaining why we think it is relevant here.

web/modules/custom/dpl_cache_settings/src/Form/CacheSettingsForm.php
```diff
@@ -0,0 +1,86 @@
...
+
+    $this->logger->info($this->t(
+      'Cache has been cleared manually.', [], ['context' => 'DPL admin UX']));
```
kasperg, line 81: :+1: for logging this action explicitly.

This will allow us to keep track of how often this is used.

web/modules/custom/dpl_cache_settings/dpl_cache_settings.module
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Displaying the DPL CMS version in the secondary admin toolbar.
+ *
+ * @param array<mixed> $items
+ *   See hook_toolbar_alter().
+ */
+function dpl_cache_settings_toolbar_alter(array &$items): void {
+  $items['cache_clear'] = [
+    '#type' => 'toolbar_item',
+    'tab' => [
```
kasperg, line 21: I do not think it is a good idea to place access to this item in the secondary admin toolbar.

Clearing the caches it not something that we want administrators to do. We consider it a problem with our code if that is necessary. Putting it in the toolbar puts the option within plain sight. It takes fewer clicks to go to the clear cache form than the create article form.

Administrators already use the settings part of the administration interface so I do not see a reason why we cannot put our own custom form there as well.

kasperg (changes requested):
:+1: but I think we should reconsider the placement of this.

web/modules/custom/dpl_cache_settings/src/Form/CacheSettingsForm.php
rasben: I get the confusion, I should have written it somewhere - which I've done now, also as a code comment:

> These lines of code, are taken from a contrib module that does almost the same thing as us, but without the confirmation step that we need. Module is here: https://www.drupal.org/project/cacheflush

web/modules/custom/dpl_cache_settings/dpl_cache_settings.module
rasben: Any suggestions where to put it?
I get your point, but, I would imagine DDF is gonna get spammed when 85 libraries suddenly write, asking where the cache clear button is.

I've done stuff to make the cache clear page 'scary' - I've added the long warning text, and made the button red, like delete buttons.

kasperg: I suggest placing the form as an item under `/admin/config/system`.

I do not think we should be afraid that the functionality is placed within the administration settings:

1. We do not want or expect clearing the cache to be necessary
2. The functionality was also placed within administration settings in previous versions of the system

rasben: I've moved it away as you suggested

kasperg (approved):
:+1: Nice

PR #1022, opened 2024-04-21
DDFLSBP-566/add-twig-templates-for-webform-paragraph-elements

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-566

#### Description

In this PR we add new template for the webform paragraph. We also add a new custom template called form-element--webform.html.twig which are only targeting webform elements. 

The PR also adds 2 new hooks to the novel.theme. A preprocess hook for adding specific classes to input__submit elements on webforms only. We also add a theme suggestion hook for adding our new form-element--webform.html.twig template available. 

#### Screenshot of the result
Screenshot 2024-04-21 at 02 05 16

There are a few things I would like to pinpoint with this PR:
1. We currently have another [PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1005) open, which adds a new custom module called webform_configuration. I think it would make sense to move the 2 webform hooks added in this PR to the webform_configuration.module file. But this can be refactored at a later point if needed.

2. Right now we have some inconsistencies between the design-system contact-form story and the actual contact form. I think the design-system story needs some refactoring to make it up to date with how the rest of the paragraph stories is implemented. This can be discussed with the team at a later point.

### DIFF
--- a/web/themes/custom/novel/novel.theme
+++ b/web/themes/custom/novel/novel.theme
@@ -326,3 +326,18 @@ function novel_preprocess_views_view_unformatted(&$variables): void {
     }
   }
 }
+
+/**
+ *
+ */
+function novel_preprocess_input__submit(&$variables) {
+  $org_class = $variables['attributes']['class'];
+  $variables['attributes']['class'] += array_merge($org_class, [
+    'btn-primary',
+    'btn-filled',
+    'btn-small',
+    'arrow__hover--right-small',
+    'mt-48',
+    'mb-22',
+  ]);
+}
--- a/web/themes/custom/novel/novel.theme
+++ b/web/themes/custom/novel/novel.theme
@@ -328,16 +328,27 @@ function novel_preprocess_views_view_unformatted(&$variables): void {
 }
 
 /**
- *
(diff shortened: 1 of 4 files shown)


Adamik10 (approved):
The inconsistencies between design system and this PR you talk about in the PR description - those are eliminated in https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/602 right?

web/themes/custom/novel/novel.theme
```diff
@@ -326,3 +326,29 @@ function novel_preprocess_views_view_unformatted(&$variables): void {
...
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function novel_preprocess_input__submit(array &$variables): void {
+  $original_classes = $variables['attributes']['class'];
+
+  if (in_array('webform-button--submit', $original_classes)) {
+    $variables['attributes']['class'] += array_merge($original_classes, [
```
spaceo, line 337: The `+=` is not needed since `array_merge` is used.
So I think it should be:
```php
    $variables['attributes']['class'] = array_merge($original_classes, [
```

PR #1027, opened 2024-04-22
Opening hours repeated

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-632

#### Description

This adds support for handling repetitions for opening hours. The API and database schema is updated to support repetitions of different types.

In this first version focus is on weekly repetition and creation:

- When an opening hours instance with weekly repetition is to be created then the following things happen:
   - A repetition is created with the end date added as additional data 
   - An instance on each of the days in the repetition is created with the same data (start- and end-time, branch and category)
- When an opening hours instance is updated then the instance is converted to one without repetition. The remaining instances are not changed.
- When an opening hours instance is deleted then the instance is deleted. The remaining instances are not changed.

In the process I have created a kernel test case to verify that the opening hours resources work as expected.

I suggest that you review this commit by commit. There are a few fixups that I have added. They will be autosquashed in the final version. I hope you can disregard those.

The Cypress tests may be failing because the PR does not use the updated React components with support for repeating events.


web/modules/custom/dpl_opening_hours/src/Mapping/RepetitionMapper.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Mapper between value objects and OpenAPI request/response objects.
+ */
+class RepetitionMapper {
+
+  /**
+   * Map a request repetition to a value object.
+   */
+  public function fromRequest(RequestRepetition|ResponseRepetition $data): Repetition {
+    $repetitionType = ($data->getType() !== NULL) ? OpeningHoursRepetitionType::tryFrom($data->getType()) : NULL;
```
rasben, line 21: Cant you just do it like 
```suggestion
    $repetitionType =OpeningHoursRepetitionType::tryFrom($data?->getType());
```

to make it a bit simpler?

rasben (approved):
A small comment, to make a line more readable.

Apart from that, I suppose it looks good to me.
It's a bit hard for me to review, because the whole opening_hours codebase is unknown territory for me, but nothing jumps out

web/modules/custom/dpl_opening_hours/src/Mapping/RepetitionMapper.php
rasben: Fair enough, although, couldnt you just mark the string parameter as  `?string`? i get it's a bit silly to pass null to something, as it will never be a valid value, but it makes stuff like this much simpler

kasperg: For the record: `tryFrom()` is not a function we have defined. It is from the PHP Enum handling. As such we cannot change it.

PR #1034, opened 2024-04-23
DDFLSBP-528/create contact page and contact page content

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-528

#### Description

This PR adds a few new things:
1. Adds a new static content module used for creating a default contact page containing a contact webform paragraph. It also adds 4 taxonomy terms in webform_email_categories, which are used in the contact webform.

2. Updates the contact webform configuration with the following changes:
- Adds a new html element that adds a notice for the user not to enter CPR numbers or other sensitive data.
- Adds danish labels for the contact elements as well as the naming of the form itself.

3. Updates the webform_email_categories taxonomy configuration with the following changes:
- Make sure that all terms added to the vocabulary is unique as we don't want duplicate terms.
- Make sure that the terms default language is danish.

4. Adds a new template for status-messages of the type "status" which are now used when showing status messages to the user. An example of this could be when a user submits a webform successfully. I have created a story for it in the design system as well as adding the new styling: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/607 

#### Screenshot of the result
Screenshot 2024-04-24 at 01 11 42

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -58,6 +58,7 @@ module:
   dpl_something_similar: 0
   dpl_static_content: 0
   dpl_static_content_20240404_opening_hours: 0
+  dpl_static_content_20240422_contact_page: 0
   dpl_url_proxy: 0
   dpl_webform_configuration: 0
   drupal_typed: 0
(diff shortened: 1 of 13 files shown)


clausbruun (approved):
LGTM 🎉

PR #1042, opened 2024-04-25
API endpoints for events - `/api/v1/events*`

rasben:
**Update OpenAPI permissions and settings for events API. DDFFORM-609**

We dont want the external user to have 'edit eventinstances' permission,
as the updating of states actually is managed by Drupal, not the
user. The user just have access to the PATCH endpoint.

**Move event logic to centralized eventhelper service. DDFFORM-609**

We have logic spread out, across plugins, classes and hooks, for
stuff such as pulling the relevant event address, and getting the
event state. This commit takes that logic, and places it in a
central service instead, that can be injected where needed.
This also makes it much easier to maintain, when we start to
need some of this functionality for the event API.

**Extendable EventResourceBase, for uses for endpoints. DDFFORM-609**

**Create Mapping "service", for Event API endpoints to understand OpenAPI.**

The EventRestMapper takes either an EventInstance and turns it into
an OpenAPI response object (generated by CodeGen), or a OpenAPI request
(also generated), and turns it into an eventinstance.

This is necessary for our API endpoints, which will be in a future
commit. DDFFORM-609

**All events endpoint - `GET`:`/api/v1/events` - DDFFORM-609**

We use the EventRestmapper to turn eventinstances into objects
that we can serialize and return as responses.
We also make sure to cache the responses as precisely as possible.

**Update event endpoint - `PATCH`:`/api/v1/event/{uuid}` - DDFFORM-609**

Taking a patch request, trying to find relevant eventinstance, and update it with event state date.
**This request only handles event state.**
"External Data" will be handled in the future, as no fields exist at the moment.

-----

https://reload.atlassian.net/browse/DDFFORM-609


web/modules/custom/dpl_event/src/Plugin/rest/resource/v1/EventsResource.php
```diff
@@ -221,30 +229,59 @@
...
-    return new ResourceResponse([]);
+    $storage = $entity_type_manager->getStorage('eventinstance');
+    $ids = $storage->getQuery()
+      ->accessCheck(FALSE)
+      ->condition('status', TRUE)
+      ->execute();
+
+    $response_events = [];
+
+    $cache_manager = DrupalTyped::service(CacheBackendInterface::class, 'cache.data');
```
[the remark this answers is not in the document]
rasben: Especially considering this is an open endpoint, and if a user just non-stop refreshes the page, they will be running the mapper non-stop.

web/modules/custom/dpl_event/src/Plugin/rest/resource/v1/EventResource.php
```diff
@@ -89,35 +93,38 @@
...
+
+    $event_instance = reset($event_instances);
+
+    if (!($event_instance instanceof EventInstance)) {
+      return new ModifiedResourceResponse('Even not found', 404);
+    }
+
+    $state = $request_data->getState();
+    $event_instance->set('field_event_state', $state);
+    $event_instance->save();
```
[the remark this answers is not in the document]
rasben: I'll update the ticket link, and I've created this ticket for the editor link:

https://reload.atlassian.net/browse/DDFFORM-724

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
```diff
@@ -0,0 +1,246 @@
...
+
+    if (empty($start) || empty($end)) {
+      return NULL;
+    }
+
+    $date_start = new DateTime();
+    $date_start->setTimestamp(strtotime($start));
+
+    $date_end = new DateTime();
+    $date_end->setTimestamp(strtotime($end));
```
kasperg, line 105: Consider moving all this logic to `EventWrapper`.

web/modules/custom/dpl_rest_base/dpl_rest_base.info.yml
```diff
@@ -0,0 +1,7 @@
+name: "DPL REST base"
+type: module
+description: "A base module, sharing functionality between e.g. events and opening hours."
```
kasperg, line 3:
```suggestion
description: "A base module for providing shared functionality between modules exposing REST services."
```

web/modules/custom/dpl_event/src/Plugin/rest/resource/v1/EventsResource.php
```diff
@@ -221,30 +222,41 @@
...
+
+    foreach ($ids as $id) {
+      $event_instance = $storage->load($id);
+
+      if ($event_instance instanceof EventInstance) {
+        $event_responses[] = $this->mapper->getResponse($event_instance);
+      }
+    }
+
+    $event_responses = $this->serializer->serialize($event_responses, 'application/json');
```
kasperg, line 249:
```suggestion
    $event_responses = $this->serializer->serialize($event_responses, $this->serializerFormat($request));
```

This is the same approach we use elsewhere instead of hard coding the type

PR #1043, opened 2024-04-25
Tidy up admin toolbar, organizing in DDF and Drupal. DDFFORM-617

rasben:
The toolbar has grown massive, with a lot of useful links. A lot of the toolbar also promotes stuff that 99% of the editors will never want to see or use. (Or even have access to). Instead, we create a DDF dropdown, with relevant settings, and a Drupal dropdown, for Drupal developers (and admins).

#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-617

Screenshot 2024-04-25 at 22 02 44
Screenshot 2024-04-25 at 22 03 01

### DIFF
--- a/config/sync/core.menu.static_menu_link_overrides.yml
+++ b/config/sync/core.menu.static_menu_link_overrides.yml
@@ -37,3 +37,33 @@ definitions:
     weight: 5
     expanded: false
     enabled: true
+  system__admin_structure:
+    menu_name: admin
+    parent: dpl_admin.drupal
+    weight: -8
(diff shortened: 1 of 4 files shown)


rasben:
@kasperg and i have decided to find an easier solution, and re-visit this fantastic code later on.

Dresse:
> We have decided to find an easier solution, and re-visit this fantastic code later on.

I haven't look into this until now, but I think this would be a really nice addition! 
Can I ask why you decided to walk a different path?

PR #1044, opened 2024-04-25
Sanity-check and cleanup of permissions, accross roles. DDFFORM-711 DDFFORM-625

rasben:


config/sync/user.role.administrator.yml
```diff
@@ -190,6 +188,7 @@ permissions:
   - 'role-assign users by role'
   - 'schedule publishing of eventseries'
   - 'schedule publishing of nodes'
+  - 'select account cancellation method'
```
rasben, line 191: local_admin had this permission, but admin doesnt. seems like a mistake.

config/sync/user.role.administrator.yml
```diff
@@ -86,7 +85,6 @@ permissions:
   - 'administer nodes'
   - 'administer openid connect clients'
   - 'administer orphaned events entities'
-  - 'administer paragraphs categories'
```
rasben, line 89: this just confused the editors, as they cant edit the paragraphs to assign categories.

config/sync/user.role.local_administrator.yml
```diff
@@ -61,6 +60,7 @@ permissions:
   - 'access files overview'
   - 'access media overview'
   - 'access site in maintenance mode'
+  - 'access site reports'
```
rasben, line 63: this seems pretty harmless, and useful.

config/sync/user.role.external_system.yml
```diff
@@ -5,14 +5,14 @@ dependencies:
...
     - rest
     - system
+    - toolbar
 id: external_system
 label: 'External system'
 weight: 7
 is_admin: null
 permissions:
-  - 'edit eventinstance entity'
+  - 'access toolbar'
```
rasben, line 16: external users needs access to the toolbar, to be able to edit their user password etc.

config/sync/user.role.mediator.yml
```diff
@@ -100,11 +100,14 @@ permissions:
...
   - 'use text format basic'
   - 'use text format limited'
+  - 'view all media revisions'
   - 'view any unpublished content'
   - 'view own unpublished content'
   - 'view own unpublished media'
   - 'view scheduled content'
   - 'view scheduled eventseries'
+  - 'view scheduled media'
+  - 'view scheduled taxonomy_term'
```
kasperg, line 110: Do you have any comments about why these should be added?

kasperg (approved):
:+1: LGTM

I appreciate the comments you have added to this PR. I would have appreciated them even more if they were commit comments in separate commits 😉 .

config/sync/user.role.mediator.yml
rasben: Because it matches other permissions, and i think they're missing by mistake

PR #1045, opened 2024-04-25
Limit some text body fields to basic format. DDFFORM-613

rasben:
Also includes migration, so the editors wont end up with a strange warning about having used the old one.

https://reload.atlassian.net/browse/DDFFORM-613

### DIFF
--- a/web/modules/custom/dpl_admin/dpl_admin.deploy.php
+++ b/web/modules/custom/dpl_admin/dpl_admin.deploy.php
@@ -1,0 +1,55 @@
+<?php
+
+/**
+ * @file
+ * Admin deploy hooks.
+ *
+ * These get run AFTER config-import.
+ */
+
+/**
+ * We've limited the text formats for CKEditor fields. Let's move old to new.
+ */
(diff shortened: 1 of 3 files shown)


web/modules/custom/dpl_admin/dpl_admin.deploy.php
```diff
@@ -0,0 +1,55 @@
...
+
+  $paragraphs = \Drupal::entityTypeManager()->getStorage('paragraph')->loadMultiple($pids);
+
+  foreach ($paragraphs as $paragraph) {
+    /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
+
+    $values = $paragraph->get($field_name)->getValue();
+
+    foreach ($values as &$value) {
+      $value['format'] = 'basic';
```
kasperg, line 41: Consider renaming the function and making the entity type and format arguments as well.

This would help make the function more generic if we get to a point where we want to move this and other similar functions to a general module with migrations.

kasperg (approved):

PR #1046, opened 2024-04-26
Add front-facing user name on articles. DDFFORM-542

rasben:
Create a authorname field, and displaying it by default on the articles.

Also allows people to change the author (As in, the user), of articles.

### DIFF
--- a/web/themes/custom/novel/templates/fields/username.html.twig
+++ b/web/themes/custom/novel/templates/fields/username.html.twig
@@ -3,8 +3,4 @@
   {% set name = account.get('field_author_name').getString()|default(name) %}
 {% endif %}
 
-{% if link_path -%}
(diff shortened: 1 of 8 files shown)


web/themes/custom/novel/templates/fields/username.html.twig
```diff
@@ -1,5 +1,10 @@
+{# If the user has an author name, we want to use that instead of username. #}
+{% if account.hasField('field_author_name') %}
+  {% set name = account.get('field_author_name').getString()|default(name) %}
+{% endif %}
```
kasperg, line 4: Consider if this belongs in a preprocess hook instead of in the template.

kasperg (approved):
:+1: LGTM

I wonder if it would be helpful to add display of author name and support for search in author name in the user list.

Monosnap People | DPL CMS | Logged in 2024-04-26 13-59-10

PR #1049, opened 2024-04-26
DDFFORM 475

LasseStaus:
#### Link to issue
[DDFFORM-475](https://reload.atlassian.net/browse/DDFFORM-475)

#### Description

This PR does several things. At large this is: 

- Replace old view.event_admin with view.event_series_admin which shows the event series on /admin/content/events. 
- Create a tab on the eventseries edit page, showing a list of all eventinstances in that seris. 
- Enforces that tab is only showed conditionally, see commits for info. 
- Create eventsubcriber for redirecting both editors and users on eventinstances/eventseries conditionallity. See commits for info. 
- Disable the ability of cloning eventseries / eventinstances. 
- Disables the possiblity of viewing revisions for eventseries / eventinstances. 

#### Manual test

Test that you; 
- Cannot access the frontend of an event series with only 1 instance. (you are redirected)
- Cannot access the editing page of an event instance, which is the only instance of the series. (You are redirected) 
- Can see a view of eventseries, which relevant information correctly formatted and with necessary links. 
- Can see a tab for "Edit instances" for series that have more than one eventinstance. 
- Can see a view of eventinstances under the "edit instances" tab, which relevant information correctly formatted and with necessary links. 
- Cannot see "clone" or "revisions" as a non-superadmin user. 

#### Screenshot of the result

New series view: 
image

New instance view: 
image

Please carefully consider if anything has been missed or overlooked in the replacement/removal of the old view for events. 

Also consider if there are any permissions or general view settings that have been missed. 

[DDFFORM-475]: https://reload.atlassian.net/browse/DDFFORM-475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/modules/custom/dpl_event/src/Access/EventSeriesTabAccessCheck.php
```diff
@@ -0,0 +1,55 @@
...
+
+    $eventSeriesEntity = $this->entityTypeManager->getStorage('eventseries')->load($eventSeriesId);
+    if (!$eventSeriesEntity || !method_exists($eventSeriesEntity, 'getInstanceCount')) {
+      return AccessResult::forbidden();
+    }
+
+    $instanceCount = $eventSeriesEntity->getInstanceCount();
+    return AccessResult::allowedIf($instanceCount > 1)
+      ->addCacheableDependency($eventSeriesEntity)
+      ->cachePerPermissions();
```
kasperg, line 52: About the caching problem we saw earlier: I think the problem may be that creating an event instance does not clear the cache for the related event series.

This may also explain the problems we have for displaying the number of instances for each event.

This will probably require some help from either @rasben or myself.

web/modules/custom/dpl_event/src/EventSubscriber/EventsRedirect.php
```diff
@@ -0,0 +1,108 @@
...
+  protected $entityTypeManager;
+
+  /**
+   * Constructs a new EventsRedirect object.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager) {
+    $this->entityTypeManager = $entity_type_manager;
```
kasperg, line 34: Consider changing this to constructor property promotion like we have for `EventSeriesTabAccessCheck`.

You reduce the need for an explit property declaration and we also get actual type safety by setting the type on the propety in PHP instead of in a DocBlock.

web/modules/custom/dpl_event/src/Access/EventSeriesTabAccessCheck.php
```diff
@@ -0,0 +1,55 @@
...
+   *   otherwise forbidding it.
+   */
+  public function access(RouteMatchInterface $route): AccessResult {
+    $eventSeriesId = $route->getParameter('eventseries');
+    if (!$eventSeriesId) {
+      return AccessResult::forbidden();
+    }
+
+    $eventSeriesEntity = $this->entityTypeManager->getStorage('eventseries')->load($eventSeriesId);
+    if (!$eventSeriesEntity || !method_exists($eventSeriesEntity, 'getInstanceCount')) {
```
kasperg, line 45: Please check whether the loaded object is an instance of the expected class instead of checking whether the method exists.

This is a more explict way of ensuring we have the expected value. 

In fact we know that the entity storage for event series will always return an `EventSeries` object. [We can tell that to PHPStan](https://github.com/mglaman/phpstan-drupal?tab=readme-ov-file#entity-storage-mappings) but PHPStorm will not know anyway so I guess having an explicit check here is good for now.

web/modules/custom/dpl_event/src/Access/EventSeriesTabAccessCheck.php
```diff
@@ -0,0 +1,55 @@
...
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+
+/**
+ * Provides an access checker for the event series tab.
+ *
+ * This access check ensures that the tab for editing event instances
+ * is only available when there are more than one instance in the series.
+ */
+class EventSeriesTabAccessCheck implements AccessInterface {
```
kasperg, line 16: I appreciate the code comment.

As I see it the current comment explains what the code does. It could be improved by adding a sentence or two about why this is necessary. This also goes for other DocBlocks.

web/modules/custom/dpl_event/src/Access/EventSeriesTabAccessCheck.php
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * Checks if there are more than one instance in the event series and grants
+   * or denies access based on this condition.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route
+   *   The route match interface to get route parameters.
+   *
+   * @return \Drupal\Core\Access\AccessResult
+   *   The access result, allowing access if there are more than one instance,
+   *   otherwise forbidding it.
```
kasperg, line 36: Consider leaving out these parameters comments.

They are not required by our coding style and if they do not add any value they are just unecessary lines of code.

web/modules/custom/dpl_event/src/EventSubscriber/EventsRedirect.php
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents(): array {
+    return [
+      KernelEvents::REQUEST => [
+        ['checkEventSeriesRedirect'],
+        ['checkEditInstanceRedirect'],
+      ],
+    ];
```
kasperg, line 46: Please reconsider this.

On the PR environment, I cannot tell that viewing the single instance series, https://varnish.pr-1049.dpl-cms.dplplat01.dpl.reload.dk/bronshoj-bibliotek/events/design-teknologi/fernisering-moderne-dans-0, redirects to the instance, https://varnish.pr-1049.dpl-cms.dplplat01.dpl.reload.dk/bronshoj-bibliotek/events/design-teknologi/fernisering-moderne-dans.

On the other hand the redirect from the instance, https://varnish.pr-1049.dpl-cms.dplplat01.dpl.reload.dk/bronshoj-bibliotek/events/design-teknologi/fernisering-moderne-dans to the edit form for the series work, https://varnish.pr-1049.dpl-cms.dplplat01.dpl.reload.dk/events/series/3/edit .

If you cannot determine why this occurs then consider splitting this into two different event subscribers.

kasperg (changes requested):
:+1: I think this looks good. I have tested as requested and it mostly works well.

I have a few comments. My main reason for blocking this is the redirects and caching issues we saw earlier.

web/modules/custom/dpl_event/src/Access/EventSeriesTabAccessCheck.php
LasseStaus: This should be looked at together.

web/modules/custom/dpl_event/src/Access/EventSeriesTabAccessCheck.php
kasperg: > Currently they are required in phpcs tooling. Not sure if i should change that, so i left them in

OK. I am a bit surprised by this. The next time you run into this please let me know.

kasperg (approved):
:+1: LGTM

PR #1054, opened 2024-04-29
DDFFORM 126

LasseStaus:
#### Link to issue

[DDFFORM-126](https://reload.atlassian.net/browse/DDFFORM-126)

Related design system PR: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/613

#### Description

This PR adds a section on an eventseries page that will show the eventinstances for that series. It uses the same view as stack_parent + stacked_event as on /events page. 

#### Screenshots

image

[DDFFORM-126]: https://reload.atlassian.net/browse/DDFFORM-126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web/modules/custom/dpl_event/dpl_event.module
+++ b/web/modules/custom/dpl_event/dpl_event.module
@@ -195,6 +195,23 @@ function dpl_event_preprocess_eventseries(array &$variables): void {
   $variables['formatted_date'] =
     \Drupal::service('dpl_event.reoccurring_date_formatter')
       ->getSeriesDateString($event_series);
+
+  if ($event_series->getInstanceCount() < 2) {
+    return;
+  }
+
+  // Load event instances related to this series and prepare them for rendering.
+  $eventinstances_in_series = $event_series->get('event_instances')->referencedEntities();
+
+  $variables['event_instances'] = [];
+  foreach ($eventinstances_in_series as $index => $eventInstance) {
+    $viewMode = $index === 0 ? 'list_teaser_stacked_parent' : 'stacked_event';
+
+    $renderable_instances = \Drupal::entityTypeManager()
+      ->getViewBuilder('eventinstance')->view($eventInstance, $viewMode);
+
+    $variables['event_instances'][] = $renderable_instances;
+  }
 }
 
 /**
--- a/web/themes/custom/novel/templates/components/full-event.html.twig
+++ b/web/themes/custom/novel/templates/components/full-event.html.twig
@@ -29,5 +29,14 @@
     } only %}
   </section>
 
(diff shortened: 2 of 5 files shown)


web/themes/custom/novel/templates/components/full-event.html.twig
```diff
@@ -29,5 +29,14 @@
...
 
+  {% if event_instances %}
+    <section class="event-list-stacked">
+      <h2 class="event-list-stacked__heading">{{ 'Upcoming events'|t }}</h2>
+      {% for instance in event_instances %}
+          {{ instance }}
+      {% endfor %}
+    </section>
+  {% endif %}
+
```
kasperg, line 40: Consider if the instances should be shown before the paragraphs.

I would have expected it to be below as I consider the paragraphs a part of the event "content" while the list of instances is something extra but I have not been part of the conversation with the business so you know best here.

web/themes/custom/novel/templates/fields/field--field-event-paragraphs.html.twig
kasperg, line 1: :+1: nice catch!

web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -195,6 +195,24 @@ function dpl_event_preprocess_eventseries(array &$variables): void {
   $variables['formatted_date'] =
     \Drupal::service('dpl_event.reoccurring_date_formatter')
       ->getSeriesDateString($event_series);
+
+  $event_series_instance_count = $event_series->getInstanceCount();
+  if (!$event_series_instance_count > 1) {
+    return;
+  }
```
kasperg, line 202:
```suggestion
  if ($event_series->getInstanceCount() < 2) {
    return;
  }
```

As I see it there is no reason to keep `event_series_instance_count` in a separate variable.

kasperg (approved):
:+1: Nice changes.

I have a few questions and considerations.

When you want to show this to the business you need to setup the PR to use your changes to the design system. You can use a task to do so:

```
BRANCH=DDFFORM-126 VERSION=2024.18.0-rc2 task dev:composer:update-design-system
```

web/themes/custom/novel/templates/components/full-event.html.twig
LasseStaus: Yeah i understand. 

We agreed to put it above the paragraphs to begin with, but maybe they will change their mind when they see it 🤷

PR #1055, opened 2024-04-29
Remove `<p>` tags from events and articles

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-753

#### Description
This pull request removes the `<p>` for events and articles. 

#### Screenshot of the result
image

image

image

### DIFF
--- a/web/themes/custom/novel/templates/fields/field--node--field-subtitle--list-teaser.html.twig
+++ b/web/themes/custom/novel/templates/fields/field--node--field-subtitle--list-teaser.html.twig
@@ -1,5 +1,3 @@
-<p>
-  {% for item in items %}
-    {{ item.content }}
-  {% endfor %}
-</p>
+{% for item in items %}
+  {{ item.content }}
+{% endfor %}
--- a/web/themes/custom/novel/templates/components/list-item.html.twig
+++ b/web/themes/custom/novel/templates/components/list-item.html.twig
@@ -29,6 +29,7 @@
         {{ description }}
       </div>
     {% endif %}
+
     {% if bottom.0 %}
       <div class="content-list-item__content-bottom-container">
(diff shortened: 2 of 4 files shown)


web/themes/custom/novel/templates/fields/field--node--field-subtitle--list-teaser.html.twig
```diff
@@ -1,5 +1,3 @@
-<p>
-  {% for item in items %}
-    {{ item.content }}
-  {% endfor %}
-</p>
+{% for item in items %}
+  {{ item.content }}
+{% endfor %}
```
LasseStaus, line 3: I'm unsure if this template is used/necessary, or is a leftover from [this commit](https://github.com/danskernesdigitalebibliotek/dpl-cms/commit/c0434c73bf5c03752b0bbe453b80e1cf4b9d415d#diff-cf874922e4745a51aecacb001512f6a8f38febaec430dd708d1955dc1151d53d)

LasseStaus (approved):
Looks good when im testing it. 

I'm not sure if everything that is not used is removed, see comment. If the template is not used then remove it. Maybe also check the commit referenced in the comment in full to see if there are other things that were originally added that should be removed. 

I will approve, but i think you could clean the commits before merging. The first commit creates a bunch of things, that are then either modified or deleted which i think is unncessary.

web/themes/custom/novel/templates/fields/field--node--field-subtitle--list-teaser.html.twig
kasperbirch1: It is needed because is else will use `web/themes/custom/novel/templates/fields/field--field-subtitle.html.twig`

PR #1058, opened 2024-04-29
Make sure structure menu is accessible for editors. DDFFORM-617

rasben:
Apparantly, to have access to the structure menu in general, users need a generic permission.

There's a bug in Core, that means that links show up in the config dropdown, despite not being accessible. I've added a patch against this.

https://reload.atlassian.net/browse/DDFFORM-617


kasperg (approved):
:+1: 🎉

PR #1068, opened 2024-04-30
Programmatically add translations for un-capitalized weekdays. DDFFORM-622

rasben:
This is particularly important for eventseries, where it currently says "Hver saturday" - DDF will not be able to translate it, until the actual word shows up, so instead, we do it for them.

https://reload.atlassian.net/browse/DDFFORM-622

### DIFF
--- a/web/modules/custom/dpl_event/dpl_event.install
+++ b/web/modules/custom/dpl_event/dpl_event.install
@@ -6,14 +6,18 @@
  */
 
 use Drupal\Core\Field\BaseFieldDefinition;
+use Drupal\drupal_typed\DrupalTyped;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\locale\SourceString;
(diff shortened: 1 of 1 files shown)


kasperg (commented):
Normally [we use code scanning for handling translations of UI strings](https://danskernesdigitalebibliotek.github.io/dpl-docs/DPL-CMS/translation/#translation-system).

If you do it this way around you have to ensure that this does not get overridden by the current automations and processes we already have in place.

I cannot tell form the PR whether this is the case or not.

rasben:
The problem is that until the system detects it, it wont show up in the translation system.
So, for example, `thursday` doesnt even exist in the system at the moment as a source string, and that makes it impossible for DDF to translate.

Seeing as this is pretty unambigious translations (how would "Monday" ever be translated to other than "Mandag"?), then I dont worry too much about it overriding or being overriden

rasben:
I will close this PR, and replace it with https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1119 as lagoon cannot deploy this, with this long branch name.

PR #1074, opened 2024-04-30
Transform Azure mailer env variable into consumable configuration

spaceo:
Issue:
https://reload.atlassian.net/browse/DDFLSBP-614

By creating a separate module for handling mail functionality, starting with azure mailer conf, we have asolution that is modular, encapsulated and easy to test.


web/profiles/dpl_cms/dpl_cms.profile
```diff
@@ -98,3 +99,28 @@ function dpl_cms_locale_translation_batch_fetch_finished(bool $success, array $r
...
+  array_map(function ($element) use (&$config) {
+    [$key, $value] = explode('=', $element, 2);
+    if ($key === 'endpoint') {
+      $value = parse_url($value, PHP_URL_HOST);
+    }
+    $config['azure_mailer.settings'][$key] = $value;
+  }, $parts);
+
+  return $config;
+}
```
Dresse, line 126: I don't think the function in dpl_cms.profile is needed anymore after moving the logic to its own module?

Dresse (changes requested):
LGTM - But we need to remove the changes in dpl_cms.profile

PR #1065, opened 2024-04-30
Disable `daily` and `yearly` events

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-521

#### Description

This pull request disables the `daily` and `yearly` events in favor of using `Consecutive Event`. This change aims to reduce confusion

#### Screenshot of the result
image

PR #1075, opened 2024-05-01
Demo 2024.18.0

kasperg:
**This was previously a PR environment for a feature. Due to temporary problems spawning new environments we started using it for testing of release 2024.18.0 as the branch is based on this release and a minimum of other changes.**

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-521

#### Description

This PR reintroduces the changes from #1065 which were reverted in #1073 because the site raised a warning after creating a series.

The warning was caused by a problem in the recurring events module. This PR also patches the module to address this.

To test this I sugges you do the following:

- Go to the creation form for an event series
- See that daily and yearly events are disabled
- Create a series (whatever type should work)
- See that no error is raised

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -335,7 +335,8 @@
                 "Combat NoCorrespondingEntityClassException": "patches/purge-5340bcf-combat-nocorresponding-entity-exception.patch"
             },
             "drupal/recurring_events": {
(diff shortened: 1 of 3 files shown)


LasseStaus (approved):
GJ finding the error & making the change. 

Works when i test it so LGTM 🚀

PR #1076, opened 2024-05-01
DDFLSBP-542 - Setup permissions for webforms

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-542

#### Description

DISCLAIMER:  I suggest looking at the overall file changes instead of going by commits as I haven't had time to properly sort the commits into meaningful chunks. 

This PR is adding logic and configuration for handling webform permissions and other restrictions such as element restriction.

This is done by adding 7 new custom permissions for managing different parts of the webform configuration:

- 3 of them completely restricts the access of the "access", "form" and "submission" menues.
- 3 of them restricts parts of configuration in "confirmation", "general" and "handler" menues with form_alter hooks.
- The last permission restricts parts of the element configuration when adding new elements.

The PR also adds configuration of the default webform permissions to restrict several part of the webform configuration to administrators only.

Lastly the PR also adds configuration limiting the number of available elements to choose from when creating new webforms. We currently only allow elements for which we have added styling for in the dpl-design-system.

This PR is based on conversations between us and the client and is supposed to be the first version of a possible permission setup. The setup will likely change in the future as more or less restricting of webform configuration is required or when new elements needs to be useable for building forms.


spaceo (approved):
LGTM 👍

PR #1079, opened 2024-05-01
Ensure menu cache is cleared when first item is added

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-618

#### Description

This PR switches the patch uses to solve this issue. It turns out the current approach is not correct so we try again by providing a new patch.

This can be tested on the PR environment by:

1. Emptying any existing menu items
2. Rendering the front page
3. Adding a new menu item to the main menu
4. Rendering the front page
5. Seeing that the new menu item is added


LasseStaus (approved):
LGTM 👍  

Works when testing on the PR. 

Reminder that CI fails.

PR #1081, opened 2024-05-01
Update `checkMultipleOpeningHoursExistAdmin` to `checkAtLeastOneOpeningHoursExistAdmin`

kasperbirch1:
#### Description

This revision ensures the system properly handles scenarios where only a single opening hour exists at the end of the first month or the start of the next month. This behavior stems from the use of `oneMonthFromToday` as the end date in our opening time series.


kasperg (approved):
:+1: LGTM

PR #1085, opened 2024-05-02
Allow editors to create tags from content creation. DDFFORM-671

rasben:
Replaces a multiselect field, with the "tag-style" search field. That way, the editors can actually create new tags directly.

https://reload.atlassian.net/browse/DDFFORM-671

### DIFF
--- a/config/sync/core.entity_form_display.eventinstance.default.default.yml
+++ b/config/sync/core.entity_form_display.eventinstance.default.default.yml
@@ -29,7 +29,8 @@ dependencies:
     - paragraphs
     - paragraphs_ee
     - paragraphs_features
-    - select2_multicheck
+    - path
+    - select2
 third_party_settings:
   field_group:
     group_teaser_card:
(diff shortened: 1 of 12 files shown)


LasseStaus (changes requested):
I think we have agreed not to use select2_multicheck or autocomplete, and instead go forward with select2. 

I suggest you: 
- Change the current alterations to use "Select2" instead, this has the feature of creating tags on the fly aswell. 
- Expand the scope of the ticket to change anything using select2_multicheck to just use select2, in order to streamline our field usage as agreed. 
- Uninstall the select_multicheck since it should no longer be used, and if there are other unused modules as part of it remove these aswell. 

Also, we've found out that something can go (goes?) wrong in lagoon, for branches with names that exceed 50 characters. So if the deploy is acting weird, this could be the issue. 

Merge is pointing to develop atm also! 🚀

LasseStaus (approved):
LGTM 🚀

PR #1089, opened 2024-05-02
Implement new breadcrumb design. DDFFORM-608

rasben:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-608

Depends on to be merged:

https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/618

### DIFF
--- a/web/themes/custom/novel/templates/components/breadcrumb.html.twig
+++ b/web/themes/custom/novel/templates/components/breadcrumb.html.twig
@@ -1,15 +1,5 @@
-<div class="p-8">
-  <nav role="navigation">
-    <h2 class="hide-visually">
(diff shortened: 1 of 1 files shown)


web/themes/custom/novel/templates/components/breadcrumb.html.twig
```diff
@@ -1,15 +1,5 @@
-<div class="p-8">
-  <nav role="navigation">
-    <h2 class="hide-visually">
-      {{ 'Breadcrumb'|t }}
```
rasben, line 4: This is removed on purpose - as we use a `<nav>`, it will be plenty for the screenreaders.

kasperg (approved):

PR #1098, opened 2024-05-03
Added quickfix for disabling cache on the user creation page, to prevent

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-618

#### Description

This PR adds a quickfix for disabling cache on the user creation page, to prevent the possibility for cached redirects.


spaceo (approved):
LGTM 👍

PR #1100, opened 2024-05-06
DDFLSBP-603: Disable relations option on Webform Emails Categories taxonomy

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-603

#### Description

This PR aims to limit the configuration possibilities when adding webform paragraphs to content as well as limiting the nesting possibility on specific taxonomy terms to avoid accidentally adding nested terms.

The following changes is done:
- Added and enabled the 'flat_taxonomy' module.
- Enabled flat taxonomy on 'Opening Hours Categories' and 'Webform Email Categories' taxonomies.
- Disabled 'default_data' on webform paragraph, which removes the possibility to add custom YAML configuration on the chosen webform.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -114,6 +114,7 @@
         "drupal/facets": "^2.0",
         "drupal/field_group": "^3.4",
         "drupal/field_inheritance": "^2.0",
+        "drupal/flat_taxonomy": "^2.0",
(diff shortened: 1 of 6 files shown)


spaceo (approved):
Great. Seems good to me 👍

PR #1104, opened 2024-05-06
Opening hours API cacheable list responses

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-634

#### Description

[Add support for caching opening hours list requests](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1104/commits/a1ffa8f3616679fd55216142e9466c41588b49c4)
[a1ffa8f](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1104/commits/a1ffa8f3616679fd55216142e9466c41588b49c4)

This updates the list resource to return cacheable responses. These
can then be cached at different points on the way to the end user -
most prominently in Varnish.

We only use a single cache tag to handle all lists and use cache
contexts to create variations. This way we only have to invalidate
this tag after any operation that might affect any list. Somewhat
crude but also simple.

Update our test to show that cache tags are returned an invalidated
as expected.

If you want to test this do the following:

- Go to a page that fetches opening hours in the frontend of your local development environment (patron facing, Swagger UI or curl)
- Check the API response in the developer tools.
- See that the response headers contain `X-Varnish-Cache` and `X-Drupal-Dynamic-Cache`. From now on if you are accessing the API from swagger UI we use the header `X-Drupal-Dynamic-Cache` and from patron facing UIs or Curl ``X-Varnish-Cache`.
- See that it probably says `MISS` on the initial page load
- Refresh the page
- See that the header now says `HIT`
- Update opening hours using the editor, Swagger UI or the like (create, edit or delete and instance)
- Fetch opening hours again
- See that it says `MISS` on the initial request
- Fetch the opening hours again
- See that the header now says `HIT`
- Call the API with other parameters e.g. from date or to date
- See that the header now says `MISS`

API response when using Swagger UI as an authenticated user
Monosnap OpenAPI Documentation | DPL CMS | Logged in 2024-05-07 10-30-48

API response when using curl (remember to include `-v` to see headers)

Monosnap kasper@Hypercycle:~ 2024-05-07 10-33-29

### DIFF
--- a/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursCreateResource.php
+++ b/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursCreateResource.php
@@ -64,7 +64,10 @@ final class OpeningHoursCreateResource extends OpeningHoursResourceBase {
     try {
       $requestData = $this->deserialize(OpeningHoursRequest::class, $request);
       $instance = $this->mapper->fromRequest($requestData);
+
       $createdInstances = $this->repository->insert($instance);
+      $this->invalidateCache();
+
       $responseData = array_map(function (OpeningHoursInstance $instance) {
         return $this->mapper->toResponse($instance);
       }, $createdInstances);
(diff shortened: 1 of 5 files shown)


rasben (approved):
I havent tested, because your description and images were enough "documentation" for me.

The code looks fine, I dont see any gotchas :)

PR #1105, opened 2024-05-06
Opening hours API url update

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-634

#### Description

Introduce the `/api/v1` prefix for the opening hours API. This brings it in line with the approach taken by the event API.

Update the specification accordingly.


kasperbirch1 (approved):
Nice! It works locally and I have added change the url in https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1169

PR #1106, opened 2024-05-07
Allow mediator to control publishing for all accessible content types

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-685

#### Description

The override node options supports configuring permission in general and per node type. Configuring per node type adds a lot of permissions to an already crowded permission page.

We just want to allow mediators to control publishing for all content types where they have create and/or edit access.

We can achieve this by allowing them to override published options for all content types and override node options to only use general permissions.

Mediators will not be able to (un)publish nodes they do not have access to (e.g. branches) but will be allowed to control this for articles, pages, events etc.

#### Screenshot of the result

Monosnap Extend | DPL CMS | Logged in 2024-05-07 11-09-00

Monosnap Override Node Options settings | DPL CMS | Logged in 2024-05-07 11-08-10

Monosnap Permissions | DPL CMS | Logged in 2024-05-07 11-09-51

Monosnap Indhold | DPL CMS | Logget ind — Privat browsing 2024-05-07 11-06-13

Monosnap Redigér Artikel Jesper Stein vinder Læsernes Bogpris for Rampen | DPL CMS | Logget ind — Privat browsing 2024-05-07 11-07-13


kasperbirch1 (approved):
Nice, I can see that this is a better solution. Thanks. 🙏

Should I just merge it?

PR #1109, opened 2024-05-07
Update install hook and add update hook [DDFLSBP-401]

clausbruun:
#### Description

After adding the Pathauto module, we need to fetch the node path before updating the settings.

#### Link to issue

Old issue: https://reload.atlassian.net/browse/DDFLSBP-401


Dresse (approved):
LGTM

PR #1110, opened 2024-05-07
Downgrade and pin Drush to 12.1.2

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-759

#### Description

During deployment of release 2024.8.2 we have seen errors in the form:

```
[warning] Undefined array key "language.en" ConfigImporter.php:353
[warning] Trying to access array offset on value of type null ConfigImporter.php:353 
TypeError: array_diff(): Argument #2 must be of type array, null given in array_diff()
```

This seems to be related to an issue introduced in Drush 12.1.3 https://github.com/drush-ops/drush/issues/5838.
The fix has been merged in the 13.x branch. There is no patch available which can be applied cleanly. We do not want to upgrade to a dev release so instead we downgrade to Drush 12.1.2 which according to the isse is the last version before the bug was introduced.

The fact that this works can be confirmed by recreating the error:

- Download a database from before the problem occurred
- Check out version 2024.8.2 locally
- Run drush deploy which is similar to what occurs when deploying in production

The error occurs locally with the current version of Drush, 12.4.3. It does not occur when downgrading to 12.1.2

We should upgrade to Drush 13 once this is released.

We should not be able to see anything in practice in this PR. The main point of running these is to ensure that nothing fails now that we downgrade to an older version of Drush.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -156,7 +156,7 @@
         "drupal/view_unpublished": "^1.1",
         "drupal/views_infinite_scroll": "^2.0",
         "drupal/webform": "^6.2",
-        "drush/drush": "^12",
+        "drush/drush": "12.1.2",
         "mnsami/composer-custom-directory-installer": "^2.0",
         "myclabs/php-enum": "^1.8",
         "npm-asset/select2": "^4.0",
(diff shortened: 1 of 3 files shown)


rasben (approved):

kasperg:
In the end @spaceo and I have decided not to do this. Downgrading Drush increases the frequency of errors during Drupal install to a level that will become problematic. We have to look for other solutions like upgrading to Drush 13.x.

PR #1112, opened 2024-05-08
Prevent failing event deployment hook on new sites

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-760

#### Description

Apparently existing deployment hooks are run on both existing and new installations. This causes problems with an event deployment hook where the entity query refers to a field which does not exists on new installations:

```
[notice] Deploy hook started: dpl_event_deploy_port_description_fields
[error]  'field_event_description' not found
[error]  Deploy hook failed: dpl_event_deploy_port_description_fields
```

Update the deployment hook to load all entities but only affect entities which actually have a nonempty value the source field.

This way the deploy hook can run on new sites but since they do not have the old field - or it is empty because it is hidden - it will not do anything.

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -10,6 +10,7 @@ tasks:
     - run:
         name: If drupal is not installed
         command: |
+          set -e
           if tables=$(drush sqlq "show tables like 'node';") && [ -z "$tables" ]; then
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #1113, opened 2024-05-08
Make multiline scripts fail if single command fails

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-760

#### Description

We have seen deployments where a command which is a part of the deployment fails but the deployment continues.

Add set -e to all multiline scripts to make them fail the first time a command fails.

This should be added to the .lagoon.yml file for all sites also.

This depends on #1112 as the deployments hook currently fail. Without this change they will continue anyway.

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -10,6 +10,7 @@ tasks:
     - run:
         name: If drupal is not installed
         command: |
+          set -e
           if tables=$(drush sqlq "show tables like 'node';") && [ -z "$tables" ]; then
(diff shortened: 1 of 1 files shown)


spaceo (approved):
This fixes the problem in dpl-cms, but what about dpl-platform?

kasperg:
@hypesystem has promised to ensure that the change is propagated to the individual sites.

hypesystem (approved):

PR #1115, opened 2024-05-08
Add translations and test to opening hours repeat edit delete

kasperbirch1:
#### Link to issue
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1161
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1169

#### Description

This pull request adds tests and translations for the new modal for removal / editing a single opening hour instance or all feature instances.

### DIFF
--- a/cypress/e2e/opening-hours-editor.cy.ts
+++ b/cypress/e2e/opening-hours-editor.cy.ts
@@ -10,11 +10,24 @@ enum OpeningHourCategories {
 
 type TimeString = `${number}:${number}`;
 
-type OpeningHourFormType = {
-  openingHourCategory: OpeningHourCategories;
+type TimeDurationType = {
   start: TimeString;
   end: TimeString;
+};
+
+type PartialTimeDurationType = {
+  start?: TimeString;
+  end?: TimeString;
+};
+
+type OpeningHourFormType = {
+  openingHourCategory: OpeningHourCategories;
+  timeDuration: TimeDurationType;
   endDate?: string;
+};
+
+type PartialOpeningHourFormType = Omit<OpeningHourFormType, "timeDuration"> & {
+  timeDuration?: PartialTimeDurationType;
 };
 
 const oneMonthFromToday = () =>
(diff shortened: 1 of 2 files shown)


cypress/e2e/opening-hours-editor.cy.ts
```diff
@@ -8,10 +8,12 @@ enum OpeningHourCategories {
...
 }
 
+type TimeString = `${number}:${number}`;
+
 type OpeningHourFormType = {
   openingHourCategory: OpeningHourCategories;
-  start: `${number}:${number}`;
-  end: `${number}:${number}`;
+  start: TimeString;
+  end: TimeString;
```
kasperg, line 16: :+1: Nice to add a proper type for this.

You could consider creation a new `TimeDuration` type which is a pair of time strings with a start and an end. The values seems to be passed around in pairs anyway.

cypress/e2e/opening-hours-editor.cy.ts
```diff
@@ -140,50 +142,90 @@ const checkConfirmationDialog = ({
...
 
-const checkMultipleOpeningHoursExistAdmin = () => {
-  cy.get('tbody[role="presentation"]')
+const validateOpeningHoursChangesAdmin = ({
+  openingHourCategory,
+  originalStart,
+  originalEnd,
+  updatedStart,
+  updatedEnd,
+  editSeriesFromIndex,
```
kasperg, line 191: In general I find it very hard to grok what the parameter `editSeriesFromIndex` actually means. Perhaps it is just me.

Based on careful reading of the test consider refactoring the method to just check that you have the expected number of opening hour entries with the provided start and end times. Then the test can call this function twice and check that we have the expected number of original times and the expected number of updated times.

cypress/e2e/opening-hours-editor.cy.ts
```diff
@@ -366,4 +527,119 @@ describe("Opening hours editor", () => {
...
+      openingHourCategory,
+      editSeriesFromIndex,
+    }: EditRestOfOpeningHoursSeriesType = {
+      editSeriesFromIndex: 2,
+      openingHourCategory: OpeningHourCategories.SelfService,
+      originalStart: "10:00",
+      originalEnd: "16:00",
+      updatedStart: "09:00",
+      updatedEnd: "15:00",
+    };
```
kasperg, line 569: I find it quite hard to read this object creation/deconstruction that seems to be going on here.

Does it provide any value compared to just assigning values e.g. `const editSeriesFromIndex: int = 2` etc.? 

It also occurs a couple of other places in this PR.

cypress/e2e/opening-hours-editor.cy.ts
```diff
@@ -272,9 +338,48 @@ const updateOpeningHour = ({
...
+      openingHourCategory,
+      start,
+      end,
+    });
+    visitOpeningHoursPage();
+    // Because we use oneMonthFromToday as endDate we can check the four next weeks
+    for (let i = 0; i < 5; i++) {
+      validateOpeningHoursPage({ openingHourCategory, start, end });
+      cy.getBySel("opening-hours-next-week-button").click();
+    }
```
kasperg, line 381: :+1: As we are using the week view here this now makes sence.

kasperg (approved):
:+1: Nice

I have some code style suggestions and comments you may want to look into before merging.

PR #1119, opened 2024-05-09
Programmatically add translations for un-capitalized weekdays. DDFFOR…

rasben:
…M-622

This is particularly important for eventseries, where it currently says "Hver saturday" - DDF will not be able to translate it, until the actual word shows up, so instead, we do it for them.

https://reload.atlassian.net/browse/DDFFORM-622

-----

This was originally created as a PR here, but apparantly the branch name was too long for lagoon to be able to deploy, so i've re-created it here.

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1068

### DIFF
--- a/web/modules/custom/dpl_event/dpl_event.install
+++ b/web/modules/custom/dpl_event/dpl_event.install
@@ -6,14 +6,18 @@
  */
 
 use Drupal\Core\Field\BaseFieldDefinition;
+use Drupal\drupal_typed\DrupalTyped;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\locale\SourceString;
+use Drupal\locale\StringDatabaseStorage;
 
 /**
  * Implements hook_install().
  */
 function dpl_event_install(): void {
   _dpl_event_create_mock_title();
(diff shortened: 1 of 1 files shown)


rasben:
# Comments from other PR:

@kasperg :
> Normally [we use code scanning for handling translations of UI strings](https://danskernesdigitalebibliotek.github.io/dpl-docs/DPL-CMS/translation/#translation-system).
> 
> If you do it this way around you have to ensure that this does not get overridden by the current automations and processes we already have in place.
> 
> I cannot tell form the PR whether this is the case or not.

----

@rasben : 

> The problem is that until the system detects it, it wont show up in the translation system. So, for example, `thursday` doesnt even exist in the system at the moment as a source string, and that makes it impossible for DDF to translate.
> 
> Seeing as this is pretty unambigious translations (how would "Monday" ever be translated to other than "Mandag"?), then I dont worry too much about it overriding or being overriden

kasperg (approved):
This uses a different approach to translation that we would do normally.

In this situation where we know with good confidence that our translations are good and do not need further update this is OK.

I assume that you have tried to run this update hook and then import translations on your local environment and ensured that they are not overriden or cause problems with our normal translation procedure.

PR #1122, opened 2024-05-10
Related-content slider on articles and eventinstances. DDFFORM-394

rasben:
Place a dynamic 'related content' list on articles and eventinstances. Content displayed is based on the field values of the current page entity.

We go through several checks, until we find a result that shows more than 3 items:

- Only tags
- Tags OR Categories
- Tags, Categories OR events with same branch
- Upcoming events / Newest articles

If nothing is available, it wont display the slider. See more about the logic in the JIRA ticket:
https://reload.atlassian.net/browse/DDFFORM-394

### DIFF
--- a/web/modules/custom/dpl_react/dpl_react.module
+++ b/web/modules/custom/dpl_react/dpl_react.module
@@ -84,20 +84,6 @@ function dpl_react_cover_service_url(): string {
 
 /**
  * Implements hook_theme().
- *
- * @param mixed[] $existing
- *   An array of existing implementations that may be used
- *   for override purposes.
- * @param string $type
- *   Whether a theme, module, etc. is being processed.
- * @param string $theme
- *   The actual name of theme, module, etc. that is being being processed.
- * @param string $path
- *   The directory path of the theme or module,
- *   so that it doesn't need to be looked up.
- *
- * @return mixed[]
- *   An associative array of information about theme implementations.
  */
 function dpl_react_theme(
   array $existing,
--- a/web/modules/custom/dpl_related_content/dpl_related_content.info.yml
+++ b/web/modules/custom/dpl_related_content/dpl_related_content.info.yml
@@ -1,5 +1,5 @@
 name: DPL related content
 description: Showing related content as a slider on articles and eventinstances.
-package: DDB
+package: DPL
 type: module
 core_version_requirement: ^10
--- a/web/modules/custom/dpl_related_content/dpl_related_content.info.yml
+++ b/web/modules/custom/dpl_related_content/dpl_related_content.info.yml
@@ -1,0 +1,5 @@
+name: DPL related content
+description: Showing related content as a slider on articles and eventinstances.
+package: DDB
+type: module
+core_version_requirement: ^10
--- a/web/modules/custom/dpl_related_content/dpl_related_content.module
+++ b/web/modules/custom/dpl_related_content/dpl_related_content.module
@@ -1,0 +1,292 @@
+<?php
+
+use Drupal\Core\Database\Database;
+use Drupal\Core\Datetime\DrupalDateTime;
+use Drupal\Core\Entity\FieldableEntityInterface;
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
+use Drupal\node\Entity\Node;
+use Drupal\node\NodeInterface;
+use Drupal\recurring_events\Entity\EventInstance;
+
+/**
+ * Get terms from field as an array of IDs.
+ *
+ * @return array<int>
+ *   Term IDs.
+ */
+function _dpl_related_content_get_term_ids(FieldableEntityInterface $entity, string $field_name): array {
+  if (!$entity->hasField($field_name)) {
+    return [];
+  }
+
+  $terms = $entity->get($field_name)->getValue();
+  return array_column($terms, 'target_id');
+
+}
+
+/**
+ * Get matching eventinstance IDs.
+ *
+ * Allows for passing along various term IDs, that we look for in an OR group.
+ * Notice - we only look for the term values on series level, ignoring
+ * inheritance overriding in favor of a simpler codebase.
+ *
+ * @param string|null $current_uuid
+ *   A possible entity UUID, that will not get included in results.
+ * @param array<int> $tags
+ *   Tag term IDs, to look for.
+ * @param array<int> $categories
+ *   Category term IDs, to look for.
+ * @param array<int> $branches
+ *   Branch term IDs, to look for.
+ *
+ * @return array<int|string>
+ *   Matching eventinstance IDs
+ *
+ * @throws \Exception
+ */
+function _dpl_related_content_get_eventinstance_ids(?string $current_uuid, array $tags = [], array $categories = [], array $branches = []): array {
+  $date = new DrupalDateTime('today');
+  $date->setTimezone(new \DateTimezone(DateTimeItemInterface::STORAGE_TIMEZONE));
+  $formatted_date = $date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT);
+
+  $connection = Database::getConnection();
+
+  // Prepare a subquery for eventseries IDs based on terms.
+  $subquery = $connection->select('eventseries', 'es');
+  $subquery->fields('es', ['id']);
+
+  // Join the term field tables.
+  // @todo this only looks at the values on the eventseries.
+  // Ideally, we'd want to also look if values exist on the instance level.
+  $subquery->leftJoin('eventseries__field_tags', 'es_tags', 'es.id = es_tags.entity_id');
+  $subquery->leftJoin('eventseries__field_categories', 'es_cats', 'es.id = es_cats.entity_id');
+  $subquery->leftJoin('eventseries__field_branch', 'es_bra', 'es.id = es_bra.entity_id');
+
+  if (!empty($tags) || !empty($categories) || !empty($branches)) {
+    $or_group = $subquery->orConditionGroup();
+
+    // To avoid errors, related to the OR GROUP only containing one condition,
+    // we'll add a fake condition to fill out a possible empty space.
+    $or_group->condition('title', 'ALWAYS_FALSE');
+
+    if (!empty($tags)) {
+      $or_group->condition('es_tags.field_tags_target_id', $tags, 'IN');
+    }
+
+    if (!empty($categories)) {
+      $or_group->condition('es_cats.field_categories_target_id', $categories, 'IN');
+    }
+
+    if (!empty($branches)) {
+      $or_group->condition('es_bra.field_branch_target_id', $branches, 'IN');
+    }
+
+    $subquery->condition($or_group);
+  }
+
+  $subquery->distinct(TRUE);
+
+  // Main query to select eventinstance ids, joining with
+  // eventinstance_field_data for condition fields.
+  $query = $connection->select('eventinstance_field_data', 'eid');
+  $query->join('eventinstance', 'ei', 'ei.id = eid.id');
+  $query->addField('eid', 'id', 'eventinstance_id');
+
+  if (!empty($tags) || !empty($categories) || !empty($branches)) {
+    // Use the subquery to filter by eventseries_id.
+    $query->condition('eid.eventseries_id', $subquery, 'IN');
+  }
+
+  $query->condition('ei.uuid', $current_uuid, '<>');
+
+  // We only want events in the future (e.g. - active events).
+  $query->condition('eid.date__value', $formatted_date, '>=');
+  $query->orderBy('eid.date__value', 'ASC');
+  $query->range(0, 16);
+
+  // Add a GROUP BY clause to make results distinct by eventseries_id.
+  $query->groupBy('eid.eventseries_id');
+
+  // Execute the query and return ids.
+  $result = $query->execute();
+
+  return $result?->fetchCol() ?? [];
+}
+
+/**
+ * Get matching node IDs.
+ *
+ * Allows for passing along various term IDs, that we look for in an OR group.
+ *
+ * @param string|null $current_uuid
+ *   A possible entity UUID, that will not get included in results.
+ * @param array<int> $tags
+ *   Tag term IDs, to look for.
+ * @param array<int> $categories
+ *   Category term IDs, to look for.
+ *
+ * @return array<int|string>
+ *   Matching article node IDs
+ */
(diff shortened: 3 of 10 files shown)


web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+  $subquery->leftJoin('eventseries__field_tags', 'es_tags', 'es.id = es_tags.entity_id');
+  $subquery->leftJoin('eventseries__field_categories', 'es_cats', 'es.id = es_cats.entity_id');
+  $subquery->leftJoin('eventseries__field_branch', 'es_bra', 'es.id = es_bra.entity_id');
+
+  if (!empty($tags) || !empty($categories) || !empty($branches)) {
+    $or_group = $subquery->orConditionGroup();
+
+    // To avoid errors, related to the OR GROUP only containing one condition,
+    // we'll add a fake condition to fill out a possible empty space.
+    $or_group->condition('title', 'ALWAYS_FALSE');
```
kasperg, line 71: :+1: Appreciate this comment.

I am a bit supprised if errors occur for condition groups with only one condition.

web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+    // Use the subquery to filter by eventseries_id.
+    $query->condition('eid.eventseries_id', $subquery, 'IN');
+  }
+
+  $query->condition('ei.uuid', $current_uuid, '<>');
+
+  // We only want events in the future (e.g. - active events).
+  $query->condition('eid.date__value', $formatted_date, '>=');
+  $query->orderBy('eid.date__value', 'ASC');
+  $query->range(0, 16);
```
kasperg, line 106: Please add a comment explaining why this range is necessary.

In general I would have expected us to include all the matching ids. If we wanted to do any limiting we could do it later on in the prcoess.

web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+
+/**
+ * Get matching eventinstance IDs.
+ *
+ * Allows for passing along various term IDs, that we look for in an OR group.
+ * Notice - we only look for the term values on series level, ignoring
+ * inheritance overriding in favor of a simpler codebase.
+ *
+ * @param string|null $current_uuid
+ *   A possible entity UUID, that will not get included in results.
```
kasperg, line 35: Consinder changing this here and elsewhere to ids instead of.

I wonder why this is a UUID when all the other values we pass around here are ids. This also goes elsewhere in this PR.

web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+ *   Matching eventinstance IDs
+ *
+ * @throws \Exception
+ */
+function _dpl_related_content_get_eventinstance_ids(?string $current_uuid, array $tags = [], array $categories = [], array $branches = []): array {
+  $date = new DrupalDateTime('today');
+  $date->setTimezone(new \DateTimezone(DateTimeItemInterface::STORAGE_TIMEZONE));
+  $formatted_date = $date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT);
+
+  $connection = Database::getConnection();
```
kasperg, line 53: I would expect some kind of comment about why we use direct queries here instead of an entity query.

Perhaps it is because of the subquery?

web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+ *   Category term IDs, to look for.
+ *
+ * @return array<int|string>
+ *   Matching article node IDs
+ */
+function _dpl_related_content_get_article_ids(?string $current_uuid, array $tags = [], array $categories = []): array {
+  $query = \Drupal::entityQuery('node')->accessCheck(TRUE)
+    ->condition('type', 'article')
+    ->condition('uuid', $current_uuid, '<>')
+    ->sort('created', 'DESC')
```
kasperg, line 136: This does sorting but there is no mention about if/why this is important/necessary.

web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+    if (isset($events[$i])) {
+      $content[] = $event_view_builder->view($events[$i], 'card');
+    }
+
+    if (isset($articles[$i])) {
+      $content[] = $node_view_builder->view($articles[$i], 'card');
+    }
+  }
+
+  return $content;
```
kasperg, line 255: Please consider the cache information here.

This does not seem to have any type of cache information available apart from what is provided by the individual renders. I would it to be dependent of a lot of different information - available articles, events, tags, the current time etc.

Perhaps it would make more sense to only cache this based on the time.

web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+    return;
+  }
+
+  try {
+    $variables['related_content'] = _dpl_related_content_get_content($entity);
+  }
+  catch (\Exception $e) {
+    \Drupal::logger('dpl_related_content')->error('Could not render related content slider.');
+  }
+}
```
kasperg, line 292: Please include information from the error being thrown.

Otherwise it is impossible to know what has failed.

web/modules/custom/dpl_related_content/dpl_related_content.info.yml
```diff
@@ -0,0 +1,5 @@
+name: DPL related content
+description: Showing related content as a slider on articles and eventinstances.
+package: DDB
```
kasperg, line 3:
```suggestion
package: DPL
```

web/themes/custom/novel/templates/layout/page.html.twig
```diff
@@ -72,6 +72,16 @@
         } only %}
       </div>
     {% endif %}
+
+    {% if related_content %}
+      <div class="mt-64">
+        {% include '@novel/components/slider.html.twig'
+          with {
+          'title': 'Related content'|trans,
```
kasperg, line 80: Please remember to add a context to the translations.

web/modules/custom/dpl_related_content/dpl_related_content.module
```diff
@@ -0,0 +1,292 @@
...
+use Drupal\node\NodeInterface;
+use Drupal\recurring_events\Entity\EventInstance;
+
+/**
+ * Get terms from field as an array of IDs.
+ *
+ * @return array<int>
+ *   Term IDs.
+ */
+function _dpl_related_content_get_term_ids(FieldableEntityInterface $entity, string $field_name): array {
```
kasperg, line 17: This has a lot of "private" functions at the top of the module" while the meat, `dpl_related_content_preprocess_page()`, is at the bottom.  That makes it a bit harder to read as you start with the details but do not know where you will end.

kasperg (changes requested):
:+1: This is somewhat gnarly, but I appreciate a lot of the comments here.

I think there are some things here needs to be addressed before this can be merged.

In general: This is ~250 LOC in a `.module` file. I would have liked to see it more contained. One approach to do that which I think fits nicely here is in a Drupal Block.

web/modules/custom/dpl_related_content/dpl_related_content.module
rasben: Me too.

web/modules/custom/dpl_related_content/dpl_related_content.module
rasben: I'll add a comment in the code why - here, for your convienience:

```
   *   We use UUID instead of IDs, as UUID will be unique across entity types,
   *   and means we don't need to worry about sending a node UUID along to a
   *   event query.

File: web/modules/custom/dpl_related_content/dpl_related_content.module (hunk 4 above)
Comment (rasben) [reply web/modules/custom/dpl_related_content/dpl_related_content.module:53]: Indeed because of the subquery. I'll add a comment about it.

File: web/modules/custom/dpl_related_content/dpl_related_content.module (hunk 2 above)
Comment (rasben) [reply web/modules/custom/dpl_related_content/dpl_related_content.module:106]: There's no point in pulling more than what will be the maximum allowed in the slider, but, i'll add documentation and a variable to make it clearer.

File: web/modules/custom/dpl_react/dpl_react.module [hunk 11]
```diff
@@ -84,20 +84,6 @@ function dpl_react_cover_service_url(): string {
...
- * @param string $type
- *   Whether a theme, module, etc. is being processed.
- * @param string $theme
- *   The actual name of theme, module, etc. that is being being processed.
- * @param string $path
- *   The directory path of the theme or module,
- *   so that it doesn't need to be looked up.
- *
- * @return mixed[]
- *   An associative array of information about theme implementations.
```
kasperg, line 100: :+1: for cleaning this up in relation to updating ignore rules.

I think `hook_theme()` is a relevant case for ignoring.

web/modules/custom/dpl_related_content/src/Services/RelatedContent.php
```diff
@@ -0,0 +1,371 @@
...
+
+  /**
+   * How many items we max display in the slider.
+   */
+  private int $maxItems = 16;
+
+  /**
+   * The field on nodes, to sort by.
+   */
+  private string $nodeSortField = 'created';
```
kasperg, line 36: Consider making these class constants if they are not default values not intented to be changed.

web/modules/custom/dpl_related_content/src/Services/RelatedContent.php
```diff
@@ -0,0 +1,371 @@
...
+
+  /**
+   * The node bundles that should show up in the results.
+   *
+   * @var array|string[]
```
kasperg, line 41:
```suggestion
   * @var string[]
```

If you only want to accept an array of strings[] I think it makes sense to declare this. Otherwise you accept an array of whatever.

web/modules/custom/dpl_related_content/src/Services/RelatedContent.php
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container): static {
+    return new static(
+      $container->get('entity_type.manager'),
+    );
+  }
```
kasperg, line 62: Please delete this function.

It is not used anywhere. In your `services.yml` you declare what is passed to the constructor. This function is only relevant for classes that implement [`ContainerInjectionInterface`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21DependencyInjection%21ContainerInjectionInterface.php/interface/ContainerInjectionInterface/9).

web/modules/custom/dpl_related_content/src/Services/RelatedContent.php
```diff
@@ -0,0 +1,371 @@
...
+      '#items' => $content,
+      // The results should be cached, but, they have a lot of dependencies -
+      // even depending on the current time (finding future events).
+      // The individual peices of content are cached themselves, so for the full
+      // result list, we'll add an easily-invalidated cache, with a bunch of
+      // cache tags, and even a time-based cache.
+      '#cache' => [
+        // Max age of 12 hours.
+        '#max-age' => (60 * 60 * 12),
+        '#tags' => ['eventinstance_list', 'eventseries_list', 'node_list'],
```
kasperg, line 195: :+1: The cache tags here makes sense.

The max age is set to 12 hours. This might be good and it might not.

As the values here depend on the date I suggest you instead add a cache context with the date. This will automatically invalidate entries when the date changes.

web/modules/custom/dpl_related_content/src/Services/RelatedContent.php
```diff
@@ -0,0 +1,371 @@
...
+   */
+  private function getEventInstanceIds(?string $excluded_uuid = NULL, array $tags = [], array $categories = [], array $branches = []): array {
+    $date = new DrupalDateTime('today');
+    $date->setTimezone(new \DateTimezone(DateTimeItemInterface::STORAGE_TIMEZONE));
+    $formatted_date = $date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT);
+
+    // Ideally, we'd use EntityQueries instead of a direct DB connection, but
+    // because we need to do some pretty complicated JOINs and subqueries, it
+    // is not an option.
+    $connection = Database::getConnection();
```
kasperg, line 285: Please inject the connection in the constructor.

It is available as a service.

web/modules/custom/dpl_related_content/templates/dpl-related-content-slider.html.twig
```diff
@@ -0,0 +1,5 @@
+{{ title }}
+
+{% for item in items %}
+  {{ item }}
+{% endfor %}
```
kasperg, line 5: Consider if this template is necessary.

If it is a dummy template then could we just delete it?

web/modules/custom/dpl_related_content/dpl_related_content.module
kasperg: As far as I can tell has been changed to a property but I do not see an explaination.

With the sorting being `created`, `DESC` then we end up with the latest content first. I guess that is the priotization.

kasperg (approved):
:+1: I appreciate the restructure.

There are a few things that still needs a second look, but you can address them on your own before you merge.

web/modules/custom/dpl_related_content/src/Services/RelatedContent.php
rasben: In a future PR, they will be overwritten (and set as public), so I'll keep them as properties.

web/modules/custom/dpl_related_content/templates/dpl-related-content-slider.html.twig
rasben: It is necessary; If the template doesnt exist, Drupal loses its mind.
But yeah, this markup will never be displayed - it is overwritten by `novel`

web/modules/custom/dpl_related_content/src/Services/RelatedContent.php
rasben: @kasperg 
>I suggest you instead add a cache context with the date

Isnt that the same as `max-age`? What context do you have in mind?

rasben:
Closing this pull request, in favor of another, larger PR that was built and reviewed ontop of this:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1157

PR #1126, opened 2024-05-10
EventInstance teaser fields should respect inheritance.

rasben:
The teaser text and teaser image should be taken from the eventseries if possible.


LasseStaus (approved):
LGTM 🚀

PR #1128, opened 2024-05-10
Allow library admins to create 'external user' roles.

rasben:
This is necessary for creating roles for third-party event companies that need access to the event-API, such as Place2Book.


kasperg (approved):
:+1: LGTM

PR #1133, opened 2024-05-13
Restore backup from Lagoon locally

kasperg:
#### Description

As a part of working with live sites it can be helpful to duplicate the state to a local development environments.

This adds three tasks which support this:

1. Restoring database from backup
2. Restoring files from backup
3. Restoring both files and database from backup

The current implementations works - for me at least - but I have some questions that I would like to get feedback on before moving this along.

If we choose to add this we should update our documentation accordingly.

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -99,7 +99,7 @@ Replace files locally:
 2. Start a local environment using `task dev:reset`
 3. Restore the filesš by running `task dev:restore:files`
 
-## Get a specific release of dpl-react - without using composer install
+### Get a specific release of dpl-react - without using composer install
 
 In a development context it is not very handy only
 to be able to get the latest version of the main branch of dpl-react.
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -49,7 +49,23 @@ temporarily through a task:
 4. When you are finished, hit `enter` in the terminal where you enabled XDebug.
    This will disable XDebug
 
-### Copy database from Lagoon environment to local setup
+### Download database and files from Lagoon
+
+#### Retrieve the latest backup of database and files from Lagoon
+
+Prerequisites:
+
+* [A connected Lagoon CLI](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/docs/runbooks/connecting-the-lagoon-cli.md)
+* [`jq`](https://jqlang.github.io/jq/) installed locally
+
+Run the following command to retrieve the latest backup of database and files
+from a Lagoon project:
+
+```bash
+LAGOON_PROJECT=<lagoon-project-name> task lagoon:backup:restore
+```
+
+#### Copy a specific database snapshot from Lagoon environment to local setup
 
 Prerequisites:
 
@@ -68,7 +84,7 @@ database, not any files from the site.
 3. Start a local environment using `task dev:reset`
 4. Import the database by running `task dev:restore:database`
 
-### Copy files from Lagoon environment to local setup
+#### Copy a specific snapshot of files from Lagoon environment to local setup
 
 Prerequisites:
 
--- a/task/Taskfile.lagoon.yml
+++ b/task/Taskfile.lagoon.yml
@@ -20,16 +20,8 @@ tasks:
     vars:
       LAGOON_ENVIRONMENT: main
     cmds:
-      - |
-        BACKUPS=$(lagoon list backups -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --output-json);
-        BACKUP_ID=$(echo $BACKUPS | jq -r ".data[] | select(.source == \"mariadb\") | .backupid" | head -n 1);
-        BACKUP_URL="Error";
-        while [[ $BACKUP_URL == *"Error"* ]]; do
-          $(lagoon retrieve backup -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --backup-id ${BACKUP_ID} --force) 2>/dev/null || true;
-          BACKUP_URL=$(lagoon get backup -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --backup-id ${BACKUP_ID});
-        done;
-        curl -o ./restore/database/${LAGOON_PROJECT}-backup.tar.gz ${BACKUP_URL};
-        cd ./restore/database/ && tar -xvzf ${LAGOON_PROJECT}-backup.tar.gz;
+      - ./dev-scripts/lagoon-get-backup.sh ${LAGOON_PROJECT} mariadb {{.DIR_RESTORE_DATABASE}}/${LAGOON_PROJECT}-backup.tar.gz;
+      -  cd ./restore/database/ && tar -xvzf ${LAGOON_PROJECT}-backup.tar.gz;
       - task dev:restore:database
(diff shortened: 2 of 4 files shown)


task/Taskfile.lagoon.yml
```diff
@@ -0,0 +1,100 @@
...
+    desc: Restore the latest backup of files and database from a Lagoon project
+    cmds:
+      - task: backup:restore:database
+      - task: backup:restore:files
+
+  backup:restore:database:
+    desc: Restore the latest backup of the database from a Lagoon project
+    deps: [deps:lagoon, deps:jq, login]
+    vars:
+      LAGOON_ENVIRONMENT: main
```
kasperg, line 21: The environment is hardcoded here. Is it valuable to make it configurable?

task/Taskfile.lagoon.yml
```diff
@@ -0,0 +1,100 @@
...
+    cmds:
+      - |
+        BACKUPS=$(lagoon list backups -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --output-json);
+        BACKUP_ID=$(echo $BACKUPS | jq -r ".data[] | select(.source == \"mariadb\") | .backupid" | head -n 1);
+        BACKUP_URL="Error";
+        while [[ $BACKUP_URL == *"Error"* ]]; do
+          $(lagoon retrieve backup -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --backup-id ${BACKUP_ID} --force) 2>/dev/null || true;
+          BACKUP_URL=$(lagoon get backup -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --backup-id ${BACKUP_ID});
+        done;
+        curl -o ./restore/database/${LAGOON_PROJECT}-backup.tar.gz ${BACKUP_URL};
```
kasperg, line 31: There is some complexity here and duplication between this task and `backup:restore:files`. Do you think it is necessary/valuable to move this to a proper `.sh` script in `/devscripts`?

task/Taskfile.lagoon.yml
```diff
@@ -0,0 +1,100 @@
...
+        cd ./restore/database/ && tar -xvzf ${LAGOON_PROJECT}-backup.tar.gz;
+      - task dev:restore:database
```
kasperg, line 33: Currently `task dev:restore:database` requires there to be only one `.sql` file in `./restore/database`. Should we do something to address this e.g. delete any existing files before running the task?

task/Taskfile.lagoon.yml
```diff
@@ -0,0 +1,100 @@
+# This is a go-task file solely for Lagoon integration.
+# It can be used independently but is meant to be used
+# in combination with the go-task file in the root of the project.
```
kasperg, line 3: Stating that this can be used independently is not correct. This file uses tasks provided by the root `Taskfile.yml` where it also included. The fact that I split it out is primarily a matter of keeping taskfiles somewhat smaller and more focused.

Is there a better way to do this?

task/Taskfile.lagoon.yml
```diff
@@ -0,0 +1,100 @@
...
+    vars:
+      LAGOON_ENVIRONMENT: main
+    cmds:
+      - |
+        BACKUPS=$(lagoon list backups -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --output-json);
+        BACKUP_ID=$(echo $BACKUPS | jq -r ".data[] | select(.source == \"mariadb\") | .backupid" | head -n 1);
+        BACKUP_URL="Error";
+        while [[ $BACKUP_URL == *"Error"* ]]; do
+          $(lagoon retrieve backup -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --backup-id ${BACKUP_ID} --force) 2>/dev/null || true;
+          BACKUP_URL=$(lagoon get backup -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --backup-id ${BACKUP_ID});
```
kasperg, line 29: `lagoon get backup` will retur an error message á la.`Error: Backup is not restored` if the backup is not yet available.

task/Taskfile.lagoon.yml
```diff
@@ -0,0 +1,100 @@
...
+
+  backup:restore:database:
+    desc: Restore the latest backup of the database from a Lagoon project
+    deps: [deps:lagoon, deps:jq, login]
+    vars:
+      LAGOON_ENVIRONMENT: main
+    cmds:
+      - |
+        BACKUPS=$(lagoon list backups -p ${LAGOON_PROJECT} -e {{.LAGOON_ENVIRONMENT}} --output-json);
+        BACKUP_ID=$(echo $BACKUPS | jq -r ".data[] | select(.source == \"mariadb\") | .backupid" | head -n 1);
```
kasperg, line 25: This will always restore the latest backup. Do we need to make this more configurable e.g. by providing a timestamp or the like?

task/Taskfile.lagoon.yml
spaceo: I think the way it has been done here is fine.
In my opinion go-task has limitations working with included Taskfiles and task cmd scopes.
I like the idea of separating chunks of functionality into separate files.
I have been struggling  with task calling other tasks belonging to another file as well.

task/Taskfile.lagoon.yml
spaceo: So you made a workaround?

task/Taskfile.lagoon.yml
spaceo: Makes sense I think :)

task/Taskfile.lagoon.yml
kasperg: I did not do that for now.

task/Taskfile.lagoon.yml
kasperg: :+1: I have updated the PR with this change.

task/Taskfile.lagoon.yml
kasperg: I did not change this. For now the task ends with a nice explantory error message which makes it pretty obvious to the developer what they need to do to proceed.

task/Taskfile.lagoon.yml
kasperg: :+1: I updated the description a bit.

task/Taskfile.lagoon.yml
kasperg: I did not opt for implementing this for now. If needed it can be added later on.

docs/local-development.md
```diff
@@ -49,7 +49,23 @@ temporarily through a task:
...
+* [`jq`](https://jqlang.github.io/jq/) installed locally
+
+Run the following command to retrieve the latest backup of database and files
+from a Lagoon project:
+
+```bash
+LAGOON_PROJECT=<lagoon-project-name> task lagoon:backup:restore
+```
+
+#### Copy a specific database snapshot from Lagoon environment to local setup
```
Dresse, line 68: Nice with updated documentation 👍

Dresse, line 68: Nice with updated documentation 👍

Dresse (approved):
LGTM

spaceo (approved):
LGTM 👍

PR #1136, opened 2024-05-13
DDFLSBP-568 - Remove login tracking

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-568

#### Description

This PR is removing tracking functionality previously added in [this PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/942).
We have decided to remove it for now as it is breaking other tracking functionality. We will revisit this at a later point when dealing with [this JIRA task DDFLSBP-634](https://reload.atlassian.net/browse/DDFLSBP-634).

### DIFF
--- a/cypress/e2e/mapp-tracking.cy.ts
+++ b/cypress/e2e/mapp-tracking.cy.ts
@@ -13,7 +13,6 @@ describe("Mapp Tracking", () => {
       .parent()
       .get('[value="Save configuration"]')
       .click();
-    cy.drupalLogout();
     cy.visit("/");
     cy.getRequestCount({
       urlPathPattern: `^/resp/api/get/${customerId}.*`,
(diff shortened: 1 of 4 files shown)


spaceo (approved):
This is a bit rubberstamp approval 😅
But it looks ok.
And you are sure that this change is needed as well?:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1136/files#diff-63df581bfd6b0b85955bcce3de8b5a1208944a5b74af53cc9ad2b233856bae30L16

Dresse:
> This is a bit rubberstamp approval 😅 But it looks ok. And you are sure that this change is needed as well?: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1136/files#diff-63df581bfd6b0b85955bcce3de8b5a1208944a5b74af53cc9ad2b233856bae30L16

You can check the [PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/942) where I added the code. 
This is also where I updated the cypress test as it was failing after adding the tracking code. 
To be on the save side, I just reverted everything back to how it was before 😄

PR #1139, opened 2024-05-13
Run shellcheck on dev scripts

kasperg:
#### Description

As the complexity of these scripts may be increasing we might as well shellcheck them like we do in other parts of the project.


spaceo (approved):
LGTM 👍

PR #1145, opened 2024-05-14
Add links to "/articles" in `card_grid`´s

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-525

#### Description
This pull request adds links to "/articles" in card_grid_automatic.

I have added a new field, field_card_grid_link, with default values. I've incorporated this field into the view and included the CSS class name. Additionally, I added a template to ensure that only the view is rendered. I also implemented dpl_filter_paragraphs_resave_card_grid_automatic to guarantee link presence in existing paragraphs.

#### Screenshot of the result
 (Links seem to be placed too far to the right because we only have 3 articles in our default content) 
image

#### Test
https://varnish.pr-1145.dpl-cms.dplplat01.dpl.reload.dk/test-card-grid-link


config/sync/core.entity_form_display.paragraph.card_grid_automatic.default.yml
```diff
@@ -34,6 +36,14 @@ targetEntityType: paragraph
 bundle: card_grid_automatic
 mode: default
 content:
+  field_card_grid_link:
```
rasben, line 39: Please make the field name more generic (or use an existing one)

e.g. `field_link` / `field_more_link` or similar.

That way, we can re-use logic in the future.

rasben (changes requested):

web/themes/custom/novel/templates/fields/field--field-more-link.html.twig
```diff
@@ -0,0 +1,5 @@
+{% for item in items %}
+  {% set url = item.content['#url'] %}
+  {% set title = item.content['#title'] %}
```
rasben, line 3: There's no point in saving these as variables, when you only use them once

web/themes/custom/novel/templates/fields/field--field-more-link.html.twig
```diff
@@ -0,0 +1,5 @@
+{% for item in items %}
```
rasben, line 1: The file is `field--field-more-link.html.twig`, but the file includes references to `card-grid` - can you target the paragraph specificially?
Like field--field--more--link--paragraph--blabla or something?

rasben (changes requested):

rasben (approved):

rasben:
@kasperbirch1 I have updated it. I'll let you test it :)

PR #1151, opened 2024-05-16
Migrate admins

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-768

#### Description

When some sites where initially created, library organizations got
access to the initial administrative Drupal user, user 1. This user
has special privileges which are not intended for the library
organizations.

To address this situation we create a new administrative user to be
used by the libraries instead.

To make this transfer as smooth as possible we transfer as much data
as possible from the existing user 1 to to the new user. This includes
user data and entity ownership.

It is not easy to transfer content locks so instead we reset all locks
that user 1 may have.

###### Output during update hook

```
> >  [notice] Update started: dpl_admin_update_10000
> >  [warning] Missing file with ID 1. ImageItem.php:331
> >  [notice] Retrieved password hash for user 1
> > Reset username and password for user 1
> > Removed roles local_administrator, editor, mediator for user 1
> > Created new user 9 with local administrator role
> > Transferred password hash from user 1 to new user 9
> > Updated owner for 78 file entities
> > Updated owner for 59 media entities
> > Updated owner for 96 node entities
> > Updated owner for 3 eventseries entities
> > Updated owner for 3 eventinstance entities
> > Deleted existing sessions for user 1
> > Released all content locks for user 1
> >  [notice] Update completed: dpl_admin_update_10000
> >  [notice] Message: Successfully updated 1 instance statuses. Skipped 0 instances due to status
> > or workflow mismatch with series.
```

###### User 1 before

billede 

###### User 1 after

billede

###### New local admin after

billede

###### User list before

billede

###### User list after

billede

###### Content list before

billede

###### Content list after

billede

(Note that all content previously owned by admin will be updated)

I have tested transfer of passwords using the following procedure:

1. Restore a site from backup
2. Login locally as user 1 using `drush uli` 
3. Note the user name for user 1
4. Change the password of user 1 to `test`
5. Log out
6. Run the update hooks using `drush deploy`
7. Log in using the original user name and new password for user 1 
8. See that I am logged in as the new user

It is not possible to see the changes to event, media or file ownership as this is not listed in the UI of these.

### DIFF
--- a/web/modules/custom/dpl_admin/dpl_admin.install
+++ b/web/modules/custom/dpl_admin/dpl_admin.install
@@ -11,6 +11,7 @@ use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Entity\Query\QueryException;
 use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Language\LanguageManagerInterface;
 use Drupal\Core\Password\DefaultPasswordGenerator;
 use Drupal\dpl_admin\Services\VersionHelper;
 use Drupal\drupal_typed\DrupalTyped;
@@ -87,7 +88,7 @@ function dpl_admin_update_10000(array &$sandbox) : string {
   $password_generator = DrupalTyped::service(DefaultPasswordGenerator::class, 'password_generator');
   $user_1->setPassword($password_generator->generate());
 
-  $language_manager = DrupalTyped::service(\Drupal\Core\Language\LanguageManagerInterface::class, 'language_manager');
+  $language_manager = DrupalTyped::service(LanguageManagerInterface::class, 'language_manager');
   $user_1->set('preferred_admin_langcode', $language_manager->getDefaultLanguage()->getId());
 
   $user_1->save();
--- a/web/modules/custom/dpl_admin/dpl_admin.install
+++ b/web/modules/custom/dpl_admin/dpl_admin.install
@@ -17,6 +17,7 @@ use Drupal\dpl_admin\Services\VersionHelper;
 use Drupal\drupal_typed\DrupalTyped;
 use Drupal\user\Entity\User;
 use Drupal\user\EntityOwnerInterface;
+use Drupal\user\UserInterface;
 
 /**
  * Implements hook_requirements().
@@ -156,4 +157,18 @@ function dpl_admin_update_10000(array &$sandbox) : string {
   $feedback[] = "Released all content locks for user 1";
 
   return implode("\n", $feedback);
+}
+
+/**
+ * Migrate administrators to local administrators.
+ */
+function dpl_admin_update_10001(array &$sandbox): string {
+  $administrators = \Drupal::entityTypeManager()->getStorage('user')->loadByProperties(['roles' => 'administrator']);
+  $updates = array_map(function (UserInterface $administrator) {
+      $administrator->removeRole('administrator');
+      $administrator->addRole('local_administrator');
+      $administrator->save();
+      return "Converted user {$administrator->id()} from administrator to local administrator";
+  }, $administrators);
+  return implode("\n", $updates);
 }
--- a/web/modules/custom/dpl_admin/dpl_admin.install
+++ b/web/modules/custom/dpl_admin/dpl_admin.install
@@ -111,7 +111,7 @@ function dpl_admin_update_10000(array &$sandbox) : string {
     ->condition('uid', $new_user->id())
     ->execute();
   $feedback[] = "Transferred password hash from user 1 to new user {$new_user->id()}";
-  
+
   // Use a direct database connection here instead of SessionManager as this
   // does not allow ending sessions from the CLI.
   \Drupal::database()->delete('sessions')
--- a/web/modules/custom/dpl_admin/dpl_admin.install
+++ b/web/modules/custom/dpl_admin/dpl_admin.install
@@ -100,7 +100,8 @@ function dpl_admin_update_10000(array &$sandbox) : string {
   }, $user_1->getRoles(exclude_locked_roles: TRUE));
   $user_1->save();
   $removed_roles_string = implode(', ', $removed_roles);
-  $feedback[] = "Removed roles {$removed_roles_string} from user 1";
+  $removed_role_count = count($removed_roles);
+  $feedback[] = "Removed {$removed_role_count} roles {$removed_roles_string} from user 1";
 
   $new_user->save();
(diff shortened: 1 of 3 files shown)


web/modules/custom/dpl_admin/dpl_admin.install
```diff
@@ -32,3 +41,102 @@ function dpl_admin_requirements(string $phase): array {
...
+  // Reset the existing user 1 account.
+  // Use random data for both username and password as this user is only meant
+  // to be used with onetime login links.
+  $random = new Random();
+  $user_1
+    ->set('field_author_name', "Administrator")
+    ->setUsername('admin_' . $random->name())
+    ->setEmail('')
+    ->setPassword($random->string(32));
+  $user_1->save();
```
Dresse, line 87: Maybe this is a minor thing, but maybe we should consider setting language settings (Site language and Administration pages language) back to english?

web/modules/custom/dpl_admin/dpl_admin.install
```diff
@@ -32,3 +41,102 @@ function dpl_admin_requirements(string $phase): array {
...
+
+  $user_1 = User::load(1);
+  if (!$user_1) {
+    throw new RuntimeException('Unable to load user 1');
+  }
+
+  // Some field values should not been transfered between the users as we
+  // intentionally want them to be unique.
+  $excluded_fields = ['uid' => TRUE, 'uuid' => TRUE, 'roles' => TRUE];
+  $user_1_fields = array_diff_key($user_1->getFields(include_computed: FALSE), $excluded_fields);
```
rasben, line 59:
```
$user_1->getFields(include_computed: FALSE)
```

uh.. i feel like an idiot for asking this, but is this even valid PHP?

web/modules/custom/dpl_admin/dpl_admin.install
```diff
@@ -32,3 +41,102 @@ function dpl_admin_requirements(string $phase): array {
...
+  $feedback[] = "Retrieved password hash for user 1";
+
+  // Transfer field data.
+  $new_user = (User::create($user_1_field_values))
+    ->addRole('local_administrator');
+
+  // Reset the existing user 1 account.
+  // Use random data for both username and password as this user is only meant
+  // to be used with onetime login links.
+  $random = new Random();
```
rasben, line 81: does this also return symbols and numbers?

web/modules/custom/dpl_admin/dpl_admin.install
```diff
@@ -32,3 +41,102 @@ function dpl_admin_requirements(string $phase): array {
...
+  $user_1
+    ->set('field_author_name', "Administrator")
+    ->setUsername('admin_' . $random->name())
+    ->setEmail('')
+    ->setPassword($random->string(32));
```
rasben, line 86: you dont reinitialize `$random`, after having used it for username. could this be a problem?

web/modules/custom/dpl_admin/dpl_admin.install
```diff
@@ -32,3 +41,102 @@ function dpl_admin_requirements(string $phase): array {
...
+  $user_1->save();
+  $feedback[] = "Reset username and password for user 1";
+
+  $removed_roles = array_map(function (string $role) use ($user_1) {
+      $user_1->removeRole($role);
+      return $role;
+  }, $user_1->getRoles(exclude_locked_roles: TRUE));
```
rasben, line 93: `$user_1->getRoles(exclude_locked_roles: TRUE)`

again, this looks wrong?

web/modules/custom/dpl_admin/dpl_admin.install
```diff
@@ -32,3 +41,102 @@ function dpl_admin_requirements(string $phase): array {
...
+  }, \Drupal::entityTypeManager()->getDefinitions());
+
+  /** @var array<string, array<EntityInterface&EntityOwnerInterface>> $user_owned_entities_by_types */
+  $user_owned_entities_by_types = array_filter($entities_by_types, function (array $entities) {
+    return count($entities) > 0 && reset($entities) instanceof EntityOwnerInterface;
+  });
+
+  $status_by_types = array_map(function (array $entities, string $entity_type) use ($new_user) {
+    $entities = array_map(fn (EntityOwnerInterface $entity) => $entity->setOwner($new_user), $entities);
+    $entities = array_map(fn (EntityInterface $entity) => $entity->save(), $entities);
```
rasben, line 126: can you add a comment on what is happening here?

web/modules/custom/dpl_admin/dpl_admin.install
kasperg: Yes :smile:

web/modules/custom/dpl_admin/dpl_admin.install
kasperg: I do not think so. There is no state within [the Random class](https://github.com/drupal/drupal/blob/10.1.x/core/lib/Drupal/Component/Utility/Random.php).

It passes calls to [`mt_rand()`](https://www.php.net/manual/en/function.mt-rand.php) which on the other hand states that it does not provide unguessable values. That might be an issue.

@spaceo, do you have input here?

web/modules/custom/dpl_admin/dpl_admin.install
kasperg: It's not :smile:

web/modules/custom/dpl_admin/dpl_admin.install
kasperg: I have replaced `Random` with [`DefaultPasswordGenerator`](https://github.com/drupal/drupal/blob/10.1.x/core/lib/Drupal/Core/Password/DefaultPasswordGenerator.php) which is also used by Drupal Core when creating new passwords. 

This uses [`random_int()`](https://www.php.net/manual/en/function.random-int.php) instead of `mt_rand()` which is also suggested by PHP. As such I think this approach is good.

web/modules/custom/dpl_admin/dpl_admin.install
kasperg: This depends on what function you use. `Random::name()` which is used for the admin name [will return letters and numbers - not symbols](https://github.com/drupal/drupal/blob/10.1.x/core/lib/Drupal/Component/Utility/Random.php#L110).

We now use a different serice for passwords which is better suited for this.

web/modules/custom/dpl_admin/dpl_admin.install
kasperg: I have tried to document the process by adding more explatory code comments.

rasben (approved):

PR #1164, opened 2024-05-22
Update Event API, to supply and recieve all external data.

rasben:
Add new 'field_external_admin_link' to eventinstances. DDFFORM-797

It is on purpose that it is not set up on the series and that there is
no inheritance:
Third parties only have access to edit/see instances.

----

Update event PATCH, to also add admin link. DDFFORM-797

----

Also supply external data with GET:/Events API. DDFFORM-797

https://reload.atlassian.net/browse/DDFFORM-797


kasperg (approved):
:+1: LGTM

I appreciate the comment in the PR about not making this part of field inheritance. I think that is the right call.

PR #1169, opened 2024-05-23
DDFFORM 611 backend anbefalings komp

LasseStaus:
#### Link to issue

Ticket: [DDFFORM-611](https://reload.atlassian.net/browse/DDFFORM-611)

This PR Depends on: 

React PR: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1223
Design system PR: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/636

#### Description

This pr creates a new field widget for the custom field dpl_fbi_work_id which mounts a react app. The react app allows searching and selected materials for display. 

The widget contains two hidden fields that are used to save the values selected through the react app. 

The react app needs a unique identifier for these fields. Comment on the implementation of this and whether it is overkill. 

Remember to thoroughly test this in drupal with all changes in cohesion.

Relevant paragraphs are recommendation & materialGridManual, unless i am forgetting others. 

#### Screenshot of the result

image

[DDFFORM-611]: https://reload.atlassian.net/browse/DDFFORM-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdSearchForMaterialWidget.php
+++ b/web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdSearchForMaterialWidget.php
@@ -1,0 +1,108 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\dpl_fbi\Plugin\Field\FieldWidget;
+
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Field\WidgetBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\dpl_react_apps\Controller\DplReactAppsController;
+
+/**
+ * Defines the 'dpl_fbi_work_id_search_for_material' field widget.
+ *
+ * @FieldWidget(
+ *   id = "dpl_fbi_work_id_search_for_material",
+ *   label = @Translation("Work ID search for material"),
+ *   field_types = {"dpl_fbi_work_id"},
+ * )
+ */
+final class WorkIdSearchForMaterialWidget extends WidgetBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state): array {
+
+    $uniqueFieldId = $items->getFieldDefinition()->getUniqueIdentifier();
+    $identifier = "{$uniqueFieldId}-{$delta}";
+
+    $element['fields_container'] = [
+      '#type' => 'container',
+      '#attributes' => [
+        'class' => ['material-search__inputs-container--hidden'],
+      ],
+    ];
+
+    $element['fields_container']['value'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Field work ID'),
+      '#default_value' => $items[$delta]->value ?? '',
+      '#attributes' => [
+        'data-field-input-work-id' => $identifier,
+      ],
+    ];
+
+    $element['fields_container']['material_type'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Field material type'),
+      '#default_value' => $items[$delta]->material_type ?? '',
+      '#attributes' => [
+        'data-field-input-material-type-id' => $identifier,
+      ],
+    ];
+
+    $element['react_app_container'] = [
+      '#type' => 'container',
+      '#theme' => 'dpl_react_app',
+      '#name' => 'material-search',
+      '#data' => [
+        'class' => ['react-app-container'],
+        'previously-selected-work-id' => $items[$delta]->value ?? NULL,
+        'previously-selected-material-type' => $items[$delta]->material_type ?? NULL,
+        'unique-identifier' => $identifier,
+        'material-search-search-input-text' => $this->t('Search for material', [], ['context' => 'Material search']),
+        'material-search-material-type-selector-text' => $this->t('Select material type', [], ['context' => 'Material search']),
+        'material-search-material-type-selector-none-option-text' => $this->t('Select material type', [], ['context' => 'Material search']),
+        'material-search-no-material-selected-text' => $this->t('No material selected', [], ['context' => 'Material search']),
+        'material-search-preview-title-text' => $this->t('Title', [], ['context' => 'Material search']),
+        'material-search-preview-author-text' => $this->t('Author', [], ['context' => 'Material search']),
+        'material-search-preview-publication-year-text' => $this->t('Publication year', [], ['context' => 'Material search']),
+        'material-search-preview-source-text' => $this->t('Source', [], ['context' => 'Material search']),
+        'material-search-preview-work-id-text' => $this->t('Work ID', [], ['context' => 'Material search']),
+        'material-search-loading-text' => $this->t('Loading...', [], ['context' => 'Material search']),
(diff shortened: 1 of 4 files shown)


web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdSearchForMaterialWidget.php
```diff
@@ -0,0 +1,108 @@
...
+        'material-search-material-type-selector-none-option-text' => $this->t('Select material type', [], ['context' => 'material search']),
+        'material-search-no-material-selected-text' => $this->t('No material selected', [], ['context' => 'material search']),
+        'material-search-preview-title-text' => $this->t('Title', [], ['context' => 'material search']),
+        'material-search-preview-author-text' => $this->t('Author', [], ['context' => 'material search']),
+        'material-search-preview-publication-year-text' => $this->t('Publication year', [], ['context' => 'material search']),
+        'material-search-preview-source-text' => $this->t('Source', [], ['context' => 'material search']),
+        'material-search-preview-work-id-text' => $this->t('Work ID', [], ['context' => 'material search']),
+        'material-search-loading-text' => $this->t('Loading...', [], ['context' => 'material search']),
+        'material-search-amount-of-results-text' => $this->t('Total amount of results', [], ['context' => 'material search']),
+        'material-search-aria-button-select-work-with-text' => $this->t('Select work with', [], ['context' => 'material search']),
```
kasperbirch1, line 76: Remember to change to `Select work with title @title` if you integrate placeholders within your translation in React.

web/modules/custom/dpl_react/dpl_react.libraries.yml
```diff
@@ -249,8 +264,7 @@ opening-hours-editor:
     /libraries/dpl-react/OpeningHoursEditor.js: {}
   css:
     base:
-      /themes/custom/novel/assets/dpl-design-system/css/opening-hours-editor.css:
-        {}
+      /themes/custom/novel/assets/dpl-design-system/css/admin-base.css: {}
```
kasperbirch1, line 267: Thanks, this should have been done earlier 🙏

kasperbirch1:
I think it looks fine. ☺️ but it is probably a good idea that someone else with more Drupal experience looks at it as well

kasperbirch1 (approved):
I think it looks fine. ☺️ but it is probably a good idea that someone else with more Drupal experience looks at it as well

rasben (approved):

web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdSearchForMaterialWidget.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+final class WorkIdSearchForMaterialWidget extends WidgetBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state): array {
+
+    $uniqId = uniqid((string) rand(), TRUE);
+    $identifier = "{$uniqId}-{$delta}";
```
kasperg, line 29: Please consider this approach.

I understand the intention to generate a unique identifier for the field - and I am pretty sure that this will be the case for the current implementation - but is seens as if you misunderstand parts of what the different elements do.

[`uniqid()`](https://www.php.net/manual/en/function.uniqid.php) generateres a unique id based on the current time in microsecords. You then use a random value for the prefix.

The description of the `prefix` argument is to separate different ids which might be generated at the same microsecond - e.g. across servers or the like. 

You then append the delta - the numeric instance of the field if it has multiple values.

As I see it we only need to ensure uniqueness on a single page/form.

You already have the delta. You can get a unique identifier for the field through `FieldItemListInterface->getFieldDefinition()->getUniqueIdentifier()`. I would think that you would get something sensible by combining the two?

web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdSearchForMaterialWidget.php
```diff
@@ -0,0 +1,108 @@
...
+    $element['react_app_container'] = [
+      '#type' => 'container',
+      '#theme' => 'dpl_react_app',
+      '#name' => 'material-search',
+      '#data' => [
+        'class' => ['react-app-container'],
+        'previously-selected-work-id' => $items[$delta]->value ?? NULL,
+        'previously-selected-material-type' => $items[$delta]->material_type ?? NULL,
+        'unique-identifier' => $identifier,
+        'material-search-search-input-text' => $this->t('Search for material', [], ['context' => 'material search']),
```
kasperg, line 65: I would suggest context that either used proper casing e.g. `Material search` or had the module name as a prefix e.g. `DPL FBI material search` but having looked through other places our usage is all over the place so this is fine here as well.

web/modules/custom/dpl_react/dpl_react.libraries.yml
```diff
@@ -239,6 +239,21 @@ material-grid-manual:
...
+    url: https://github.com/danskernesdigitalebibliotek/dpl-react/blob/master/LICENSE
+    gpl-compatible: true
+  js:
+    /libraries/dpl-react/MaterialSearch.js: {}
+  css:
+    base:
+      /themes/custom/novel/assets/dpl-design-system/css/admin-base.css: {}
+  dependencies:
+    - dpl_react/base
+    - dpl_react/handler
```
kasperg, line 255: Please move `admin-base.css` to a separate library and then include it as a dependency like we do for `dpl_react/base`.

composer.json
```diff
@@ -88,8 +88,8 @@
         "composer/installers": "1.12.0",
         "cweagans/composer-patches": "1.7.3",
         "danskernesdigitalebibliotek/cms-api": "*",
-        "danskernesdigitalebibliotek/dpl-design-system": "2024.20.1",
-        "danskernesdigitalebibliotek/dpl-react": "2024.20.1",
+        "danskernesdigitalebibliotek/dpl-design-system": "2024.20.2-rc1",
+        "danskernesdigitalebibliotek/dpl-react": "2024.20.2-rc1",
```
kasperg, line 92: We have to remember to revert these changes once the related PRs have landed in the respective repos.

kasperg (approved):
:+1: Well done. This is nice and simple.

I have a few thoughts about how you handle ids.

It also seems from the screenshot that the materials are too large to fit within their elemennts in the list. Having tried it out in practice that is not actually the case.

web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdSearchForMaterialWidget.php
LasseStaus: Changed the casing.

web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/WorkIdSearchForMaterialWidget.php
LasseStaus: I see. I didn't realise `FieldItemListInterface->getFieldDefinition()->getUniqueIdentifier()` was available. have updated 👍

PR #1172, opened 2024-05-23
Replace hashing of openid_connect authmap sub ident

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-664

#### Description

A bug was reported that one user was able to se another users data.
We traced down the issue and found out that our technique of hashing the user ident used in the openid_connect authmap was not working correctly. Several users could end up having the same ident and thereby sharing the same issued tokens.
This PR changes the hashing method and also contains tests that proves it works, opposed to the former solution.

### DIFF
--- a/web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
+++ b/web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Drupal\dpl_login;
+
+use Drupal\Core\Site\Settings;
+
+/**
+ * Access Token.
+ */
+class OpenIdUserInfoService {
+
+  /**
+   *
+   */
+  public function __construct(
+    private Settings $settings
+  ) {}
+
+  /**
+   *
+   */
+  public function getOpenIdUserInfoFromAdgangsplatformenUserInfoResponse(array $response): array {
+    $name = uniqid();
+    // Drupal needs an email. We set a unique one to apply to that rule.
+    $userinfo['email'] = sprintf('%s@dpl-cms.invalid', $name);
+    // Drupal needs a username. We use the unique id to apply to that rule.
+    $userinfo['name'] = $name;
+    // openid_connect module needs the sub for creating the auth map.
+    $userinfo['sub'] = $this->getSubHashFromUserInfo($response);
+
+    return $userinfo;
+  }
+
+  /**
+   *
+   */
+  public function getSubHashFromUserInfo(array $userinfo): string {
+    if (!$cpr = $userinfo['attributes']['cpr'] ?? FALSE) {
+      if (!$uniqueId = $userinfo['attributes']['uniqueId'] ?? FALSE) {
+        throw new \Exception('Unable to identify user. Both CPR and uniqueId are missing.');
+      }
+    }
+
+    $id = $cpr ?: $uniqueId;
+    return $this->hashIdentifier($id);
+  }
+
+  /**
+   *
+   */
+  public function hashIdentifier(string $identifier): string {
+    return crypt($identifier, $this->settings::getHashSalt());
+  }
+
+}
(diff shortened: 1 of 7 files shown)


web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
```diff
@@ -0,0 +1,110 @@
...
+namespace Drupal\dpl_login;
+
+use Drupal\Core\Site\Settings;
+
+use function Safe\sprintf;
+
+/**
+ * Access Token.
+ */
+class OpenIdUserInfoService {
```
kasperg, line 12: Please update this DocBlock.

The current content does not make any sense to me.

web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
```diff
@@ -0,0 +1,110 @@
...
+   * @param mixed[] $userinfo
+   *   The userinfo from the adgangsplatformen userinfo endpoint.
+   *
+   * @return string
+   *   The hashed opemid_connect sub id.
+   */
+  public function getSubIdFromUserInfo(array $userinfo): string {
+    $identifier_data = $this->getIdentifierDataFromUserInfo($userinfo);
+    return $this->hashIdentifier($identifier_data['id']);
+  }
```
kasperg, line 56: Consider updating the documentation here.

It is not very clear what the "sub" actually is and what it is used for. We found out that it stands for subject identifier, and [it seems to be part of the OpenID Connect specification](https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes). Adgangsplatformen supports OAuth2 - not necessarily OpenID connect so that is probably one of the reasons why we need to do some messaging of data.

web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Get the identifier data from the userinfo.
+   *
+   * @param mixed[] $userinfo
+   *   The userinfo from the adgangsplatformen userinfo endpoint.
+   *
+   * @return mixed[]
+   *   The identifier data: Raw id and type of id.
```
kasperg, line 65:
```suggestion
   * @return array{"id": string, "type": AuthorizationIdType}
   *   The identifier data: Raw id and type of id.
```

We could use array shapes for a little more safety here.

web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Get the user info from the response.
+   *
+   * From the adgangsplatformen userinfo endpoint.
+   *
+   * @param mixed[] $response
+   *   The response from the userinfo endpoint.
+   *
+   * @return mixed[]
+   *   The openid_connect user info.
```
kasperg, line 30:
```suggestion
   * @return array{'email': string, 'name': string, 'sub': string}
   *   The openid_connect user info.
```

web/modules/custom/dpl_login/tests/src/Unit/OpenIdUserInfoServiceTest.php
```diff
@@ -0,0 +1,310 @@
...
+        '1234567809',
+      ],
+      'a pair of cpr that starts with the same nine characters' => [
+        '1234567890',
+        '1234567891',
+      ],
+      'a pair of unique_ids that start with the same character' => [
+        'e1a0eccd-6a89-47ad-8865-6ac91c104f22',
+        'e0aca8c3-5e36-42e0-b9e5-bd867e3b4599',
+      ],
```
kasperg, line 190: So many options 😅

kasperg (approved):
:+1: I think the core change looks good. 

I have a few suggestions and at least one DocBlock needs to be updated.

web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
spaceo: 👍

web/modules/custom/dpl_login/src/OpenIdUserInfoService.php
spaceo: I'll make it more clear with more documentation and renaming

kasperg:
Closed in favour of #1172.

PR #1174, opened 2024-05-24
DDFLSBP-641-remove access to email configuration

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-641

#### Description

The library editors kept setting their own values in 'from mail' and 'from name' in the webform email handler configurations. This meant that the emails would not be send out, as Azure Communication Services require a specific 'from mail' and 'from name'. 

After ACS have verified 'mail@folkebibliotekernescms.dk', we wan't to add that as the new site email instead of the old 'DoNotReply@folkebibliotekernescms.dk'.

To change the specific email from 'DoNotReply@folkebibliotekernescms.dk' to 'mail@folkebibliotekernescms.dk' the PR adds the following:
- Set the email in settings.php to make sure that we always override with the correct email.
- Updated system.site.yml to use the new email. As system.site is in config ignore, this won't have an effect, but is updated nonetheless for consistency. 
- Added update hook that updates the email configuration.

We also want to prevent library editors from changing the generel site email, so we introduce a new form alter hook that disables the editing of the field in the admin UI. 

To make sure that all webform email handlers works correctly, we are removing the access to changing them, and instead we now set them to the correct values in a form alter hook.
A new update hook has also been added which runs through all existing webforms and saves them, as this will update the "wrong" values of the 'from mail' and 'from_name' and force all email handlers to use the default values instead.


clausbruun (approved):
LGTM 🎉

PR #1175, opened 2024-05-24
Feature/DDFLSBP-642 - Enable more form elements and conditional logic

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-642

#### Description

The libraries wants more configuration functionality added to the webform module.
This PR enables the following elements:
- Hidden
- Number
- Select other
- Telephone
- Date
- Date/Time

The elements does not need any styling, and thus it is enough to just add them in configuration.

The libraries also requested the "Conditions" functionality, which give them access to setup conditional logic on elements in a form. This permission has now been added to the "editor" and "local_administrator" roles.

### DIFF
--- a/config/sync/user.role.editor.yml
+++ b/config/sync/user.role.editor.yml
@@ -25,6 +25,7 @@ dependencies:
     - dpl_admin
     - dpl_footer
     - dpl_opening_hours
+    - dpl_webform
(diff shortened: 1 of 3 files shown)


spaceo (approved):
LGTM 👍

PR #1176, opened 2024-05-24
Introduce logic for config ignoring, using config_ignore_auto`

rasben:
https://reload.atlassian.net/browse/DDFFORM-82

Enable & setup config_ignore_auto module. DDFFORM-82

We want site administrators to be able to add their own modules, along with their own settings, and this is not possible with Drupal configuration management.
In the past, we used `config_ignore`, but this required a massive amount of work for developers, as all new configuration had to be added manually.

The `config_ignore_auto` module works, by us enabling it on production sites, and then treating any config changes as something that cannot be overriden in the future by `drush cim`.

See the future ADR for more documentation about pros, cons and effects.


web/modules/custom/dpl_base/dpl_base.install
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/**
+ * Run on the initial site setup. Remember to add individual update hooks.
+ */
+function dpl_base_install(): string {
+  $messages[] = dpl_base_update_10001();
+
+  return implode('\r\n', $messages);
+}
```
kasperg, line 19: I do not understand how to interpret the code comment "Remember to add individual update hooks."

Do we need to add all future update hooks here as well?

kasperg (changes requested):
:+1: I have read through the code and documentation here. I think it looks promising but there are some things that I think needs further clarification.

docs/architecture/adr-001-configuration-management.md
```diff
@@ -32,23 +32,46 @@ A solution to configuration management must live up to the following test:
...
+       drush config:get config_ignore_auto.settings ignored_config_entities
+         --format json
+     ```
+
+      and then either decide to override it, or migrate the existing.
+
+3) A last, and final consequence, is that we need to treat permissions more
+   strictly that we do now.
+   - An exampls is `adminster site settings` also both allows stuff we want to
+     ignore (site name), but also things we don't want to ignore (404 node ID).
```
kasperg, line 74: To me this needs further elaboration.

1. I think it is worth spelling out the basic premise of the module: If you have the possibility (permission) to modify a configuration element then any changes you make should not be overridden on future deployments,
2. If we do not need to work with the `config_ignore` module directly is there a reason to reference the module/option at all? Could we migrate existing setup to `config_ignore_auto`?
3. Consequence 2 with rolling back a permission, referencing Drush etc. is not clear to me. We would not want to detect or update such a situation using Drush but rather through an update or deploy hook.
4. Given consequence 3 have you gone through the site and checked that users (primarily administrators) do not have permission to update configuration that we do not want to ignore?

web/modules/custom/dpl_base/dpl_base.info.yml
```diff
@@ -0,0 +1,5 @@
+name: "DPL Base"
+type: module
+description: "Generalized logic, such as enabling/disabling modules, migrating data or similar."
+package: DPL
```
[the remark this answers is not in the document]
spaceo: I second that. It is dangerous to name a module base or something to broad since it introduce that all kind of hacks or just features where people do not decide which module should be responsible for it.

docs/architecture/adr-001-configuration-management.md
spaceo, line 74: The way I understand it we need to solve cases like "site settings" with fine grained ignore rules. So I think it is worthwhile stating that "being careful" implies that we should make sure that what should be ignored is clearly targeted - with a full path.

web/modules/custom/dpl_base/dpl_base.info.yml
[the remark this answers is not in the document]
spaceo: But I am not sure that I agree in using the profile for it.
Could it make sense to call the module dpl_webmaster_config (or just dpl_config) and add the config_ignore_auto as a dependency to that?

web/modules/custom/dpl_base/dpl_base.install
spaceo: Normally we use update hooks in the profile for general site updates, right?

composer.json
```diff
@@ -100,6 +100,7 @@
         "drupal/azure_mailer": "^1.0",
         "drupal/color_field": "^3.0",
         "drupal/config_ignore": "^2.4",
+        "drupal/config_ignore_auto": "^2.0",
```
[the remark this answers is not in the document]
rasben: - v3 requires config_ignore v3
- we use config_ignore v2 at the moment
- if i try to update to config_ignore v3, it says a patch needs to be removed
- when i look at the patch details: https://www.drupal.org/project/config_ignore/issues/3252291 the latest answer scares me:

```quote
This issue is being closed because Config Ignore 2.x is deprecated.

The new 3,x version has been re-written from the ground up and works in a totally different way, yet it provides the same backwards compatible functionality. All the 2.x tests pass in 3.x and the 3.x codebase is easier to maintain.
Users of 2.x are strongly encouraged to upgrade to 3.x, as there is a very simple update path.

So likely this issue is no longer relevant. If you think that this issue still applies please open it again and target the new branch. All issues on the 2.x branch were closed in bulk.

File: docs/architecture/adr-001-configuration-management.md (hunk 2 above)
Comment (rasben) [reply docs/architecture/adr-001-configuration-management.md:74]: @kasperg Did you get answer for these things during the meeting?
Is there anything you'd like to update to the text?

**Comment by rasben** [comment]
## Testing results:

This is done locally, with `$config['config_ignore_auto.settings']['status'] = TRUE;`

**Adding, and enabling Google Analytics as "Administrator":**

Screenshot 2024-06-03 at 12 51 51

Screenshot 2024-06-03 at 12 52 16

**Administrator has access to manage permissions, including setting it on e.g. local_admin**
Screenshot 2024-06-03 at 12 53 56

**As administrator, we manage the google analytics settings:**

Screenshot 2024-06-03 at 12 54 15


**At this stage, the config auto settings looks like this:**

Screenshot 2024-06-03 at 12 58 22

**Running `drush cim -y` overrides the permissions set on the local_administrator, as we expect:**

```
dce drush cim -y
WARN[0000] /Users/rasben/code/dpl/cms/docker-compose.yml: `version` is obsolete
+------------+-------------------------------+-----------+
| Collection | Config                        | Operation |
+------------+-------------------------------+-----------+
|            | user.role.anonymous           | Update    |
|            | user.role.external_system     | Update    |
|            | user.role.patron              | Update    |
|            | user.role.mediator            | Update    |
|            | user.role.local_administrator | Update    |
|            | user.role.editor              | Update    |
|            | user.role.authenticated       | Update    |
|            | config_ignore_auto.settings   | Update    |
+------------+-------------------------------+-----------+
``` 

(The last line about config_ignore_auto.settings does nothing - it will be ignored.)

**Now, the permission has disappeared as we expect:**

Screenshot 2024-06-03 at 12 55 09

File: docs/architecture/adr-001b-configuration-management.md [hunk 5]
```diff
@@ -0,0 +1,115 @@
...
+
+The base module maintains a list of patterns for configuration which will be
+ignored  during the configuration import process. This allows us to avoid
+updating local configuration.
+
+Here, we can add some of the settings files that we already know needs to be
+ignored and admin-respected.
+But in reality, we don't need to do this manually, because of the second module:
+
+**Config Ignore Auto** is only enabled on non-development sites.
```
kasperg, line 48: This sounds like we only install the module on these sites but that is not actually the case.

Instead we rely on the status/enabling the module which is part fot the settings.

spaceo (approved):
I agree in the latest feedback coming from @kasperg 
I'll let  you two guys take it from here.

kasperg (approved):

PR #1179, opened 2024-05-24
DDFLSBP-665 - Enable confirmation message configuration

Dresse:
Link to issue
https://reload.atlassian.net/browse/DDFLSBP-665

Description
This PR makes it possible for users with the roles 'local_administrator' and 'editor' to change how confirmation messages should be shown by giving the roles the correct permission.

We also change the default confirmation type to be 'message', which means that the confirmation message will be shown on the same page as the webform.


web/modules/custom/dpl_webform/dpl_webform.module
```diff
@@ -16,6 +16,11 @@ function dpl_webform_webform_create(Webform $webform): void {
   // Therefore we disable saving of results for all new webforms.
   $webform->setSetting('results_disabled', TRUE);
   $webform->setSetting('results_disabled_ignore', TRUE);
+
+  // We want to set a new "default_value" to the "confirmation_type" setting.
+  // If we do not set this, the editors will have to manually change it when
+  // creating a new webform.
+  $webform->setSetting('confirmation_type', 'message');
```
spaceo, line 23: So what happens if the administrator deliberately want to choose another `confirmation_type` when they create the webform? Or is that even bot a possibility because that setting is hidden?

spaceo (changes requested):
Looks good just one question...

web/modules/custom/dpl_webform/dpl_webform.module
Dresse: The 'administer advanced webform confirmation settings' is now given to the 'editor' and 'local_administrator' roles. This permission gives them access to configuring the `confirmation_type`. So they can just change it to something else than the default value we set. 

The reason we set the default value at all is, that the normal default value redirects the user to the specific webforms page and show the confirmation message there. But we want the default behaviour to be to just reload the current page and show the confirmation massage.

spaceo (approved):
LGTM 👍

PR #1182, opened 2024-05-27
Move pincode-section-description-text to be a global text parameter

Adamik10:
#### Link to issue
Log on systemadministration.

#### Description
Move the pincode-section-description-text translation from being a single app translation to being global.

#### Screenshot of the result
-

-


clausbruun (approved):
Nice fix 🎉

PR #1183, opened 2024-05-27
Quick fix for opening hours Cypress test.

kasperbirch1:
#### Description

Update test to select the second opening hour instead of the third due to dates like 27 May having only 2 visible opening hours in the month view.

I need to investigate Cypress test configuration to set a specific "current" date to prevent issues with special view days.


LasseStaus (approved):
LGTM

PR #1186, opened 2024-05-27
Update card image styles, to respect focal point.

rasben:
Up until now, large and medium card image styles have been using a fake focal point - e.g., in the actual template, we only use the 'x large' image style.
This is because the size is decided by CSS.

This commit includes an update from the design system, that allows us to send along alternatively-cropped image styles, that CSS takes care of displaying when relevant.

DDFFORM-667

https://reload.atlassian.net/browse/DDFFORM-667

### DIFF
--- a/web/themes/custom/novel/templates/fields/field--media--field-media-image--card.html.twig
+++ b/web/themes/custom/novel/templates/fields/field--media--field-media-image--card.html.twig
@@ -1,0 +1,22 @@
+{% set alternative_image_styles = {'large': 'card_large', 'medium': 'card_medium'} %}
+{% set item = items.0 %}
+
+<div {{ attributes.addClass('image-credited__image') }}>
+  {#
+    We want to render some override images, that CSS takes care of displaying if relevant.
+    This is necessary, as the cards have different image styles.
+  #}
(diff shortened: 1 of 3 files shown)


rasben:
See https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/641 for the CSS logic. This PR also requires that to be merged first.

web/themes/custom/novel/templates/fields/field--media--field-media-image--card.html.twig
```diff
@@ -0,0 +1,22 @@
...
+    {% set url = uri|image_style(image_style) %}
+
+    {% if url %}
+      <div  class="card__override-images" data-card-style="{{ image_style_name }}"
+        style="display: none; background-image: url('{{ uri|image_style(image_style) }}')"></div>
+    {% endif %}
+
+  {% endfor %}
+
+  {# Render the default, fallback image. #}
```
LasseStaus, line 20: Marking this in case the rendering of the img is not necessary, as written in the design system repo.

LasseStaus (approved):
Have tested on the deployment and it works nicely 👍 

Will approve but see comment in design system before merging.

PR #1187, opened 2024-05-28
Add permission for uploading modules for administrator role (webmaster admin)

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-644

#### Description

This PR was originally created to test if it is possible to upload modules in a Lagoon environment.
It is possible.
And found out that I needed to add a permission to the "administrator" role of the site in order for administrators to be able to access the upload section of Drupal admin.
Since administrators are the role used for admins on websmaster sites that particular permission should only be added to the "administrator" role.


clausbruun (approved):
LGTM 🎉

PR #1191, opened 2024-05-29
DDFLSBP-650 - Add cookie information module and configuration

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-650

#### Description

This PR adds the cookieinformation module as well as its configuration.

The following configuration has been added:
- Enable cookieinformation module.
- Enable cookieinformation 'consent popup'.
- Exclude 'consent popup' on admin pages.
- Exclude 'consent popup' for UID 1 (super admin).

#### Screenshot of the result

Image showing the cookie consent popup:
Screenshot 2024-05-29 at 09 30 18

Cookie information dashboard setup:
This PR only handles the implementation of the cookieinformation contrib module and its configuration. Before the cookieinformation module works properly, it also needs to be setup in the cookieinformation dashboard. 
As the old librarie sites also uses cookieinformation, the cookieinformation dashboard configuration should already be in place for live sites. 
To test the cookieinformation module, we have also added 'dapple-cms.docker' to the list of dev environment domains in cookieinformation dashboard. 

Styling overrides:
When comparing the styling of the cookieinformation consent popup with how it looks on the existing sites, it looks like both versions have some overwrites some part of the default styling. We have asked the client about this, and will create another PR, if we need to change the styling of the content popup.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -102,6 +102,7 @@
         "drupal/config_ignore": "^2.4",
         "drupal/config_translation_po": "^1.0",
         "drupal/content_lock": "^2.3",
+        "drupal/cookieinformation": "^2.1",
         "drupal/core-composer-scaffold": "~10.1.7",
         "drupal/core-project-message": "~10.1.7",
         "drupal/core-recommended": "~10.1.7",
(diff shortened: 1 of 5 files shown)


clausbruun (approved):
LGTM 🎉 - However I don't see any cookie categories, but I assume that's because Cookie Information is not configured for http://dapple-cms.docker/

PR #1194, opened 2024-05-29
DDFFORM 456 feed med json

LasseStaus:
#### Link to issue

Jira ticket: [DDFFORM-456](https://reload.atlassian.net/browse/DDFFORM-456)

#### Description

This PR adds an endpoint `opening_hours/instances', which displays the opening hours of each branch in a legacy format used, which is requested/used by the app.  

The endpoint requires the parameters `from_date`, `to_date` and `nid`. 

Test that you can access the endpoint anonymously and view opening hours that have been created for a branch. 

Test that you are required to use all the necessary parameters before any opening hours are displayed, and that you receive an error message if you haven't used them all. 

The endpoint requires the parameters `from_date`, `to_date` and `nid`. 

Example, change the `nid` depending on which branch you are testing. : 
`/opening_hours/instances?nid=11&from_date=2021-03-04&to_date=2024-07-08`

#### Screenshot of the result

image

[DDFFORM-456]: https://reload.atlassian.net/browse/DDFFORM-456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursLegacyResource.php
```diff
@@ -0,0 +1,251 @@
...
+ *     "canonical" = "/opening_hours/instances",
+ *   },
+ * )
+ */
+final class OpeningHoursLegacyResource extends OpeningHoursResourceBase {
+
+  /**
+   * The cache tag to use for list operations.
+   */
+  const CACHE_TAG_LIST = 'dpl_opening_hours:list';
```
kasperg, line 35: Please remove the cache tag declaration here.

As I see it, the constant replicates what is already available from the base class.

web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursLegacyResource.php
```diff
@@ -0,0 +1,251 @@
...
+        ],
+      ],
+      "required" => [
+        "nid",
+        "category_tid",
+        "date",
+        "start_time",
+        "end_time",
+        "notice",
+      ],
```
kasperg, line 223: Please reconsider the definition of the notice part.

1. The description does not seem right. Perhaps a little bit of copy/paste here? :smile:
2. The field is marked as required but we explicitly set it to null - which is not a string as declared by the schema. That does not seem right. I suggest you look into marking the value nullable in the schema.

web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursLegacyResource.php
```diff
@@ -0,0 +1,251 @@
...
+    }
+    return $mimeType;
+  }
+
+  /**
+   * Invalidate the response cache.
+   */
+  protected function invalidateCache(): void {
+    $this->cacheTagsInvalidator->invalidateTags([self::CACHE_TAG_LIST]);
+  }
```
kasperg, line 249: Please remove this function.

1. It is not used anywhere within this class
2. It replicates what is already provided by the parent class

openapi.json
```diff
@@ -675,6 +675,100 @@
...
+            "in": "query",
+            "required": false
+          }, 
+          {
+            "name": "to_date",
+            "type": "string",
+            "format": "date",
+            "description": "Retrieve opening hours which occur before and including the provided date. In ISO 8601 format.",
+            "in": "query",
+            "required": false
```
kasperg, line 704: I am not sure this changed between reviews but I notice that the two parameters are not mentioned as required but we throw an exception if they are not provided. That does not seem right.

kasperg (approved):
:+1: LGTM

I have one more comment about the specification for you to consider before you merge. I do not know if has changed since my initial review or not. Please do not force push after a review. If you want to clean up afterwards you can commit the changes as fixups and rebase with autosquash before you merge.

openapi.json
LasseStaus: Whoops, good catch. my bad

PR #1202, opened 2024-06-01
`Banner` paragraph - DDFFORM 540

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-540

#### Description

This pull request introduces a new `banner` paragraph that includes a mandatory link and title. It can optionally include a description and background image to enhance its visual appeal.

The title field is implemented as a WYSIWYG field with a unique format named `Underlined title`, allowing only the `<u>` (underline) tag. This tag features a custom SVG underline.

Additionally, a new image style named Banner has been created, along with a corresponding view display mode for images, also named Banner.

**This is approved by DDF see screenshot in ticket**
For now, I have chosen not to change the arrow to a button if a background image is not selected because the whole paragraph is a link. I think it should be a different paragraph if it has to change more expression than it already does. see Figma link: https://www.figma.com/design/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?node-id=446-6958&t=KdryqdDHtKoKPuu7-4. 

#### Test
http://varnish.pr-1202.dpl-cms.dplplat01.dpl.reload.dk/test-banner


web/themes/custom/novel/templates/paragraphs/paragraph--banner.html.twig
```diff
@@ -0,0 +1,23 @@
+{%
+  set bannerContentClasses = [
+  'banner-content arrow__hover--right-large',
+  content.field_banner_image.0 ? '' : 'banner-content--no-image'
+]
+%}
+
+<a href="{{ content.field_banner_link.0["#title"] }}" class="banner ">
```
LasseStaus, line 8:
```suggestion
<a href="{{ content.field_banner_link.0["#title"] }}" class="banner">
```

LasseStaus: There is a general problem with the URL's at the moment, in that they don't support proxy'ies. [This is mentioned here](https://reload.atlassian.net/browse/DDFFORM-811). 

This applies to this current paragraph aswell. 

I suggest however you disregard that as i intend to fix it across the solution, i just wanted to mention it so you knew.

config/sync/field.field.paragraph.banner.field_banner_title.yml
```diff
@@ -0,0 +1,24 @@
...
+label: 'Banner Title'
+description: ''
+required: true
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  allowed_formats:
+    - underlined_title
+field_type: text_long
```
LasseStaus, line 24: I'm unsure on some things for this field. 

1. I think you should decrease the amount from rows from 5 to for example 2. I don't think a title should ever be longer than that, and there's no need for that many empty lines. 

2. I understand that this is banner paragraph, but the implementation for the underline is something that i assume should be use/replaced across multiple paragraphs since this behavior is wanted multiple places. In that regard i would consider: 
- Renaming this from explicitly using "banner" in the machine name, and make it more generic. This would make more sense it if became re-used across more paragraphs.
- Creating a ticket describing a task to update other paragraphs with the new underline implementation and get DDF to check it / write which paragraphs should be updated-

LasseStaus (changes requested):
Good stuff, i have some suggestions for you to consider.

Also see comments in design system about the backgroundImage

composer.json
```diff
@@ -88,7 +88,7 @@
         "composer/installers": "1.12.0",
         "cweagans/composer-patches": "1.7.3",
         "danskernesdigitalebibliotek/cms-api": "*",
-        "danskernesdigitalebibliotek/dpl-design-system": "2024.25.0",
+        "danskernesdigitalebibliotek/dpl-design-system": "2024.25.1-rc1",
         "danskernesdigitalebibliotek/dpl-react": "2024.25.0",
```
kasperg, line 92: Remember to revert this before we merge.

config/sync/filter.format.underlined_title.yml
```diff
@@ -0,0 +1,17 @@
...
+format: underlined_title
+weight: 0
+filters:
+  filter_html:
+    id: filter_html
+    provider: filter
+    status: true
+    weight: -10
+    settings:
+      allowed_html: '<br> <p> <u>'
```
kasperg, line 15: Please remove `<p>` from the allowed html.

[`<p>` tags are not allowed within headings](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#technical_summary).

web/modules/custom/dpl_example_content/content/node/fd774222-4b22-4803-ae45-2d4e80d84fac.yml
```diff
@@ -456,6 +461,42 @@ default:
...
+              entity: d97a25fd-a6ae-48b9-9a1a-7aec5372688e
+          field_banner_link:
+            -
+              target_uuid: e68b90e6-cd35-44c6-9e32-c05c982ed315
+              title: ''
+              options: {  }
+          field_underlined_title:
+            -
+              value: '<p>Opret en <u>bruger</u> på biblioteket</p>'
+              format: underlined_title
```
kasperg, line 499: :+1: for updating our example content with this.

config/sync/paragraphs.paragraphs_type.banner.yml
```diff
@@ -0,0 +1,18 @@
...
...TaGYDRrfFw0II3Ci9Ys7bMziavpsp/W5XaRNDQpLRNL1yONxFVaYCx2NqilxAkhMcHSlPDMQBG6ilMUtK6JrqhABVUZVqGzuU00KPQ7+OTH1tTFbPySm7nkhWXrQ2CN5ym+37sTdhyReLrBu+8oo32/ob/KgYpW4JKIIXE0RlqVxnJktd1XRMecRa5ENcFLZHaliwvYdjVRzVIa8Kb8xQE8MVVUCmlTisXLB/xu/jLRUuommhBUM+I6hWunBNCmb8jvdZetN2Ry2mV0UWIIW2Qr416ZapHvAvt+7E3p5IvF1gGjryijcQrcauY1WdtXAL1CPC5jk0hC6TpQCpcyr2UqtQSonlrhroinQscNQjY461QsrKglCjRRVCskQmmaw7blH06ThGWI1bU6J2rCLiqoGudp+VwVAqBUVFRUu0TSxMMWFWgtLflcmZTlsn5ibsjf2utgrA5De6zCsgvt36E3p5g2QVi6XXlFFFEq1Cto/gKzM7r1FtYa/ZNVUXIHEE+NzStRdECGFHcobi7Eqi43WG0tgkcXNqHCiNqiZZpeG+rS3Qd6phrVO0JuKbloVhPcIM+U/wAXDK863A0XFci+qZlohcLrJ+Ym7Iq0WgRD7lOtUpNQ6iZbpBvqm+oMHWpLRDJC6jxshuirKP8AIq3W79CZ08kXjZBWHZ2Qo3FTCsigFAFaW4oXjwmmiLk2Uuc4U2TXgGhKOqLGlCNoXZGNh3XDYNUXJzqhMxh2uouNw0QKadVLo68IXAVTIqoMDRQBEJ3Q7+M5YCuCPuuESQAV8E/7p9kc1pNU3kC6zGj01W2cxM03KdI5xq43Eotxbp/DwjDUFM2F1l6isQWNqtpBwJu15zC9q7qwnR2Qo3FTfmKIiiOoIUjcMjx5Wg3KlnDNGBcV5eHEqJ4ewFOrVYimv+6qnFAVVFUBVu3NwTXKXsbxcC0blMDCeoLUDRYysSd0u/jkx9QQU4/xuTdzyAgoOsKtGk+FaJnyOIcdAdFVyqVrdSoTboTQrEsQU5qQm7cgZAgrF3QuKcjfaBqCFE5DpVsbhtFfupwd+yDGkIsirShCsz8DuGduyKIUjMTdDqmYmtAN7nUWMppNdUDUKxNaZwXbDVWmAtxSPeA5x6bgUQXABPsM7Y8dBRCN5OjVHYiRVzqI2MDuvhIKagkqWxNArGSCo5nsdhejcT8rv4X/2Q=='
```
kasperg, line 16: I notice a lot of binary data here but as far as I can tell from the source of the YAML file it is ~70KB which is within the limits suggested by @rasben .

kasperg (approved):
:+1: Looks good to me.

Before you merge please look into if it is possible to remove the `<p>` tags from the filter format.

config/sync/filter.format.underlined_title.yml
kasperg: I checked this out and as far as I can tell this is not actually possible. I see that we have not enabled converting line breaks to `<p>` and `<br>` so that is probably sufficient.

LasseStaus: Yeah i dont see a way to do this either. 

image

LasseStaus (approved):

PR #1215, opened 2024-06-04
Re-export all example content, based on `develop`

rasben:
Re-export all example content, based on `develop`

Apparantly, various fields have been updated, but the
updates have not been passed along to the example content.
This is the result of running the following commands on `develop`:

```
drush default-content:export-module dpl_example_content
drush default-content:export-module dpl_example_breadcrumb
```

And, then finally doing a search/replace to fix any links:

"http://dpl-cms.docker/" => "/"


Dresse (approved):
I only glanced through the changed files. I assume they are correct. 

Rubberstamp LGTM 👍

PR #1077, opened 2024-05-01
Adding JSNLog module [DDFLSBP-569]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-569

#### Description

Add JSNLog module and corresponding ADR

#### Screenshot of the result

image

#### Sibling PR

https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1257

PR #1219, opened 2024-06-05
Init feature/cookie-implementation

Dresse:
This feature branch include all the work done on implementing cookie handling in dpl-cms. 

Tasks:

https://reload.atlassian.net/browse/DDFLSBP-651
https://reload.atlassian.net/browse/DDFLSBP-652


web/modules/custom/dpl_cookies/dpl_cookies.module
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+/**
+ * Implements hook_preprocess_media().
+ *
+ * Adds cookie consent attributes to media video fields.
+ */
+function dpl_cookies_preprocess_field__media__video(array &$variables): void {
+  foreach ($variables['items'] as &$item) {
+    if (isset($item['content']['#attributes']['src'])) {
```
spaceo, line 15: Did not double check but if we can use Drupals field helpers I think we should do that.
IS it possible to do something like:
```php
if(!$item->get('content')->isEmpty()) && !$item['content']->get('src')->isEmpty()) {

Comment (Dresse) [reply web/modules/custom/dpl_cookies/dpl_cookies.module:15]: Maybe a version of this is possible, but then we first need to load the field as an entity object and then use the above if statement. 

I don't think this is viable here.

File: .lagoon.yml [hunk 2]
```diff
@@ -122,3 +122,8 @@ environments:
         schedule: "M/30 * * * *"
         command: drush dpl_po:import-remote-config-po da https://danskernesdigitalebibliotek.github.io/dpl-cms/translations/da.config.po
         service: cli
+    routes:
+      - nginx:
+          - reload.dresscoder.dk:
+              tls-acme: "true"
+              insecure: Redirect
```
clausbruun, line 129: Skal dette med videre eller er det test indstillinger?

config/sync/core.extension.yml
```diff
@@ -33,6 +33,7 @@ module:
   dpl_breadcrumb: 0
   dpl_cache_settings: 0
   dpl_campaign: 0
+  dpl_cookies: 0
```
clausbruun, line 36: Denne bør udskiftes med en update hook i `dpl_update`, se fx. https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1077/commits/1ec893443d56749feefc6a8095d67d9469fec909

clausbruun (changes requested):

.lagoon.yml
Dresse: You are right, this should be removed, and is removed now!

config/sync/core.extension.yml
Dresse: I have added both the enabling to both the install and update hook now in dpl_update. It seems the process has been to keep the enabling of modules in core.extension.yml as well, even though it is now ignored. I have therefore kept it.

clausbruun (approved):
Very nice 🎉

PR #1231, opened 2024-06-06
Fix WSOD error when rendering 'Recommendation' paragraph. DDFFORM-830

rasben:
If an editor has entered a invalid workID in the recommendation paragraph, the site fails hard.
This could happen if the editor has entered
https://SOMESITE.dk/work/work-of:870970-basis:137107074?type=bog instead of
work-of:870970-basis:137107074

I mitigate the error by adding a `try/catch`, but this is a temporary solution, as the real solution comes with the upcoming recommendation backend, where the editor has to search for materials, and cannot make this mistake.

---

This is pretty critical, because this also stops the editor from being able to edit the page - the only way around it, is to delete the node and start over.

https://reload.atlassian.net/browse/DDFFORM-830


rasben:
The lagoon error is unrelated to the code.

web/modules/custom/dpl_paragraphs/dpl_paragraphs.module
```diff
@@ -159,15 +159,21 @@ function dpl_paragraphs_preprocess_paragraph__recommendation(array &$variables):
...
-        ['absolute' => TRUE, 'query' => $query_parameters]);
+        $material_url_object = Url::fromRoute('dpl_react_apps.work', $route_parameters,
+          ['absolute' => TRUE, 'query' => $query_parameters]);
 
-      $material_url = $material_url_object->toString();
+        $material_url = $material_url_object->toString();
+      }
+      catch (\Exception $e) {
+        \Drupal::logger('dpl_paragraphs')->error('Could not render Recommendation paragraph - most likely cause the @wid WorkID is in wrong format.', ['@wid' => $work_id]);
+      }
```
kasperg, line 176: Please reconsider this structure.

1. It seems as if the change wraps some part of the code (L165-168) in the try catch which is not related to the problem. If that is the case then please keep it out.
2. Please catch the relevant exception class (or parent class if we think a group of exceptions is relevant) instead of the generic `\Exception` class. This makes it possible to have other problems fly under the radar which should be found during development.
3. Please include the error message from the exception in the logged message. Otherwise it is even more difficult to know what has happened.
4. Consider what should happen if this error occurs. The current approach passes `NULL` as the material url. I wonder what will happen in this case. The error message assumes that the recommendation paragraph is not rendered at all but is that actually the case? It seems as we create the render array for the React app anyway. If it should not be rendered I would not have expected this.
5. This addreses the problem at runtime. Do we need to add some kind of validation to the form elements, update hooks to fix existing problems or the like?

kasperg (changes requested):

web/modules/custom/dpl_paragraphs/dpl_paragraphs.module
rasben: 1) Fixed
2) Fixed
3) Fixed
4) I set the URL to NULL, and then I would assume the material theme / react taking care of what should happen.

This is what happens - the skeleton shows up, and then an error shows up afterwards:
Screenshot 2024-06-09 at 16 04 18
Screenshot 2024-06-09 at 16 04 26

Which, is honestly what I would expect to happen - especially when you consider 5:

5) Nope, cause in a moment, we'll have the new backend where this is not an issue at all. I see this update mostly as a quickfix

kasperg (approved):

PR #1232, opened 2024-06-06
Show the paragraph and field label of fields that fail validation

rasben:
The current error message is too cryptic and confuses editors. DDFFORM-806

See the video here for more info:

https://www.loom.com/share/1ebf2e9e4b1a4723944a60802c6027c8?sid=ea433568-4494-4c3d-8a77-7241f9e9b9a6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -323,6 +323,9 @@
                 "3171530 + 3116760: Add support for parameter and response descriptions": "patches/openapi-parameter-response-descriptions-14-3.patch",
                 "3343816: Specification includes disabled resources": "https://git.drupalcode.org/project/openapi_rest/-/merge_requests/4.patch"
             },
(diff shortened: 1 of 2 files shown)


rasben:
The lagoon error is unrelated to the code.

kasperg (approved):
:+1: I think this is an improvement to the existing situation and I appreciated the video.

I would have liked some text in the error message that said someting like: "Please edit the paragraph to address the problem" or something like that to make it clearer to editors what they should do to address this.

To me this seems like different scenarions and configurations colliding.

- At first I thought Drupal would not detect the deleted reference but it actually does.
- If the paragraph form had been expanded then this would not be a problem.
- We have decided not to expand paragraph to forms by default for admin UX reasons.

Oh well.

PR #1234, opened 2024-06-06
Disable config-ignore-auto on local development machines.

rasben:
This was already done in development.settings.php but apparantly this is not the right file for local development.

This means that as a developer, you cannot export your config at the moment.

When config_ignore_auto is enabled, it means that saving config-forms will add the saved changes as 'ignored', and won't be overriden on next deploy. This is a problem for development, as we obviously want to be able to "cex" and "cim" our updates.

### DIFF
--- a/assets/local.settings.php
+++ b/assets/local.settings.php
@@ -9,3 +9,10 @@ $config['system.performance']['js']['preprocess'] = FALSE;
 
 # Disable caching.
 $settings['cache']['default'] = 'cache.backend.null';
+
+// Disable the "config auto ignore" module on local development.
+// When config_ignore_auto is enabled, it means that saving config-forms will
+// add the saved changes as 'ignored', and won't be overriden on next deploy.
+// This is a problem for development, as we obviously want to be able to
+// "cex" and "cim" our updates.
+$config['config_ignore_auto.settings']['status'] = FALSE;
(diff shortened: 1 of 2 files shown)


Dresse (approved):
LGTM 👍

assets/local.settings.php
```diff
@@ -9,3 +9,10 @@
 
 # Disable caching.
 $settings['cache']['default'] = 'cache.backend.null';
+
+// Disable the "config auto ignore" module on local development.
+// When config_ignore_auto is enabled, it means that saving config-forms will
+// add the saved changes as 'ignored', and won't be overriden on next deploy.
+// This is a problem for development, as we obviously want to be able to
+// "cex" and "cim" our updates.
+$config['config_ignore_auto.settings']['status'] = FALSE;
```
kasperg, line 18: Please consider the duplication of the setting in different `settings.php`.

We already have the same setting in [`development.settings.php`](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/ac5e411337120c3aca5b889bf6caaa7f25b65b24/assets/development.settings.php#L11). Apparently this does not get used?

If this is the case then we should at least remove the setting from the other file - or remove `development.settings.php` entirely.

kasperg (changes requested):

PR #1235, opened 2024-06-07
Expand event API fields. #DDFFORM-836 #DDFFORM-837 #DDFFORM-838 #DDFFORM-839

rasben:
I've added seperate, descriptive commits.
Review it commit by commit, and hopefully it should make sense.

You can see it in action (along with updated example content on the PR environment):

https://varnish.pr-1235.dpl-cms.dplplat01.dpl.reload.dk/api/v1/events?_format=json

(might not work, as lagoon is having problems..)


openapi.json
```diff
@@ -906,7 +910,7 @@
                   },
                   "description": {
                     "type": "string",
-                    "description": "An editorial description of the event."
+                    "description": "An editorial WYSIWYG/HTML description of the event."
```
kasperg, line 929: Please reconsider this.

As I see this this is effectively changing the API by repurposing a field `description` to contain something else - both in regard to length (if I understand where you are going this might contain more text) and format (with/without HTML).

This may effect our the integration partners. 

I would suggest that we instead add a new field for the longer text. If you want a second opinion on this I suggest taking it out with @spaceo, Arne or Fini.

openapi.json
```diff
@@ -836,6 +836,14 @@
                       }
                     }
                   },
+                  "branches": {
+                    "type": "array",
+                    "description": "The associated library branches.",
+                    "items": {
+                      "type": "string",
+                      "description": "The translated label of a branch."
```
kasperg, line 844: Please reconsider using the concept label here. 

As I see it label has little meaning outside of the Drupal community. As this is meant to be consumed by external parties I suggest using something like "name" or "title" instead. The description might even include an example if we want to clarify things.

openapi.json
```diff
@@ -863,6 +871,14 @@
                     },
                     "required": ["street", "zip_code", "city", "country"]
                   },
+                  "tags": {
+                    "type": "array",
+                    "description": "The tags associated with the event.",
+                    "items": {
+                      "type": "string",
+                      "description": "The translated label of a tag."
```
kasperg, line 879: See comment above regarding using "label".

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
```diff
@@ -83,6 +85,51 @@ public function getResponse(EventInstance $event_instance): EventsGET200Response
...
+
+    foreach ($branches as $branch) {
+      $label = $branch->label();
+
+      if ($label instanceof TranslatableMarkup) {
+        $return[] = $label->render();
+      }
+      elseif (is_string($label)) {
+        $return[] = $label;
+      }
```
kasperg, line 109: Consider using [`getTitle()`](https://github.com/drupal/drupal/blob/10.1.x/core/modules/node/src/NodeInterface.php#L54-L61) instead of `label()`.

`getTitle()` seems to have a nice and simple signature. Having looked through the code I wonder why you use `label()`.

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
```diff
@@ -83,6 +85,51 @@ public function getResponse(EventInstance $event_instance): EventsGET200Response
     ]);
   }
 
+  /**
+   * Getting associated branches.
+   *
+   * @return ?string[]
+   *  The translated branch labels.
+   */
+  private function getBranches(): ?array {
```
kasperg: Consider why we would want to return `NULL` here.

As I read the code an empty array would work just as well and simplify the code.

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
```diff
@@ -83,6 +85,51 @@ public function getResponse(EventInstance $event_instance): EventsGET200Response
...
+    return empty($return) ? null : $return;
+  }
+
+  /**
+   * Getting associated tags.
+   *
+   * @return ?string[]
+   *  The translated tag labels.
+   */
+  private function getTags(): ?array {
```
kasperg: See comment about `NULL` vs. empty value above.

kasperg (changes requested):
:+1: but I have a few concerns regarding the API and data handling. 

We now have three fields related to tickets in the admin UI. Do you think it would make sense place them in an expanded field group?

Monosnap Edit default Event Genbrugsstof fremfor plastik | DPL CMS | Logged in 2024-06-10 09-40-52

rasben:
>We now have three fields related to tickets in the admin UI. Do you think it would make sense place them in an expanded field group?

Mayhaps, but, I'd rather deal with that in a seperate PR, perhaps as part of a greater UX overhaul

openapi.json
rasben: Good point - I've updated it, so the new data is in a new `body` field instead.

kasperg (approved):

PR #1236, opened 2024-06-07
Don't show expired or unpublished events in filtered events paragraph DDFFORM-796

rasben:
A bit hard to test, cause lagoon refuses to create an environment, but the title and the code is pretty straight forward:

Add the same kind of filters that we have on the /events view.

https://reload.atlassian.net/browse/DDFFORM-796


kasperg (approved):
:+1: LGTM

PR #1240, opened 2024-06-10
Fix broken header, when logged into admin interface.

rasben:
Screenshot 2024-06-10 at 09 21 45

This is driving me insane. Here's a quickfix.


spaceo (commented):
Looks fine. Just one question:
What are the ideas behind adding the styling in a module called dpl_admin (in a file called: dpl_frontend.css) - and not in the theme.
Are there any clear conventions?

rasben:
The file is named `dpl_frontend.css` because it is also loaded on non-admin pages - which is necessary for the toolbar.

The reason I didnt add this into the design system, is that I'd prefer to keep Drupal stuff out of the design system, especially admin-ux hacks like this.

When you look at dpl_frontend.css and dpl_admin.css, you'll see this is also what we've used it for up until now.

spaceo:
> The file is named `dpl_frontend.css` because it is also loaded on non-admin pages - which is necessary for the toolbar.
> 
> The reason I didnt add this into the design system, is that I'd prefer to keep Drupal stuff out of the design system, especially admin-ux hacks like this.
> 
> When you look at dpl_frontend.css and dpl_admin.css, you'll see this is also what we've used it for up until now.

I don't get the point of keeping "Drupal stuff" out of the desi

spaceo (approved):
LGTM 👍

PR #1263, opened 2024-06-17
Automatic paragraphs: Re-enable RelatedContent + Introduce `OUTER_AND` option. DDFFORM-690

rasben:
https://reload.atlassian.net/browse/DDFFORM-690

## **Re-introduce RelatedContent as engine for automatic paragraphs.**

-------

## **Introduce `OUTER_AND` option for automatic paragraphs. DDFFORM-690**

We use RelatedContent engine for letting the editor choose filters to
find content for automatic paragraphs.
However, up until now, the editor only had two condition choices:
AND or OR between *everything*
In reality, they need a third option - AND between the outer items
(categories, tags), and OR between the inner items (tag1, tag2).
This mimics the old setup for automatic paragraphs, that we had with
views and contextual filters.

-----

## **RelatedContent: Treat branches same way as content types. DDFFORM-690**

We want to treat branches as a non-grouped filter.
E.g. - the logic about AND/OR should not relate. If you choose two
branches, it will always be OR, just like it will always be OR with
content types, as a peice of content can only ever have one branch or
one type.

This only affects the automatic paragraphs, as branches are not used in
the inspiration slider.

----

## **Enable broad search for automatic paragraphs. DDFFORM-690**

Rather than first looking up based on tags, and then looking up
based on tags and categories, we'll do the look-up directly
on tags and categories - this matches the old logic for the
automatic paragraphs.

If no filters are chosen for automatic paragraphs, we will do a broad
look up, that may result in a wide mix of content, based on date.

This is only done for paragraphs - this logic is turned off for the
inspiration slider that shows up on all content.


kasperbirch1 (approved):
After we looked at it together, it looks good to me, but I assume you and Stefan have tested it.

PR #1265, opened 2024-06-17
Re-introduce automatic paragraph logic via dpl_filtered_paragraphs. DDFFORM-861

rasben:
- Re-enable `dpl_filtered_paragraphs` module, and re-implement previous views
- Update previous views to also have content type filters
- Port existing automatic paragraphs to only pull articles, as they de-facto did before by default.
- Hide `cond type` field, that is currently inactive
- Temporarily disable paragraphs logic in `dpl_related_content`

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -41,6 +41,7 @@ module:
   dpl_fbi: 0
   dpl_fbs: 0
   dpl_fees: 0
+  dpl_filter_paragraphs: 0
   dpl_footer: 0
   dpl_instant_loan: 0
   dpl_library_agency: 0
--- a/config/sync/views.view.content_paragraphs.yml
+++ b/config/sync/views.view.content_paragraphs.yml
@@ -1,0 +1,340 @@
+uuid: 66163bca-af6b-4db4-b820-3579418763ea
+langcode: en
+status: true
+dependencies:
+  config:
+    - core.entity_view_mode.node.card
+  module:
+    - node
+    - user
+id: content_paragraphs
+label: 'Content lists, embedded in Paragraphs'
+module: views
+description: 'Lists for displaying content, through embedded views in paragraphs. Also allows the editor to set filter options in the paragraphs, that will be sent along using contextual filters. See dpl_filter_paragraphs.module'
+tag: ''
+base_table: node_field_data
+base_field: nid
+display:
+  default:
+    id: default
+    display_title: Default
+    display_plugin: default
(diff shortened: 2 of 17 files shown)


web/modules/custom/dpl_filter_paragraphs/dpl_filter_paragraphs.module
```diff
@@ -2,9 +2,162 @@
 
 /**
  * @file
- * This file is left empty on purpose.
+ * DPL filtered Paragraphs.
```
rasben, line 5: This huge chunk of code is mostly just the old code that is being taken back.

The differences:

This has been added:

```php
/**
 * Implements theme_preprocess_paragraph__NAME().
 *
 * @see _dpl_filter_paragraphs_prepare_filter_view()
 */
function dpl_filter_paragraphs_preprocess_paragraph__content_slider_automatic(array &$variables): void {
  $variables = _dpl_filter_paragraphs_prepare_filter_view($variables, 'content_paragraphs', 'slider', NULL);
}
```

and added to line 100:

```
'field_filter_content_types',
```

config/sync/views.view.content_paragraphs.yml
```diff
@@ -0,0 +1,340 @@
+uuid: 66163bca-af6b-4db4-b820-3579418763ea
```
rasben, line 1: This config is 99% original, except I've added:

- A new display, for showing sliders
- Removed the filter that says content must be articles
- Added content type as an contextual filter

config/sync/views.view.filtered_event_list.yml
```diff
@@ -0,0 +1,344 @@
+uuid: 7a8ef9f1-371e-41d6-abc9-d6d5534dc149
```
rasben, line 1: This is 100% the old view - nothing new.

config/sync/core.extension.yml
```diff
@@ -41,6 +41,7 @@ module:
   dpl_fbi: 0
   dpl_fbs: 0
   dpl_fees: 0
+  dpl_filter_paragraphs: 0
```
kasperg, line 44: Should this be renenabled in `dpl_update`?

kasperg (changes requested):
:+1: but remember the update hook.

config/sync/core.extension.yml
rasben: Yes, and now it is :)

LasseStaus (approved):

PR #1280, opened 2024-06-19
DDFFORM 725 breadcrumb children subtitle

LasseStaus:
- **Add field `field_show_subtitles` to paragraph breadcrumb children.**
- **Extend breadcrumb children preprocess to retrieve show_subtitles value**

#### Link to issue

[DDFFORM-725](https://reload.atlassian.net/browse/DDFFORM-725)

#### Description

This PR adds a 'show subtiles' button to the paragraph breadcrumb children. 

Create an article, add the paragraph breadcrumb children, test that subtitles will be shown if enabled. (There might not be a subtitle on whichever test content you are using, so add it ) 

#### Screenshot of the result

image

[DDFFORM-725]: https://reload.atlassian.net/browse/DDFFORM-725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/sync/field.field.paragraph.breadcrumb_children.field_show_subtitles.yml
+++ b/config/sync/field.field.paragraph.breadcrumb_children.field_show_subtitles.yml
@@ -1,0 +1,23 @@
+uuid: fd72e42e-f823-43e6-993b-31a3bca8c752
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.paragraph.field_show_subtitles
+    - paragraphs.paragraphs_type.breadcrumb_children
+id: paragraph.breadcrumb_children.field_show_subtitles
+field_name: field_show_subtitles
+entity_type: paragraph
+bundle: breadcrumb_children
+label: 'Show subtitles'
+description: ''
+required: false
+translatable: false
+default_value:
+  -
+    value: 0
+default_value_callback: ''
+settings:
+  on_label: 'On'
+  off_label: 'Off'
+field_type: boolean
--- a/web/modules/custom/dpl_breadcrumb/dpl_breadcrumb.module
+++ b/web/modules/custom/dpl_breadcrumb/dpl_breadcrumb.module
@@ -52,6 +52,9 @@ function dpl_breadcrumb_preprocess_paragraph__breadcrumb_children(array &$variab
 
   $variables['items'] = $service->getRenderedReferencingNodes($breadcrumb_item);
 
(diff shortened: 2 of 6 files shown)


web/modules/custom/dpl_breadcrumb/dpl_breadcrumb.module
```diff
@@ -37,7 +37,7 @@ function dpl_breadcrumb_preprocess_paragraph__breadcrumb_children(array &$variab
   $paragraph = $variables['paragraph'] ?? NULL;
 
   // In the preview (AKA backend view), we don't want to load dynamics.
-  if (!($paragraph instanceof Paragraph) || $variables['view_mode'] === 'preview') {
+  if (!($paragraph instanceof Paragraph) || $variables['view_mode'] === 'preview' || !$paragraph->hasField('field_show_subtitles')) {
```
rasben, line 40: i think it's a bit extreme to add the check here.

wouldnt it be better to wrap the place you use it (line 54) with an IF instead?

rasben (changes requested):

config/sync/field.field.paragraph.breadcrumb_children.field_show_subtitles.yml
```diff
@@ -0,0 +1,21 @@
...
+    - paragraphs.paragraphs_type.breadcrumb_children
+id: paragraph.breadcrumb_children.field_show_subtitles
+field_name: field_show_subtitles
+entity_type: paragraph
+bundle: breadcrumb_children
+label: 'Show subtitles'
+description: ''
+required: false
+translatable: false
+default_value: {  }
```
kasperg, line 16: Consider setting a default value to off. Having an undefined value for such values can be confusing.

web/modules/custom/dpl_breadcrumb/dpl_breadcrumb.module
LasseStaus: Fine by me, altho we do it like this mostly everywhere else i think. 

I moved it to 54: 

```
  if ($paragraph->hasField('field_show_subtitles')) {
    $variables['show_subtitles'] = (bool) $paragraph->get('field_show_subtitles')>value;
  }
```

I think it is a bit strange tho, since the field will be shown for the admin even if this fails 🤷

rasben (approved):

PR #1283, opened 2024-06-20
Fix WSOD crash when getting invalid eventstate through patch DDFFORM-896

rasben:
A third-party event supplier accidentally sent along a EventState with a typo - this caused the value to be saved in the field, but caused all lists where the event showed up, to fail with an ENUM error.

This commit does two things:
- Makes sure that we dont accept invalid states in the future. If it is not an valid state, it doesnt get saved.
- Make sure Drupal doesnt fail when existing invalid states are exposed

I've added logging for both scenarios.

https://reload.atlassian.net/browse/DDFFORM-896


web/modules/custom/dpl_event/src/EventWrapper.php
```diff
@@ -122,7 +124,18 @@ public function getState(): ?EventState {
...
+
+      if (is_null($enum)) {
+        $logger = DrupalTyped::service(LoggerInterface::class, 'dpl_event.logger');
+
+        $logger->error("Could not map '@value' to EventState enum.", [
+          '@value' => $value ?? 'NULL',
+        ]);
+      }
+
+      return $enum;
```
kasperg, line 138:
```suggestion
      try {
        return EventState::from($value['value']);
      } catch (\ValueError $e) {
         $this->logger->error($e->getMessage())
         return NULL;
      }
```

I think this can be simplified a bit using the exceptions already provided.

Note that the logger should be injected in the constructor instead of using `DrupalTyped`.

web/modules/custom/dpl_event/src/Plugin/rest/resource/v1/EventResource.php
```diff
@@ -114,6 +115,19 @@ public function patch(string $uuid, Request $request): Response {
...
+
+      if (is_null($state_enum)) {
+        $this->logger->error("Could not map '@value' to EventState enum.", [
+          '@value' => $state,
+        ]);
+      }
+
+      $state = $state_enum?->value;
+    }
+
```
kasperg, line 130: Please reconsider this change.

As I see we must notify the client and reject the request if an invalid value is passed. Otherwise: How are they going to know that they have done something wrong? We could also log the error internally but as I see it this is less important.

```suggestion
    if ($state) {
      try {
        $state_enum = EventState::from($state);
      } catch (\ValueError $e) {
        throw new \Symfony\Component\HttpFoundation\Exception\BadRequestException("Invalid value for state: {$state}")
      }
    }
```

web/modules/custom/dpl_event/src/EventWrapper.php
kasperg: In general I am a bit on the fence with this change.

As I see it we should check that our data is valid before we store it. Thus we could assume that the value will always be valid. On the other hand I see that there might be situations where that is not the case - perhaps we remove a value from an enum or the like. With that in mind I think this is OK.

kasperg (changes requested):

Taskfile.yml
```diff
@@ -218,6 +218,9 @@ tasks:
       - task dev:cli -- drush user:create local_administrator --password="test"
       - task dev:cli -- drush user:role:add 'local_administrator' local_administrator
 
+      - task dev:cli -- drush user:create external_system --password="external_system"
+      - task dev:cli -- drush user:role:add 'external_system' external_system
+
```
kasperg, line 223: :+1: Nice.

web/modules/custom/dpl_event/src/Plugin/rest/resource/v1/EventResource.php
```diff
@@ -114,6 +116,18 @@ public function patch(string $uuid, Request $request): Response {
...
 
     $state = $request_data->getState();
+
+    if ($state) {
+      try {
+        $state_enum = EventState::from($state);
+        $state = $state_enum->value;
+        $event_instance->set('field_event_state', $state);
+      }
+      catch (\Error) {
```
kasperg, line 126: Consider why you choose to catch the more generic `\Error`.

You could instead catch `ValueError` which is what occurs if an enum cannot be extracted from a string. [`\Error`](https://www.php.net/manual/en/class.error.php) is the base class for all internal PHP errors and could thus be thrown in other situations as well. This would lead to `BadRequest` responses even though it is nothing that the client can do about the situation.

Also: I did not know you did not have to assign the exception being thrown to a variable. This is nice for situations where the exception is not relevant/used. Nice. I learned something new today.

kasperg (approved):
:+1: Nice. I have a single comment for your consideration before you merge.

web/modules/custom/dpl_event/src/Plugin/rest/resource/v1/EventResource.php
rasben: I chose Error, because if something goes wrong, it needs to inform the API consumer :)

PR #1286, opened 2024-06-21
Don't show unrelated breadcrumb children due to ID collision DDFFORM-859

rasben:
The `field_content` on the structure taxonomy only supports nodes. If we don't do a check for node here, we will experience a collision as an eventinstance ID probably is identical to an unrelated node ID.

This commit also makes sure that structure terms are published.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-859

### DIFF
--- a/web/modules/custom/dpl_breadcrumb/src/Services/BreadcrumbHelper.php
+++ b/web/modules/custom/dpl_breadcrumb/src/Services/BreadcrumbHelper.php
@@ -306,17 +306,21 @@ class BreadcrumbHelper {
    * Find a breadcrumb item by entity, if it is added to the content structure.
    */
   public function getBreadcrumbItem(FieldableEntityInterface $entity): ?TermInterface {
-    $storage = $this->entityTypeManager->getStorage('taxonomy_term');
-
     $id = $entity->id();
 
-    if (!$id) {
+    // The field_content on the taxonomy term only supports nodes.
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_breadcrumb/src/Services/BreadcrumbHelper.php
```diff
@@ -306,17 +306,21 @@ private function getCategoryBreadcrumb(FieldableEntityInterface $entity, Breadcr
...
   public function getBreadcrumbItem(FieldableEntityInterface $entity): ?TermInterface {
-    $storage = $this->entityTypeManager->getStorage('taxonomy_term');
-
     $id = $entity->id();
 
-    if (!$id) {
+    // The field_content on the taxonomy term only supports nodes.
+    // If we don't do a check for node here, we will experience a collision
+    // as an eventinstance ID probably is identical to an unrelated node ID.
+    if (!$id || $entity->getEntityTypeId() !== 'node') {
```
kasperg, line 314: :+1: Appreciate this explanation.

Given the situation I wonder if it would be appropriate to change the signature for `getBreadcrumbItem()` to only accept instances implementing `NodeInterface` instead of `FieldableEntityInterface` but that might result in other types of PHP type errors we would have to handle elsewhere.

kasperg (approved):
:+1: LGTM

PR #1288, opened 2024-06-21
Exclude webforms from config translation (and potentiallky others in future)

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-744

#### Description

  Add concept of missing specific contexts based on regex patterns

  Since we know we want to exclude webforms we start with excluding them.
  The reason why we want exclude webform related config translation is
  that the data is too vulnerable to be able to translate. We have seen a
  case where an adminstrator ended up creating malformed yaml which
  crippled all sites after it was imported.


clausbruun (commented):
LGTM 🎉

clausbruun (approved):

PR #1289, opened 2024-06-21
Link new event_ticket_capacity inheritance on eventinstances. DDFFORM-899

rasben:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-899

I've created an upstream issue:

https://www.drupal.org/project/field_inheritance/issues/3456239

### DIFF
--- a/web/modules/custom/dpl_update/dpl_update.install
+++ b/web/modules/custom/dpl_update/dpl_update.install
@@ -60,13 +60,13 @@ function _dpl_update_update_field_inheritance(string $field_inheritance_name): s
   foreach ($entities as $entity) {
     try {
       if (!($entity instanceof EventInstance)) {
-        continue;
+        throw new Exception('Entity is not an expected EventInstance.');
       }
 
       $event_series = $entity->getEventSeries();
 
       if (!($event_series instanceof EventSeries)) {
-        continue;
+        throw new Exception('Entity parent is not an expected EventSeries.');
       }
 
       // This matches the key that is defined in field_inheritance.
@@ -85,7 +85,7 @@ function _dpl_update_update_field_inheritance(string $field_inheritance_name): s
       $entity->save();
       $count++;
     }
-    catch (\Exception $e) {
+    catch (\Throwable $e) {
       \Drupal::logger('dpl_update')->error('Could not update field_inheritance on eventinstance @id - Error: @message', [
         '@message' => $e->getMessage(),
         '@id' => $entity->id(),
@@ -93,10 +93,7 @@ function _dpl_update_update_field_inheritance(string $field_inheritance_name): s
     }
   }
 
-  return t(
-    "Updated @count eventinstances, linking field  '@field' to inherit from eventseries.",
-    ['@field' => $field_inheritance_name, '@count' => $count]
-  )->render();
+  return "Updated $count eventinstances, linking field  '$field_inheritance_name' to inherit from eventseries.";
 }
 
 /**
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -11,6 +11,93 @@
...
+        '@message' => $e->getMessage(),
+        '@id' => $entity->id(),
+      ]);
+    }
+  }
+
+  return t(
+    "Updated @count eventinstances, linking field  '@field' to inherit from eventseries.",
+    ['@field' => $field_inheritance_name, '@count' => $count]
+  )->render();
```
kasperg, line 99:
```suggestion
  return "Updated {$count} eventinstances, linking field  '{$field_inheritance_name}' to inherit from eventseries.";
```

Consider not using `t()` here. It is not really a value it makes sense to translate.

web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -11,6 +11,93 @@
...
+    try {
+      if (!($entity instanceof EventInstance)) {
+        continue;
+      }
+
+      $event_series = $entity->getEventSeries();
+
+      if (!($event_series instanceof EventSeries)) {
+        continue;
+      }
```
kasperg, line 70: Consider dropping the two instanceof checks.

As I see it they should not be necessary due to the typing. If they actually occur they will be swallowed silently which is not good either. We are not likely to notice the total count being lower than excepted.

If the values for whatever reason not are of the excepted type and subsequent code will error and you want to proceed with your loop you could change the catch statement below to use [`\Throwable`](https://www.php.net/manual/en/class.throwable.php) instead of `\Exception`. This interface includes internal PHP errors.

web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -11,6 +11,93 @@
...
+      // This matches the key that is defined in field_inheritance.
+      $state_key = $entity->getEntityTypeId() . ':' . $entity->uuid();
+      $field_inheritance = \Drupal::keyValue('field_inheritance')->get($state_key);
+
+      // In theory, an eventinstance could be set up to inherit from another
+      // entity than the eventseries - but in practice, this is really unlikely,
+      // and something we're willing to disregard.
+      $field_inheritance[$field_inheritance_name] = [
+        'entity' => $event_series->id(),
+      ];
```
kasperg, line 81: :+1: Agreed. Nice commenting.

kasperg (approved):
:+1: Nice. Looks good.

I have a few things for you to consider before this is merged.

web/modules/custom/dpl_update/dpl_update.install
rasben: The reason I use `t()` is for variable placeholders.

Would you prefer

```php
return "Updated $count eventinstances, linking field  '$field_inheritance_name' to inherit from eventseries.
```
?

rasben:
I've fixed what you've written, and I also tried to pull the database from københavn:

```
dce drush updb
 ------------ ----------- --------------- -----------------------------------
  Module       Update ID   Type            Description
 ------------ ----------- --------------- -----------------------------------
  dpl_update   10007       hook_update_n   10007 - Link new
                                           event_ticket_capacity inheritance
                                           on eventinstances.
 ------------ ----------- --------------- -----------------------------------


 Do you wish to run the specified pending updates? (yes/no) [yes]:
 > y

>  [notice] Update started: dpl_update_update_10007
>  [notice] Updated 258 eventinstances, linking field  'event_ticket_capacity' to inherit from eventseries.
>  [notice] Update completed: dpl_update_update_10007
 [success] Finished performing updates
```

kasperg (approved):

PR #1294, opened 2024-06-24
Add `metatag_open_graph` module to add specific image for Facebook

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-904

#### Description

This pull request adds the `metatag_open_graph` module, which allows us to use our teaser image for the `<meta property="og:image" content="" />` tag.

See https://varnish.pr-1294.dpl-cms.dplplat01.dpl.reload.dk/admin/config/search/metatag

#### Screenshot of the result
screencapture-dapple-cms-docker-admin-config-search-metatag-2024-06-24-13_50_34

#### Test
It can be tested by looking at the source code (right-click on the page and click "view source") for example, view-source:https://varnish.pr-1294.dpl-cms.dplplat01.dpl.reload.dk/artikler/3-gode-til-godnat-3-6-ar and checking that `<meta property="og:image" content="https://varnish.pr-1294.dpl-cms.dplplat01.dpl.reload.dk/sites/default/files/2024-02/ed-robertson-eeSdJfLfx1A-unsplash.jpg" />` is there.

If a teaser image is not set, this will not be applied.

### DIFF
--- a/config/sync/core.entity_form_display.node.article.default.yml
+++ b/config/sync/core.entity_form_display.node.article.default.yml
@@ -6,6 +6,7 @@ dependencies:
     - field.field.node.article.field_branch
     - field.field.node.article.field_canonical_url
     - field.field.node.article.field_categories
+    - field.field.node.article.field_open_graph_image
     - field.field.node.article.field_override_author
     - field.field.node.article.field_paragraphs
     - field.field.node.article.field_publication_date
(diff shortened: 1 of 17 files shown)


rasben (changes requested):
You're on the right track, but a few things:

- When I do your test, on your link, I dont see any `og:image` in the source. I'm checking as an anonymous user

- Based on your description, it seems like you're aiming for the original image to be displayed. 

This is not a great idea, as social media would like it to be served as a small and cropped version. If you check the source of e.g. a BUPL article, you can see that we have a specific `open_graph` and `twitter_card` image style. (I dont know if Twitter is still relevant enough for us to spend time on it, however..)

This is because Twitter and Facebook prefers differently cropped images:

FB: 
https://bupl.dk/sites/default/files/styles/open_graph/public/2023-11/s%C3%A6ttem%C3%B8de1.jpg

Twitter:
https://bupl.dk/sites/default/files/styles/twitter_card/public/2023-11/s%C3%A6ttem%C3%B8de1.jpg

- You've covered EventInstances and Nodes, but what about EventSeries? :)

PR #1298, opened 2024-06-25
Update patch for default content file handling

kasperg:
#### Link to issue

#### Description

The current patch was based on a merge request. This has recently been updating causing it to no longer apply.

To fix this we use a static patch instead. The content of the patch should be equivalent to the content of the merge request before the update:
https://git.drupalcode.org/project/default_content/-/compare/2.0.x...72340623794ddf66b7bc333f6cd47a3bef17a826?from_project_id=20623&straight=false

Unfortunately we cannot get a patch from a compare view on GitLab. https://stackoverflow.com/questions/76444499/gitlab-compare-get-raw-patch-diff-file


Dresse (approved):
LGTM 👍

PR #1300, opened 2024-06-25
Do not write certain extracted config translations in export

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-744

#### Description

Former changes only made sure that certain contexts were not extracted when splitting config into normal translations and config translations.

This change makes sure that the initial config translatione extraction ignores same contexts.


Dresse (approved):
LGTM 👍

PR #1306, opened 2024-06-25
Make sure nodes are re-saved when placed in the breadcrumb tree. DDFFORM-847

rasben:
On some node types, we have a structure field, that allows the editor to place the node as a child of an term.
However, when the actual node is placed in the tree as a parent (as in, a
term has been created, that links to this node), we want to take this editor control away, and automatically update the field. This happens in dpl_breadcrumb_node_presave().
However, if you first setup the node with all it's content, and then go add it to the tree, no save is triggered on the node, and then it ends up with a "broken" state.
To fix this, we run `save()` on nodes, when we detect a save on a breadcrumb term.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-847

----

I know it's a bit confusing :) If you'd like to get a better idea of the problem, then try to go through the test that Stefan has written in the ticket

### DIFF
--- a/web/modules/custom/dpl_breadcrumb/dpl_breadcrumb.module
+++ b/web/modules/custom/dpl_breadcrumb/dpl_breadcrumb.module
@@ -189,7 +189,26 @@ function dpl_breadcrumb_node_presave(Node $node): void {
 }
 
 /**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+function dpl_breadcrumb_taxonomy_term_insert(TermInterface $term): void {
+  _dpl_breadcrumb_taxonomy_term_post_save($term);
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_update().
+ */
+function dpl_breadcrumb_taxonomy_term_update(TermInterface $term): void {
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_breadcrumb/dpl_breadcrumb.module
```diff
@@ -186,7 +186,39 @@ function dpl_breadcrumb_node_presave(Node $node): void {
...
+  if ($term->bundle() !== $service->getStructureVid() || !$term->hasField('field_content')) {
+    return;
+  }
+
+  $contents = $term->get('field_content')->referencedEntities();
+
+  foreach ($contents as $content) {
+    if (!($content instanceof FieldableEntityInterface)) {
+      continue;
+    }
```
kasperg, line 218: Consider if this check is necessary.

As I see it `referencedEntities()` will return `EntityInterface[]` and `EntityInterface` declares the `save()` method.

kasperg (approved):
:+1: LGTM

It seems as if we could also replace some of the logic here with [the Corresponding Entity References module](https://www.drupal.org/project/cer).

PR #1307, opened 2024-06-26
Revert "Limit number of items per purge to 30"

kasperg:
After introducing this change we have seen the frequency of MySQL gone away errors go up.

We no longer think that the purge queue batch size is the cause of these problems so to reduce the problem we revert the change and return to the default value of 100.

Reverts danskernesdigitalebibliotek/dpl-cms#1274


spaceo (approved):
LGTM 👍

PR #1312, opened 2024-06-27
Fixes, related to the RelatedContent inspiration slider. DDFFORM-907

rasben:
https://reload.atlassian.net/browse/DDFFORM-907

**RelatedContent: Remove DateFallback.**

After trials in reality, it has become clear that the RelatedContent
should never have a datefallback - if it cannot find relevant data, it
makes no sense to just find any bundle of content.

-------

**Inspiration slider: Don't look up, based on branches.**

After testing in the real world, it has become clear that
branch-filtering is way too broad for the automatic slider that shows up
on articles and events.
We still leave the logic though, as for future automatic paragraphs, it
is still useful.

------

**Remove code comment about outdated logic.**

In the past, it was true that an empty array would mean "NO tags are
okay", but now, we do an `array_filter` later on, so it doesn't matter
if we send along `NULL` or `[]`.


kasperg (approved):
:+1: LGTM

PR #1315, opened 2024-06-28
Use Drupal dynamic cache for patrons (and other authenticated users)

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-923

#### Description

We would like to improve the response times from Drupal for authenticated users. They are currently very fast when served from Varnish but this only applies for anonymous users. Drupal also has the [Dynamic Page Cache](https://www.drupal.org/docs/8/core/modules/dynamic-page-cache/overview) for authenticated users and most pages should also be very cachable for authenticated users as all personalized data is handled directly between the client and the business system. This does not seem to be the case. 

This PR tries to address this by ensuring that the Dynamic Page Cache actually works for authenticated users.

The PR is based on the observation that if you enable cacheability headers you see that pages are returned with the cache header `X-Drupal-Dynamic-Cache: UNCACHEABLE` meaning that the dynamic page cache is not in effect. After this change the header should change value to `MISS` (or `HIT`). This means that the page cache is in effect and that subsequent requests will return cached responses even for authenticated users.

To address this this PR tries to identify reasons why pages which we assume would be cacheable are not and fix these.

I suggest reviewing this on a per commit basis. [`6576516`](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1315/commits/6576516b78c77d9fc200db5acaadf38c6c020567) contains the actual fix.

Notes:

- That there may still be pages that are not cached - e.g. the event list - because we have explicitly mentioned that it should not be so.
- This does not affect the endpoint for retrieving user tokens at `/dpl-react/user-tokens` which must not be cached.

#### Screenshot of the result

Before this change an example page, the branches list at `/branches` is marked as uncachable.

Monosnap Biblioteker | DPL CMS 2024-06-28 16-02-11

After this change the same page is marked as a cache miss meaning that the result could be cached but a cached version does not exist:

Monosnap Biblioteker | DPL CMS 2024-06-28 16-03-38

Not disabling caching locally by commenting out `$settings['cache']['default'] = 'cache.backend.null';` in `web/sites/default/local.settings.php` can turn it into a cache hit:

Monosnap Biblioteker | DPL CMS 2024-06-28 16-05-17


web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php
```diff
@@ -108,4 +88,21 @@ public function getClientSecret(): string {
     return $this->getValue('client_secret');
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  public function getConfig(): array {
+    $settings = $this->configManager->getConfigFactory()->get(self::CONFIG_KEY)->get('settings');
```
rasben, line 95: What if `$this->configManager->getConfigFactory()->get(self::CONFIG_KEY)` returns NULL? Wont `->get` fail?

Maybe you could get around it with just:

```
$settings = $this->configManager->getConfigFactory()->get(self::CONFIG_KEY)?->get('settings');

Comment (kasperg) [reply web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php:95]: > What if $this->configManager->getConfigFactory()->get(self::CONFIG_KEY) returns NULL

[It does not. It will always return an `\Drupal\Core\Config\ImmutableConfig` object.](https://github.com/drupal/drupal/blob/10.1.x/core/lib/Drupal/Core/Config/ConfigFactoryInterface.php#L12-L21)

Comment (kasperg) [reply web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php:95]: You can try to run the following in a `drush php:cli` console:

```php
$config_manager = \Drupal::service('config.manager');
$config_factory = $config_manager->getConfigFactory();
$config = $config_factory->get('foo');
$config->getRawData();
```

The above will return a `\Drupal\Core\Config\ImmutableConfig` object even though `foo` is not a valid configuration entry. `$config->getRawData()` returns an empty array though.

rasben (approved):

Dresse (approved):
LGTM 👍

PR #1319, opened 2024-07-02
Add pa11y checks for more pages

kasperg:
#### Description

Currently we only check search results, a work, an article and an eventseries.

This adds support for the article list, the event list, a branch and the branch list.


kasperbirch1 (approved):
Nice! LGTM 👍

PR #1320, opened 2024-07-02
Add Lighthouse checks for more pages

kasperg:
#### Description

Currently we only check search results, a work, an article and an eventseries.

This adds support for the article list, the event list, a branch and the branch list.

Add notes about why some Lighthouse tests are currently disabled.


kasperbirch1 (approved):
Great with multiple tests. 👍

Is there or should a Jira ticket be created for todos?

kasperg:
> Is there or should a Jira ticket be created for todos?

I do not understand the question. The comments actually include ticket ids.

PR #1333, opened 2024-07-03
Cypress chromatic tests

kasperg:
#### Description

This PR adds support for running visual regression tests for the application using Chromatic based on screenshots taken through Cypress tests. This is supported by [the official Cypress integration provided by Chromatic](https://www.chromatic.com/cypress).

This is achieved by:

- [Following the official guide for integrating Cypress with Chromatic](https://www.chromatic.com/docs/cypress/)
  - Adding Chromatic and Cypress packages and plugin
  - Running Cypress with remote debugging port
  - Adding a test which takes snapshots
  - Uploading results to Chromatic
- Adding a test which tests visual regressions for example content

Our setup requires a few extra workarounds:

1. By default Cypress will take snapshots at the end of all tests. I do not think all our current Cypress tests are appropriate for visual regression testing. Consequently I have disabled this and enabled it for a single test which runs on our example content.
1. Chromatic needs to talk to more external services. This will fail due to certificate errors because traffic passes through Wiremock. Consequently we need to configure these hosts to be ignored.
1. A few Adgangsplatformen tests fail after installing the Chromatic plugin. We need to look into why this happens.
1. Chromatic fails if inline SVGs are not encoded. Consequently we need to update the design system accordingly. This happens in danskernesdigitalebibliotek/dpl-design-system#614

- Do you agree with this approach to visual regression testing?
- If you agree with the general approach I will look into the failing tesrt


.github/workflows/ci-tests.yml
```diff
@@ -166,6 +166,17 @@ jobs:
         run: task ci:reset
       - name: Run Cypress
         run: task ci:cypress
+      - name: Prepare Cypress output for Chromatic action
+        if: always()
```
rasben, line 170: Why is this line necessary?

.github/workflows/ci-tests.yml
```diff
@@ -166,6 +166,17 @@ jobs:
...
+        run: |
+          docker compose -f docker-compose.yml -f docker-compose.ci.yml run node-chrome bash -c "chmod 0777 -R /app/cypress/downloads";
+          rm tsconfig.json;
+      - name: Run Chromatic
+        uses: chromaui/action@v11
+        if: always()
```
rasben, line 176: Why is this line necessary?

cypress/e2e/adgangsplatformen.cy.ts
```diff
@@ -95,7 +95,7 @@ describe("Adgangsplatformen", () => {
     cy.get(".footer").should("not.exist");
   });
 
-  it("can register a new user - expose the right tokens for the react apps - and force logout.", () => {
+  it.skip("can register a new user - expose the right tokens for the react apps - and force logout.", () => {
```
rasben, line 98: Add a code comment here, why we're skipping / why we're still keeping it

cypress/e2e/visual-regression.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+    cy.scrollTo("bottom", { duration: 2000 });
+  });
+
+  it("renders the article list", () => {
+    cy.visit("/articles");
+    cy.scrollTo("bottom", { duration: 2000 });
+  });
+
+  it("renders an article page", () => {
+    cy.visit("/by_uuid/node/2cd0fe5e-4159-4452-86aa-e1a1ac8db4a1");
```
rasben, line 17: I would avoid using these IDs - I think they might change at some point, when we update the example content.

You could also go to the /articles page and search for the title of the content, and then click on it from there?

cypress/e2e/visual-regression.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+    cy.scrollTo("bottom", { duration: 2000 });
+  });
+
+  it("renders the event list", () => {
+    cy.visit("/events");
+    cy.scrollTo("bottom", { duration: 2000 });
+  });
+
+  it("renders an event series page", () => {
+    cy.visit("/by_uuid/eventseries/c8177097-1438-493e-8177-e8ef968cc133");
```
rasben, line 27: Same comment as articles

cypress/e2e/visual-regression.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+    cy.scrollTo("bottom", { duration: 2000 });
+  });
+
+  it("renders the branch list", () => {
+    cy.visit("/branches");
+    cy.scrollTo("bottom", { duration: 2000 });
+  });
+
+  it("renders an branch page", () => {
+    cy.visit("/by_uuid/node/dac275e4-9b8c-4959-a13a-6b9fdbc1f6b0");
```
rasben, line 37: Same comment as articles

rasben (changes requested):
This looks very promising.

I've added a few minor comments - and also:

- In the Chromatic UI, please disable "UI Review"  - we generally dont use it on our projects, and if you dont, it will always show up in pull requests
- I've noticed that the screenshots are called "Bla bla **#snapshot 1**". This makes me wonder, if something fails, if it'll call it "snapshot 2", and then things will fail as Chromatic will treat it as a whole new picture?

.github/workflows/ci-tests.yml
kasperg: The question is whether we want to run the visual regression test if the Cypress tests fail.

Even though tests fail they can still produce output so the visual regression tests can run. If they do developers know whether they are (also) facing visual regression problems.

cypress/e2e/visual-regression.cy.ts
kasperg: I think it is appropriate to use these ideas. They provide a direct link between the tests and the example content.

The UUID for an entity stays the same even if it is updated. I think that is a good thing compared to navigating through other pages.

We use the same approach for [Lighthouse](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/.lighthouserc.js) and [Pa11y](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/.pa11yci) tests. Using it here for direct entity access is consistent.

kasperg:
> In the Chromatic UI, please disable "UI Review" - we generally dont use it on our projects, and if you dont, it will always show up in pull requests

Done.

> I've noticed that the screenshots are called "Bla bla #snapshot 1". This makes me wonder, if something fails, if it'll call it "snapshot 2", and then things will fail as Chromatic will treat it as a whole new picture?

I think snapshot 1 is selected as a name when using automatic snapshots (ie. not disabling them with `disableAutoSnapshot: true`). If we go for [manually trigged snapshots using `cy.takeShopshot()`](https://www.chromatic.com/docs/cypress/targeted-snapshots/) we can choose our own names.

cypress/e2e/adgangsplatformen.cy.ts
kasperg: Agree. As I see it this is a major problem holding back this PR.

kasperg:
Closed to make room for more pull requests.

PR #1335, opened 2024-07-03
Add `<noscript>` tag to notify users of the need for JavaScript

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-820

#### Description

Certain mobile devices have JavaScript disabled by default. Enabling JavaScript is essential for full site functionality. Therefore, we want users to enable it.

image

#### Test

Turn off JavaScript in the developer console of your browser.

http://varnish.pr-1335.dpl-cms.dplplat01.dpl.reload.dk/artikler/3-gode-til-godnat-3-6-ar

### DIFF
--- a/web/themes/custom/novel/templates/layout/html.html.twig
+++ b/web/themes/custom/novel/templates/layout/html.html.twig
@@ -60,6 +60,10 @@
       {{ 'Skip to main content'|t }}
     </a>
     {{ page_top }}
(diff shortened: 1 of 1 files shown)


kasperg (changes requested):
As discussed on the meeting I do not think we can accept this in the current state.

PR #1337, opened 2024-07-04
WIP: Change event time settings to allow events to start from 00:00

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-877

#### Description
To accommodate the use of the AM/PM time format, I have adjusted the min_time setting to 12:00 AM, equivalent to 00:00 in our timezone to be able to have events before 08. Additionally, I modified the time interval to 0, enabling events to be scheduled at any time.

#### Test
http://varnish.pr-1337.dpl-cms.dplplat01.dpl.reload.dk/arrangementer

### DIFF
--- a/config/sync/recurring_events.eventseries.config.yml
+++ b/config/sync/recurring_events.eventseries.config.yml
@@ -1,6 +1,6 @@
 _core:
   default_config_hash: HuNc0wdz6BZa5IceY8HDpVfwTwqvp1rpRQeuS5szwSo
-interval: 15
+interval: 0
 min_time: '12:00am'
 max_time: '11:45pm'
 date_format: 'd/m/y - H:i'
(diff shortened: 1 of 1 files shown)


kasperg (changes requested):
Please reconsider your approach to this task:

The original request was:

1. It must be possible to choose any number of minutes on all types of events - right now it is only possible on individual events.
2. It must be possible to start an event before 8 am.
3. It must be possible for an event to run from 00:00-00:00. In that case, the text "all day" must be displayed at the event.

Part 3 has been moved to a separate task. As I see it the current change only solves part 2, but what about part 1?

kasperg (approved):

kasperbirch1:
Venter på https://reload.atlassian.net/browse/DDFFORM-947

kasperbirch1:
@rasben I have assigned you as you have overtaken the ticket. ☺️

PR #1339, opened 2024-07-05
Add `field_relevant_ticket_manager` to eventseries. DDFFORM-937

rasben:
**Also adding it to the event API, eventinstance and setting up field inheritance.**

**Move field default+inheritance to deploy hooks, run **after** `cim`**

Otherwise, the new fields haven't been introduced yet, and the update will fail.

https://reload.atlassian.net/browse/DDFFORM-937


web/modules/custom/dpl_update/dpl_update.deploy.php
```diff
@@ -35,3 +104,53 @@ function dpl_update_deploy_set_branches_not_promoted(): string {
...
+/**
+ * Set default value for all existing eventseries:field_relevant_ticket_manager.
+ */
+function dpl_update_deploy_field_relevant_ticket_manager(): string {
+  $field_name = 'field_relevant_ticket_manager';
+
+  $ids =
+    \Drupal::entityQuery('eventseries')
+      ->accessCheck(FALSE)
+      ->execute();
```
kasperg, line 117: I am surprised by the need for this deploy hook. As you have defined a default value for the field I has expected this to apply to all existing entities as well. I guess this is not the case.

Consider if you can refactor this into something like `_dpl_update_set_default_value(string $entity_type, string $field_name, $value)`.

rasben: A default value is only set on new content.
When you create a new field, all the existing entities still exist, without that field.

That's why you should never assume that a value is set on a field :)

rasben: I've created it as a helper function now.

kasperg (approved):
:+1: LGTM

PR #1347, opened 2024-07-08
DDFLSBP-756 - Delete existing webform translations

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-756

#### Description

This PR adds a new update hook to dpl_webform.install.

A previous translation has been added to the contact webform through PoEditor. This update hook is removing that translation. The reason we want to remove the translation is that it overrides the changes the editors make on the original contact webform configuration.


kasperg (approved):
:+1: LGTM

PR #1362, opened 2024-07-09
Don't display breadcrumb if the only link is current page.

rasben:
https://reload.atlassian.net/browse/DDFFORM-635


web/modules/custom/dpl_breadcrumb/src/Services/BreadcrumbHelper.php
```diff
@@ -183,6 +192,24 @@ public function getBreadcrumb(FieldableEntityInterface $entity): Breadcrumb {
...
+    if (count($links) === 1) {
+      try {
+        $breadcrumb_entity = $links[0]->getUrl()->getOption('entity');
+
+        if ($breadcrumb_entity === $entity) {
+          $breadcrumb = new Breadcrumb();
+        }
+      }
+      catch (\Exception $e) {
+        $this->logger->error("Failed checking breadcrumb solo logic. Message: %message", ["%message" => $e->getMessage()]);
```
spaceo, line 209: In my opinion I don't think logging is the right tool to verify that this logic works. 
If that was the case we should change a lot of our user interface tests with logging 😛 
Instead I think cypress test would be a proper fit. Testing the behaviour of the breadcrumb when the only link is current page and when not.

spaceo (changes requested):
I think we need a test instead of logging

web/modules/custom/dpl_breadcrumb/src/Services/BreadcrumbHelper.php
rasben: But this is not a test - it's a response to something happening that should not happen.

The reason I'm doing this, is because the methods I'm running have a catchable exception, and PHPStorm suggests that I add it to a try catch - otherwise, the whole page might fail.

I agree that the breadcrumb needs a cypress test, but I'm working on that in another PR :)

spaceo: I don't want to be a blocker, but I still think it is wrong to log a breadcrumb functionality if it breaks. It should be code that is in our hands and we can manage if it fails. It is ok to catch the exception (after a quick digging I cannot find out from where it derives), but I think it is wrong to log it. If it was depending on a remote request it would be another case.

spaceo (approved):
Still don't approve the logging though.

PR #1368, opened 2024-07-10
DDFLSBP-695 - Allow uppercase and dashes in webform element key patterns

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-695

#### Description

This PR updates webform settings to allow for both uppercase letters and dashes in element key patterns. This will give the editors more freedom when creating element keys. This is useful in cases where specific element keys needs to have a specific name that includes either dashes or uppercase letters.

### DIFF
--- a/config/sync/webform.settings.yml
+++ b/config/sync/webform.settings.yml
@@ -125,7 +125,7 @@ form:
   filter_category: ''
   filter_state: ''
 element:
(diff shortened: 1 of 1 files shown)


LasseStaus (approved):
LGTM 👍  

However i haven't touched the webforms, so i can't give input on wether it can have any side effects allowing them to do this.

PR #1374, opened 2024-07-10
DDFLSBP-755 - Update spam protection using Honeypot and AntiBot modules

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-755

#### Description

Some libraries have reported spam submissions on webforms, which indicate that the existing implementation of the [Honeypot](https://www.drupal.org/project/honeypot) module is not enough to protect against spambots. 

This PR tries to remedy that by doing 2 things:
1 - Enable honeypots time restriction protection on all webforms.
2 - Adds and enable the [Antibot](https://www.drupal.org/project/antibot) module on all webforms.

The reasoning behind this approach is:

1. We have drawn inspiration from the old DDB CMS (ding2), where a similar setup using both spam protection modules has shown great results. 

2. Neither of the modules should be detectable by the end-users, meaning the experience when filling out forms on the sites, should be the same. * 

3. Both modules is easily managed through the [webform](https://www.drupal.org/project/webform) third-party configuration, which makes sure that both spam modules works on all webforms.

4. AntiBot per default is also setup to be enabled on Drupals user login form and other Drupal specific forms. 

* One notion to this is, that AntiBot will make sure that JavaScript needs to be enabled for the webforms to work. If not, the webform will not be shown and instead a message informing the user to enable JavaScript in the browser, before using the webform. For most users this will not be a problem. 

#### Screenshot of the result

Image showing the message, informing users to turn on JavaScript in order to use the webform:
Screenshot 2024-07-10 at 15 53 57

I haven't found it necessary to go into details on how the 2 specific modules work in this PR description. If more info is needed, please follow the links to the 2 modules found in this description, as their module info pages go into details on how they ensure spam protection.


composer.json
```diff
@@ -90,15 +90,16 @@
         "composer/installers": "1.12.0",
         "cweagans/composer-patches": "1.7.3",
         "danskernesdigitalebibliotek/cms-api": "*",
-        "danskernesdigitalebibliotek/dpl-design-system": "2024.28.0",
-        "danskernesdigitalebibliotek/dpl-react": "2024.28.0",
+        "danskernesdigitalebibliotek/dpl-design-system": "^2024.28",
+        "danskernesdigitalebibliotek/dpl-react": "^2024.28",
```
kasperg, line 94: I wonder what is up with this change. It does not seem to be related to the change at hand.

kasperg (approved):
:+1: LGTM

composer.json
Dresse: I had merge conflicts in composer.lock, which I solve by merging develop into my branch, accept incoming changes. After this I run a raycast-script which looks like this:

`task dev:cli -- composer remove danskernesdigitalebibliotek/dpl-react danskernesdigitalebibliotek/dpl-design-system \
&& task dev:cli -- composer require danskernesdigitalebibliotek/dpl-react danskernesdigitalebibliotek/dpl-design-system`

Dresse:
I am closing this PR as we have added the fix as a Hotfix here: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1454

PR #1381, opened 2024-07-12
Fix template for filtered event list

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-945

#### Description

After reintroducing related content filtered event list are rendered without any content. This is because the reintroduction changed variable names and there is no variable view_has_results which matches the current template.

The @todo in the template marks the template for deletion after reintroducing related content. However this does not work as expected. It will render the list but also the name of the paragraph.

Consequently to address this we only render the view. This makes it work the same way as
cms/web/themes/custom/novel/templates/paragraphs/paragraph--card-grid-automatic.html.twig

#### Screenshot of the result

Before:

Monosnap Test automatic | DDF School site — Privat browsing 2024-07-12 14-10-08

After:

Monosnap Edit Page Test automatic | DDF School site | Logged in 2024-07-12 14-12-28


rasben (approved):
derp, tak.

kasperg:
This has been superseded by #1389.

PR #1386, opened 2024-07-13
DDFLSBP-693 - Remove language warning when editing webforms

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-693

#### Description

Editors with their default language set to danish, kept getting a warning message when editting webforms. The warning said that they were editting the original english version of the webform. We do not care about the warning, since that is the expected behaviour. To avoid showing the warning, we simply set the langcode to danish on the webform edit forms.

#### Other comments

Before going with this solution, I tried to simply remove the language warning from the form like this `unset($form['langcode_message'])` in a form alter hook. 

But I found out that `webform` module own form alter hook was running after my form alter hook in our `dpl_webform` module, which meant whatever I did in my hook wouldn't matter as it was getting "overwritten" by the `webform` form alter hook. 
So I tried changing the order of which hook was running last by using [hook_module_implements_alter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_module_implements_alter/10), but I found out that the `webform` module was also using the same hook, and thus I ended up with the same problem as before.

Lastly I decided to try to change the weight of `dpl_webform` module to always be higher than the `webform` module. I did that using an update hook using the function [module_set_weight](https://api.drupal.org/api/drupal/core%21includes%21module.inc/function/module_set_weight/9). Even though I changed the module weight to be higher than `webform`, I couldn't make my form alter function be the last to run. 

In the end I didn't want to use more time on it, and instead I came up with the solution presented in this PR.

### DIFF
--- a/web/modules/custom/dpl_webform/dpl_webform.module
+++ b/web/modules/custom/dpl_webform/dpl_webform.module
@@ -112,3 +112,20 @@ function dpl_webform_form_webform_handler_form_alter(array &$form, FormStateInte
   $form['settings']['from']['from_mail']['from_mail']['#default_value'] = '[site:mail]';
   $form['settings']['from']['from_name']['from_name']['#default_value'] = '';
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function dpl_webform_form_webform_edit_form_alter(array &$form, FormStateInterface &$form_state, string $form_id): void {
+
+  /** @var \Drupal\webform\WebformInterface $webform */
+  $webform = \Drupal::routeMatch()->getParameter('webform');
+
+  // Editors with their default language set to danish, kept getting
+  // a warning message when editting webforms. The warning said that
+  // they were editting the original english version of the webform.
+  // We do not care about the warning, since that is the expected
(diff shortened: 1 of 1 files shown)


Dresse:
> Have you double tested deploying a site with this? I am just worried that something could conflict with our configuration management when you change the language like that. Maybe there is nothing to worry about?

I have tested this on the associated environment to this PR (https://varnish.pr-1386.dpl-cms.dplplat01.dpl.reload.dk/) 
1. Logging in with an user with the "Editor" role. 
2. Making changes to the webform and save the webform.
3. Check that the change is reflected in the UI where the webform is used (e.g /kontakt page).
4. Logging in with user 1.
5. Check that the change done previously as an Editor is also reflected in the english version.
6. Making a change as user 1 and saving the webform.
7. Check that the change is reflected in the UI where the webform is used (e.g /kontakt page).
8. Logging in with an user with the "Editor" role.
9. Check that the change done previously as user 1 i also reflected in the danish version.

From the above test, I assess that this is not a problem. However, if someone wants to take another crack at solving it less "hacky", I would support this. 👍

PR #1388, opened 2024-07-15
Delete faulty translation of `webform.settings.yml`

rasben:
Due to some faulty, prior translation, the danish version of this file has duplicate keys, which causes major issues on the site when loading the webform settings.

We do not know what has caused this, but essentially, the contact forms (and other webforms) fail because the danish translation has two keys that are duplicates. (Postnummer x2)

Tada:

https://varnish.pr-1388.dpl-cms.dplplat01.dpl.reload.dk/kontakt

### DIFF
--- a/web/modules/custom/dpl_webform/dpl_webform.install
+++ b/web/modules/custom/dpl_webform/dpl_webform.install
@@ -55,3 +55,21 @@ function dpl_webform_update_10002(): string {
 
   return 'All translations for the webform ' . $webform_id . ' have been deleted.';
 }
+
+/**
+ * Delete translation of webform.settings.yml.
+ *
+ * Due to some faulty, prior translation, the danish version of this file
+ * has duplicate keys, which causes major issues on the site when loading
+ * the webform settings.
+ */
+function dpl_webform_update_10003(): string {
+  $config_id = 'webform.settings';
+  $language = 'da';
+
+  /** @var \Drupal\language\Config\LanguageConfigFactoryOverride $language_config_factory_override */
+  $language_config_factory_override = \Drupal::service('language.config_factory_override');
+  $language_config_factory_override->getOverride($language, $config_id)->delete();
+
+  return "DA translations for the $config_id have been deleted.";
+}


web/modules/custom/dpl_webform/dpl_webform.install
```diff
@@ -55,3 +55,21 @@ function dpl_webform_update_10002(): string {
...
 }
+
+/**
+ * Delete translation of webform.settings.yml.
+ *
+ * Due to some faulty, prior translation, the danish version of this file
+ * has duplicate keys, which causes major issues on the site when loading
+ * the webform settings.
+ */
+function dpl_webform_update_10004(): string {
```
kasperg, line 66:
```suggestion
function dpl_webform_update_10003(): string {
```

It seems like this jumps an update number?

web/modules/custom/dpl_webform/dpl_webform.install
```diff
@@ -55,3 +55,21 @@ function dpl_webform_update_10002(): string {
...
+  $config_id = 'webform.settings';
+  $language = 'da';
+
+  /** @var \Drupal\language\Config\LanguageConfigFactoryOverride $language_config_factory_override */
+  $language_config_factory_override = \Drupal::service('language.config_factory_override');
+  $language_config_factory_override->getOverride($language, $config_id)->delete();
```
kasperg, line 72: I do not know if it makes sense to create a reusable function to address this. Perhaps next time.

kasperg (changes requested):

web/modules/custom/dpl_webform/dpl_webform.install
rasben: Not on purpose - fixed.

web/modules/custom/dpl_webform/dpl_webform.install
rasben: I mean, hopefully there won't be a next time, because it should never be necessary to do this

kasperg (approved):

kasperg:

PR #1389, opened 2024-07-15
Delete unused views + templates, replaced by RelatedContent.

rasben:
Deleting the views and the views template is just clean up: It has no functional difference.

However, the paragraph template needs to be deleted, for the RelatedContent system to actually work.


kasperg (approved):
:+1: Should this be merged int #1381 for additional testing?

PR #1390, opened 2024-07-15
Prevent updating eventseries schedule from failing [DDFFORM-947]

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-947

#### Description

When updating the schedule for an event series the recurring events
module shows a summary of the change and asks the user to confirm it
as this will recreate all instances.

In this situation we want to hide the default form actions as they
make little sense in this situation even though they are fully
functional.

Until now we have achieved this by removing the form elements from the
form.

After updating the Gin administration theme and Drupal Core from 10.1
to 10.3 this approach no longer works. Gin expects the actions to be
available and throws a hard PHP error if they are not.

Attempt to address this using alternate approaches using the Drupal
Form API e.g. #access=FALSE or #disabled=TRUE has not worked due to
the rather complex form manipulations of Gin.

Instead we switch to a CSS-based approach which hides the form
elements instead. This should not make a huge difference compared to
the original approach.

Gin also does interesting things to the DOM structure so the resulting
CSS selector also becomes a bit complex.

In the process I also move the handling of the improvement from the event module to the admin module. It makes just as much sense to keep them there.

The module also already has CSS files which we may need to improve
handling of events for admins.

#### Screenshot of the result

Monosnap Edit default Event Fernisering Moderne Dans | DPL CMS | Logged in 2024-07-15 13-14-06

A quick test shows that we could achieve the same by patching the `recurring_events` module and preventing access to `actions` [immediatly as a schedule change is detected and a confirmation form is added](https://git.drupalcode.org/project/recurring_events/-/blob/2.0.x/src/Form/EventSeriesForm.php?ref_type=heads#L236).

Monosnap Edit default Event Fernisering Moderne Dans | DPL CMS | Logged in 2024-07-15 13-14-06

### DIFF
--- a/web/modules/custom/dpl_admin/assets/dpl_admin.css
+++ b/web/modules/custom/dpl_admin/assets/dpl_admin.css
@@ -175,3 +175,12 @@
     max-width: 100%;
   }
 }
+
+/**
+ Hide default action button on confirm form.
+ Editors should focus on the date-related changes.
+ Selector is convoluted due to Gin DOM structure.
+*/
+body:has(.eventseries-form--confirm) .form-actions {
+  display: none;
+}
--- a/web/modules/custom/dpl_admin/dpl_admin.module
+++ b/web/modules/custom/dpl_admin/dpl_admin.module
@@ -325,3 +325,33 @@ function dpl_admin_toolbar_alter(array &$items): void {
     ],
   ];
 }
+
+/**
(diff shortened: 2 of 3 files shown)


rasben (approved):
Seeing as it's admin, it's fine, but just consider that `:has()` is not supported in IE

kasperg:

PR #1397, opened 2024-07-15
Delete faulty webform translation

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFNEXT-354

#### Description

It turns out the way configuration translation is handled by the
config_translation module is:

1. Import the translation strings from the remove source into Drupal translation
   strings.
2. Import translation strings into configuration.

This causes problems when we want to remove a local faulty translation as in
dpl_webform_update_10003(). Even though we have removed the translated
configuration (1) it still lingers as a translation string (2). This will be i
ncluded in the translated configuration when rerunning the configuration
translation process.

Consequently: To remove the faulty translation for webform.settings entirely
we have to remove it from the local string translations as well.

The problematic part is in the test names part of the configuration so we can
query based on this.

### DIFF
--- a/web/modules/custom/dpl_webform/dpl_webform.install
+++ b/web/modules/custom/dpl_webform/dpl_webform.install
@@ -5,6 +5,9 @@
  * DPL webform install file.
  */
 
+use Drupal\drupal_typed\DrupalTyped;
+use Drupal\locale\StringStorageInterface;
+
 /**
  * Add all webforms to config ignore.
  */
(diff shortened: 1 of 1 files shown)


rasben:
It appears to work.

Here's my process, that I've done locally:

- git checkout main
- git pull
- lagoon -p canary -e main ssh --command "drush sql:dump" > ../canary.sql
- dce drush sql-cli < ../canary.sql
- dce drush cr

- curl -I http://dpl-cms.docker/kontakt\?x
`HTTP/1.1 500 500 Service unavailable (with message)`

- git checkout remove-faulty-webform-translation

- [(Added the changes i've pushed in a seperate commit)](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1397/commits/709d561cbf0deb96df2e1c9d20bfa620a85105c5)

- dce drush deploy
- dce drush dpl_po:import-remote-config-po da https://danskernesdigitalebibliotek.github.io/dpl-cms/translations/da.config.po

- dce drush cr

- curl -I http://dpl-cms.docker/kontakt\?xx
`HTTP/1.1 200 OK`

rasben (approved):

PR #1406, opened 2024-07-16
Fix failing admin event search with custom, upstream patch. DDFFORM-872

rasben:
Without this patch, trying to do any kind of search on admin/content/eventseries fails with a WSOD.
Turns out it was an issue with the query, in reccurring_events module.

https://reload.atlassian.net/browse/DDFFORM-872

https://www.drupal.org/project/recurring_events/issues/3461740


kasperg (approved):
:+1: LGTM

I have checked the issue and the patch and to me it makes sense.

PR #1407, opened 2024-07-17
Allow sites to choose between DKK or EUR currency for events.

rasben:
The formatting is taken care of in PriceFormatter, that we already use in the frontend.
Also makes sure to expose the correct currency in the API. DDFFORM-752

https://reload.atlassian.net/browse/DDFFORM-752


rasben:
The unit tests fail:

20) dpl_event\tests\src\Unit\PriceFormatterTest::testPriceRangeFormatting with data set #4 (array(10, 20, 30), '10 - 30 DKK')
**Error: Call to a member function get() on null**

My guess is that I'm missing something when I include the conif factory in the test, but I cant wrap my head around how to fix it, so I'd like some guidance with that :)

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -15,11 +17,44 @@
...
+    protected ConfigFactoryInterface $configFactory,
+  ) {
+    $config = $this->configFactory->get(SettingsForm::CONFIG_NAME);
+
+    $currency = $config->get('price_currency') ?? $this->currency;
+    $this->currency = $currency;
+
+    switch ($currency) {
+      case 'DKK':
+        $this->currencyPrefix = NULL;
```
spaceo, line 48: The property is already initiated as NULL. I do not see any reason to setting the value again.

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -15,11 +17,44 @@
...
+        $this->currencySuffix = $this->translation->translate(' DKK', [], ['context' => "DPL event"])->render();
+        break;
+
+      case 'EUR':
+        $this->currencyPrefix = '€ ';
+        $this->currencySuffix = NULL;
```
spaceo, line 54: The property is already initiated as NULL. I do not see any reason to setting the value again.

web/modules/custom/dpl_event/dpl_event.services.yml
```diff
@@ -8,7 +8,7 @@ services:
     arguments: ["eventinstance"]
   dpl_event.price_formatter:
     class: Drupal\dpl_event\PriceFormatter
-    arguments: ["@string_translation"]
+    arguments: ["@string_translation", "@config.factory"]
```
spaceo, line 11: Instead of injecting the string translation as a dependency you can alsu use the `StringTranslationTrait`on your service. I think it is more neat. Less work.

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -35,10 +70,7 @@ public function formatPrice(string $price_string): string {
       // Format the numeric part of the price using formatRawPrice.
       $formatted_price = $this->formatRawPrice($price_string);
 
-      // Add the 'kr.' suffix for now.
-      // For multi-currency support this should be replaced by a configurable
-      // suffix and appropriate separators.
-      return $formatted_price . ' kr.';
+      return "{$this->currencyPrefix}$formatted_price{$this->currencySuffix}";
```
spaceo, line 73: This is a matter of taste. But if values are being concatenated like this I prefer to work with arrays and the in the end use implode.
Like:
```php
$price[] = $this->currencyPrefix;
$price[] = $this->formatRawPrice($price_string);
$price[] = $this-> currencySuffix;
return implode($price);
```
or:
```php
return implode([
  $this->currencyPrefix;
  $this->formatRawPrice($price_string);
  $this-> currencySuffix
]);
```

web/modules/custom/dpl_event/src/PriceFormatter.php
```diff
@@ -67,22 +99,32 @@ public function formatPriceRange(array $prices): string {
...
+    $highest_price_raw = max($filtered_prices);
 
     // Construct the price range string.
     if ($has_free_price) {
+      // Get the price, with prefix and suffix, as we only use this one.
+      $highest_price_formatted = $this->formatPrice((string) $highest_price_raw);
+
       // Only display "Free" and the highest price.
-      return $this->translation->translate("Free") . ' - ' . $formatted_highest_price . ' kr.';
+      return $this->translation->translate("Free") . " - $highest_price_formatted";
```
spaceo, line 110: Again a matter of taste, but I thin `sprintf` is easier to read than using the `.`string concatenation:
```php
      return sprintf("%s - %s", $this->t("Free"), $highest_price_formatted);

File: web/modules/custom/dpl_event/src/PriceFormatter.php [hunk 6]
```diff
@@ -35,10 +70,7 @@ public function formatPrice(string $price_string): string {
       // Format the numeric part of the price using formatRawPrice.
       $formatted_price = $this->formatRawPrice($price_string);
```
spaceo, line 71: This is a bit off topic because it is not a part of this PR. But I do not understand the concept of `formatRawPrice. If we are dealing solely with numbers I think the input and the return types should reflect that. Either with integers or float.
Another thing that I am thinking of (if I dare to say it :)) is why not using a library for handling prices? Are there any special circumstances that prevents that?

spaceo (changes requested):
I have some questions...

web/modules/custom/dpl_event/dpl_event.services.yml
rasben: I'd prefer not to change the existing stuff in this PR - If we want to do that in the future, i think we should check all the places that use it and do it here :)

web/modules/custom/dpl_event/src/PriceFormatter.php
rasben: I respect your preference, but I prefer my method - i feel like it's more logical and readable :)

web/modules/custom/dpl_event/src/PriceFormatter.php
rasben: I tried, but PHPStorm informs me that i need to introduce a whole package, which i think is overkill :)

```
phpstan: Function sprintf is unsafe to use. It can return FALSE instead of throwing an exception. 
Please add 'use function Safe\sprintf;' at the beginning of the file to use the variant provided by the 'thecodingmachine/safe' library.

web/modules/custom/dpl_event/src/PriceFormatter.php
rasben: The `raw` might mislead you a bit;
It's not a raw number, but a formatted number. E.g. 1000000 becomes 1,000,000 in danish.
The reason we call it raw here, is because it doesnt have the "DKK".

I'm not the original author of the file, so I cant tell you why we arent using a library - but, I would imagine that there's a good reason :)

web/modules/custom/dpl_event/src/PriceFormatter.php
spaceo: You just need to do like this:
https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/b5565c74fd2dfe3426c69e37641bdea73b38dfd2/web/modules/custom/dpl_url_proxy/src/Form/ProxyUrlConfigurationForm.php#L10

spaceo (approved):
LGTM 👍 🎉 
I have one comment that you can reconsider.

PR #1410, opened 2024-07-18
Varnish Purge Cypress test

kasperg:
#### Description

This adds a test to ensure that Varnish caching and purging is working as expected.

This builds upon the basics introduced in #1408 and ensures that the changes in #1409 are working as expected.


kasperg:
This is now part of #1409

PR #1412, opened 2024-07-18
Show campaigns for editors

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFNEXT-368

#### Description

Currently when a editor works with campaigns she cannot test them in practice by doing a search because the campaign will never be displayed. The request returns a 403 access denied exception. This happens because the request contains an authentication cookie but the resource is not configure to accept this.

Fix this by updating the permissions. Also implement a test to show the problem and that the change works.

This PR is intentionally based on #1409. While they are functionally unrelated #1409 has some authentication based refactorings that we would like to use here as well.


web/modules/custom/dpl_campaign/src/EventSubscriber/ApiRouteSubscriber.php
```diff
@@ -0,0 +1,36 @@
...
+use Drupal\Core\Routing\RouteSubscriberBase;
+use Symfony\Component\Routing\Route;
+use Symfony\Component\Routing\RouteCollection;
+
+/**
+ * Remove X-CSRF protection for campaign match API.
+ *
+ * This reduces the complexity of calling the API as the caller no longer has
+ * to obtain a token before making the actual request.
+ */
```
spaceo, line 16: Consider to make the comment more generic. Wouldn't it be plausible that we get future routes that need to bypass the X-CSRF protection?

spaceo (approved):
LGTM 👍 
One question...

web/modules/custom/dpl_campaign/src/EventSubscriber/ApiRouteSubscriber.php
kasperg: We will do that some other time.

PR #1413, opened 2024-07-18
Only run Lighthouse tests once per url

kasperg:
#### Description

Use only 1 run to improve test speed. Multiple runs to handle caching should not be required while we have downgraded performance from error to warning.


spaceo (approved):
LGTM 👍

PR #1420, opened 2024-07-19
Increase wait time for Lagoon PR environment deployments to 20 mins

kasperg:
#### Description

After enabling example content on environments deployments take even longer. The current situation leads to false negatives because the wait times out.

Increase time to 20 minutes to avoid this.


spaceo (approved):
LGTM 👍

PR #1432, opened 2024-07-25
Setup password policies, for admins and non-admins. DDFFORM-823

rasben:
Using the `password_policy` module, we can create different configurations for different roles, to make sure passwords live up to the DDF IT-security standard.
This commit sets up a policy for non-admins (Editor, Mediator, External), and admins (Administrators, Local Administrators).

Non-admins have a simple requirement of minimum 8 characters, with a mix of letters, numbers and characters.
Admin users have a more complicated requirement of minimum 12 characters, and their passwords require a reset every 90 days.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-823

------

Notice - I've made sure not to include patrons in any of these policies, and also the existing passwords have not been expired. There is a "force everyone to get a password reset" as part of the module, but the ticket explicitliy mentions that this is not something we should do - and I'd also worry that this could cause issues with patrons.


spaceo (approved):
LGTM 👍

PR #1438, opened 2024-07-30
Editorial Search DDFFORM-678 

rasben:
## Create editorial search view + index. DDFFORM-678

- Creating a search API index, that combines eventseries and content
- Set up `list_teaser` view mode for eventseries
- Simplify the hook suggestions, so we can use a single template for
categories in the list teaser
- Setting up the view, that is displayed on /web/search
- Update search API, to allow partial matches when doing fulltext
  searches

## Introduce dpl_search module, tweaking editorial search. DDFFORM-678

- Add dpl_search module that tweaks the editorial view, adding dynamic
title
- Also add link to the material search with the same query string
- Replace the search bar in the header, when on the editorial search
page.

## Add a link to editorial search on material search page. DDFFORM-678

When on the material search, we can look up the editorial (web) search
using the same fulltext query, and see if there are any results.
If there are results, we add a link.
This link should also be parsed along to the react app.
This PR prepares that, sending it along as `web-search-link-text` and
`web-search-link-url`.

React: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1367
Design: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/699

# Test 
https://varnish.pr-1438.dpl-cms.dplplat01.dpl.reload.dk/search?q=dans

#### Note
In relation to the task, I started implementing our design system for lists with Drupal content. It turned out to be a bigger task than expected, so this ticket has been created to avoid a pull request that is too large.
https://reload.atlassian.net/browse/DDFFORM-953


rasben:
You can test it by going to the test site and trying to do a search on "dans" in materials.
You can also access it directly on `/search/web?search=dans`

rasben:
Also, you might have to move the search icon in the header.

web/themes/custom/novel/templates/dpl-react-app--favorites-list.html.twig
```diff
@@ -1,6 +1,7 @@
 <div class="dpl-react-app-container--page">
     <div {{ attributes }}>
         <div class="card-list-page">
+            {# Todo: use content-list-page__heading #}
```
xendk, line 4: TODO?

xendk (changes requested):
There's some todos.

web/themes/custom/novel/templates/dpl-react-app--favorites-list.html.twig
kasperbirch1: They are here to help when we start working on https://reload.atlassian.net/browse/DDFFORM-953

xendk: Ah, okay. In general I'm not as opposed to todos as others, but if there's already an issue it would be nice mentioning it in the comment, as it gives anyone stumbling across it more context.

xendk (approved):

PR #1444, opened 2024-08-01
Event screen name

xendk:
#### Link to issue

[KBHBIB-14](https://reload.atlassian.net/browse/KBHBIB-14)

#### Description

Adds a screen name to events, for external screen systems.

Screens are simply taxonomy terms. Appropriate fields have been added to event series and instances, and the `/api/v1/events` API extended with an extra field.

A setting has been added to `/admin/config/dpl-event/settings` to enable the feature (disabled per default). When the feature is disabled, the vocabulary and the fields on events are hidden, but the data isn't deleted. 

#### Screenshot of the result

image
image

[KBHBIB-14]: https://reload.atlassian.net/browse/KBHBIB-14?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/sync/core.entity_view_display.eventinstance.default.card.yml
+++ b/config/sync/core.entity_view_display.eventinstance.default.card.yml
@@ -16,6 +16,7 @@ dependencies:
     - field.field.eventinstance.default.field_event_state
     - field.field.eventinstance.default.field_event_title
     - field.field.eventinstance.default.field_external_admin_link
+    - field.field.eventinstance.default.field_screen_names
     - field.field.eventinstance.default.field_tags
     - field.field.eventinstance.default.field_teaser_image
     - field.field.eventinstance.default.field_teaser_text
@@ -75,6 +76,22 @@ content:
       format: default
       format_custom_false: ''
       format_custom_true: ''
+    third_party_settings: {  }
+    weight: 60
+    region: content
+  event_screen_name:
+    type: entity_reference_label
+    label: visible
+    settings:
+      link: true
+    third_party_settings: {  }
+    weight: 60
+    region: content
+  event_screen_names:
+    type: entity_reference_label
+    label: visible
+    settings:
+      link: true
     third_party_settings: {  }
     weight: 60
     region: content
@@ -144,6 +161,7 @@ hidden:
   field_event_state: true
   field_event_title: true
   field_external_admin_link: true
+  field_screen_names: true
   field_tags: true
   field_teaser_image: true
   field_teaser_text: true
--- a/config/sync/core.entity_view_display.eventinstance.default.card.yml
+++ b/config/sync/core.entity_view_display.eventinstance.default.card.yml
@@ -79,17 +79,9 @@ content:
     third_party_settings: {  }
     weight: 60
     region: content
-  event_screen_name:
-    type: entity_reference_label
-    label: visible
-    settings:
-      link: true
-    third_party_settings: {  }
-    weight: 60
-    region: content
   event_screen_names:
     type: entity_reference_label
-    label: visible
+    label: above
     settings:
       link: true
     third_party_settings: {  }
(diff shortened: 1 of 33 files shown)


web/modules/custom/dpl_update/dpl_update.deploy.php
```diff
@@ -250,3 +250,10 @@ function dpl_update_deploy_field_relevant_ticket_manager(): string {
 function dpl_update_deploy_update_term_url_aliases(): string {
   return _dpl_update_generate_url_aliases('taxonomy_term');
 }
+
+/**
+ * Link new event_screen_names inheritance on eventinstances.
+ */
+function dpl_update_update_screen_name_field_inheritance(): string {
+  return _dpl_update_field_inheritance('event_screen_names');
```
kasperg, line 258: :+1: for remembering to add this!

web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -287,3 +290,50 @@ function dpl_event_gin_content_form_routes() : array {
...
+ * Implements hook_entity_query_TAG_alter().
+ *
+ * Alter vocabulary queries to hide the vocabulary when the screens feature
+ * is disabled.
+ */
+function dpl_event_entity_query_taxonomy_vocabulary_alter(QueryInterface $query): void {
+  if (!\Drupal::config(SettingsForm::CONFIG_NAME)->get('enable_screen_name')) {
+    $query->condition('vid', 'screen_name', '<>');
+  }
+}
```
kasperg, line 304: Consider elaborating on how you ended up with this solution.

To be this approach came up as a bit of a surprise. I would have expected it to be handed by altering the access control system but I guess this will work too. I do not see us using hook query alter much.

I would have expected this to be mentioned in the commit comment or in an inline comment.

config/sync/taxonomy.vocabulary.screen_name.yml
kasperg, line 1: It might be relevant to use something like [Flat taxonomy](https://www.drupal.org/project/flat_taxonomy) here but that might be something we look into at another time.

config/sync/core.entity_view_display.eventinstance.default.card.yml
```diff
@@ -144,6 +161,7 @@ hidden:
   field_event_state: true
   field_event_title: true
   field_external_admin_link: true
+  field_screen_names: true
```
kasperg, line 164: Please consider what the entry `event_screen_name` is about.

As far as I can tell the new field is called `event_screen_names` or `field_screen_names`.

kasperg (approved):
:+1: Lots of good stuff here!

I have some comments for you to consider before you merge.

web/modules/custom/dpl_event/dpl_event.module
xendk: > To be this approach came up as a bit of a surprise. I would have expected it to be handed by altering the access control system but I guess this will work too.

It *is* altering the access control system. The vocabularies page use a `EntityListBuilder` which relies on the entity access system (the famous `->accessCheck(TRUE)`, which uses query tags.

The documentation leaves something to be desired, but after digging around and reading code I think I ended up at https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21entity.api.php/function/hook_entity_access/11.x which points towards `hook_query_TAG_alter()`... Which is the way to actually affect the queries.

I'll expand the comment a bit.

config/sync/taxonomy.vocabulary.screen_name.yml
xendk: I'm seem to remember that Nino specificially would like them to be hieratical.

config/sync/core.entity_view_display.eventinstance.default.card.yml
xendk: Looks like cruft from earlier attempt. Tried removing them by hand and things still work. But I'm annoyed that Drupal will just let them be when `cim`/cex`ing.

PR #1447, opened 2024-08-01
Unset beta setting for GIN, so save button is always visible.

rasben:
As part of the module update, a BETA function for the gin admin theme was enabled, that placed save buttons in the top, as sticky. This however breaks the experience when editing views (and maybe other also places).


Dresse (approved):
LGTM 👍

PR #1448, opened 2024-08-01
Re-factor and fix admin list of users. DDFFORM-952

rasben:
The module `administerusersbyrole` introduced its own version of /admin/people, but this view has a bug, meaning editors cannot see administrators.
I've disabled this view, and re-enabled and tweaked the core view.

https://reload.atlassian.net/browse/DDFFORM-952

You can test it, by logging in as `local_administrator`, and see that https://varnish.pr-1448.dpl-cms.dplplat01.dpl.reload.dk/admin/people shows `administrator`

### DIFF
--- a/config/sync/views.view.user_admin_people.yml
+++ b/config/sync/views.view.user_admin_people.yml
@@ -1,9 +1,10 @@
 uuid: 6513da2f-8a2a-4aa6-ad90-0b472a9cb9e9
 langcode: en
-status: false
+status: true
 dependencies:
   config:
     - field.storage.user.field_author_name
+    - user.role.patron
   module:
     - user
 _core:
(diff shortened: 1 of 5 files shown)


config/sync/views.view.user_admin_people.yml
```diff
@@ -134,11 +199,11 @@ display:
...
           plugin_id: field
-          label: Name
+          label: 'Name & Email'
           exclude: false
           alter:
-            alter_text: false
-            text: ''
+            alter_text: true
+            text: '{{ field_author_name__value }} <br><small>{{ mail }}</small>'
             make_link: false
```
Dresse, line 207: Nice job adding both authorname and email. 👍

config/sync/views.view.user_admin_people.yml
```diff
@@ -203,7 +268,7 @@ display:
           exclude: false
           alter:
             alter_text: false
-            text: ''
+            text: '{{ status__value }} '
```
Dresse, line 271: What did the status row show before?

config/sync/views.view.user_admin_people.yml
```diff
@@ -815,24 +827,23 @@ display:
...
           value:
-            min: ''
-            max: ''
-            value: '0'
+            patron: patron
           group: 1
           exposed: false
           expose:
-            operator_id: '0'
+            operator_id: ''
```
Dresse, line 846: Can you explain these changes to me? Its a bit unclear.

config/sync/views.view.user_admin_people.yml
rasben: This is not something I actively changed, so I dont know :')
I could just see that it works now :D

config/sync/views.view.user_admin_people.yml
rasben: I'm filtering all users with the role "patron" out of the view :)

Dresse: Ah, makes sense to not show those. 👍

PR #1449, opened 2024-08-02
Open links in new window - DDFFORM-810

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-810

#### Description
This pull request adds the capability to open links in a new tab. To support this functionality, I enabled two modules due to the presence of links both as fields and within the text editor:

- `editor_advanced_link` for the text editor (WYSIWYG / CKEditor)
- `dpl_link`: A custom module that adds a widget for the link field, including a checkbox to open links in a new tab/window.

The paragraphs that use links and should be able to open in a new tab are:
- Hero
- Banner
- Card grid (manual and automatic)

Screenshot of the result

#### Screenshot of the result
image

image

#### Test
https://varnish.pr-1449.dpl-cms.dplplat01.dpl.reload.dk/test-links


xendk (approved):
:+1:

PR #1452, opened 2024-08-06
Remove empty module files

xendk:
Remove unneeded module files.

What's the plan for `dpl_example_breadcrumb`? When is it going to be removed completely?


kasperbirch1 (approved):
I think it's good to clean up. 👍 

In relation to `dpl_example_breadcrumb` it's probably @rasben  who has to answer that.

PR #1464, opened 2024-08-08
Set develop dependencies to use develop builds

spaceo:
#### Description

As long as we are branching out from develop it makes sense to use develop builds from the dependencies since they will be updated regularly.


Dresse (approved):
LGTM

PR #1474, opened 2024-08-13
Use `.uri` for links to avoid proxy problems where `:` are changed to `%3A` - DDFFORM-811

kasperbirch1:
#### Link to Issue
https://reload.atlassian.net/browse/DDFFORM-811

#### Description
This pull request addresses the proxy issue where `:` is changed to `%3A` in external links. To resolve this, I have used `.uri` instead of `['#url']`.

This adjustment applies to external links, including:
- Hero
- Banner
- Simple-links
- Event-link

Internal links are excluded since they do not encounter this problem:
- Card-grids
- User-registration

#### Test 
https://varnish.pr-1474.dpl-cms.dplplat01.dpl.reload.dk/paragraph-side

##### Note
A ticket might need to be created for this.

If an external link is added in card-grids, it results in an error that isn't very descriptive.

image

It is also not intended for external links to be added here, but I would have expected an error message in the paragraph.

### DIFF
--- a/web/themes/custom/novel/templates/components/hero.html.twig
+++ b/web/themes/custom/novel/templates/components/hero.html.twig
@@ -6,9 +6,9 @@
 %}
 
 <section {{ attributes.addClass(classes) }}>
-  {% if link %}
-    {% set url = link.0['#url'] %}
-    {% set target = url.options.target|default('_self') %}
+  {% if link.0 %}
+    {% set url = link.0['#url'].external ? link.0['#url'].uri : link.0['#url'].toString() %}
+    {% set target = link.0['#url'].options.target|default('_self') %}
 
     <a href="{{ url }}" target="{{ target }}" class="hero__content arrow arrow__hover--right-large">
       {% include "@novel/components/hero-inner.html.twig" %}
--- a/web/themes/custom/novel/templates/paragraphs/paragraph--simple-links.html.twig
+++ b/web/themes/custom/novel/templates/paragraphs/paragraph--simple-links.html.twig
@@ -1,7 +1,7 @@
 {% for item in content.field_link %}
   {% if item['#type'] == 'link' %}
-    <a href="{{ item['#url'].toString() }}" target="{{ item['#options'].attributes.target|default('_self') }}"
-       class="simple-link arrow__hover--right-small">
+    {% set url = item['#url'].external ? item['#url'].uri : item['#url'].toString() %}
+    <a href="{{ url }}" target="{{ item['#options'].attributes.target|default('_self') }}" class="simple-link arrow__hover--right-small">
       {{ item['#title'] }}
       {{ source(baseIconPath ~ '/arrow-ui/icon-arrow-ui-small-right.svg') }}    </a>
   {% endif %}
--- a/web/themes/custom/novel/templates/fields/field--field-event-link.html.twig
+++ b/web/themes/custom/novel/templates/fields/field--field-event-link.html.twig
@@ -1,6 +1,6 @@
 {% if has_event_link %}
   {% for item in items %}
-      {% set url = item.content['#url'].toString() %}
+      {% set url = item.content['#url'].external ? item.content['#url'].uri : item.content['#url'].toString() %}
       {% set title = item.content['#title'] %}
 
       {% set classes = [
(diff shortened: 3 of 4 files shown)


web/themes/custom/novel/templates/paragraphs/paragraph--simple-links.html.twig
```diff
@@ -1,7 +1,7 @@
 {% for item in content.field_link %}
   {% if item['#type'] == 'link' %}
-    <a href="{{ item['#url'].toString() }}" target="{{ item['#options'].attributes.target|default('_self') }}"
```
xendk, line 3: *sigh* How did we end up here?

Unless you're doing something explicitly funky, `item` is a render array, and simply doing `{{ item }}` renders the link, a tag and everything (the link formatter takes care of the target).

Of course, you don't get the classes on the link, but `simple-link` doesn't seem to be defined (so why do it exist in the HTML?), and `arrow__hover--right-small` is presentation that's more related to the *context* of the link than the link itself (it should be handled by a `.paragraph--simple-links a` instead, if you ask me).

web/themes/custom/novel/templates/components/hero.html.twig
```diff
@@ -6,9 +6,9 @@
 %}
 
 <section {{ attributes.addClass(classes) }}>
-  {% if link %}
-    {% set url = link.0['#url'] %}
-    {% set target = url.options.target|default('_self') %}
+  {% if link.0 %}
+    {% set url = link.0['#url'].external ? link.0['#url'].uri : link.0['#url'].toString() %}
+    {% set target = content.field_banner_link[0]['#url'].options.target|default('_self') %}
```
xendk, line 11: Why is target taken from `content.field_banner_link` when the URL is taken from `link`? How can we be sure that it's actually the same link?

It looks like `hero.html.twig` is included from `paragraphs/paragraph--hero.html.twig`, `components/full-event.html.twig` and `layout/node--branch--full.html.twig`, neither of which seems to pass either the `link` or `content` variables, and I can't find a preprocess function for it either. What's going on?

xendk (changes requested):

web/themes/custom/novel/templates/components/hero.html.twig
kasperbirch1: This is a copy-paste mistake, sorry 😅 Thanks for finding it.

web/themes/custom/novel/templates/paragraphs/paragraph--simple-links.html.twig
kasperbirch1: I'm not sure what to do here. I tried deleting the file to see if it will render a link with `:` correctly. It doesn't. So I think we have to do it that way.


Alternatively, we should reconsider making a widget for the link, but that will require more time on the task and we have paragraphs that have a link as a wrapper where it would not be able to be used?

image

xendk: Deleting which file? I'm not following...

I *ought* to be enough just to use `{{ item }}` in the template in order to get the standard code to render the link, but that means we'll lose the classes. And while nobody will miss `simple-link`, the styling wont be right if the other disappears. And fixing the link styling is a bit out of scope for this PR.

But I do want to point out that we're spending time fixing things that's already fixed because we're not doing things as they're done in Drupal.

xendk (approved):

web/themes/custom/novel/templates/paragraphs/paragraph--simple-links.html.twig
kasperbirch1: I tried removing `web/themes/custom/novel/templates/paragraphs/paragraph--simple-links.html.twig` to see Drupal's default rendering, but that does not work for links with : (proxy links). 

Have I misrepresented something? 😅

xendk: We should pair up on it in RL, so I can explain along the way rather than just issuing instructions.

kasperbirch1: That will be nice!

PR #1485, opened 2024-08-15
Insert missing blocked reason

spaceo:
The reason is used (in the moment of this writing) in the frontend and might be useful in the future in the backend as well.


Dresse (approved):
LGTM 👍

PR #1486, opened 2024-08-15
Allow setting any end time for events. DDFFORM-877 

rasben:
Updating the module, means that some of the upstream patches have been merged.
This commit also adds a new upstream patch, made by us. The patch fixes it, so the editor can create events that start at 00:00.

https://www.drupal.org/project/recurring_events/issues/3468300

Description of patch:

> Up until now, the min_time/max_time in the settings form, are just text fields, that take data in AM/PM.
> This is problematic, because the code that parses it, fails when an editor has inserted "00:00".
> This commit changes the form to instead use time inputs, and ports any existing data to the new format.
> This also makes the new format 24h instead of 12h.

#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-877


xendk (approved):
:+1:

PR #1488, opened 2024-08-16
DDFLSBP-785 - Add update hooks for fixing webform configuration translations

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-785

#### Description

We have experiences problems with faulty translated webform configurations, such as wrong formatted yaml files and duplicate keys. This has been fixed for most libraries using update hooks in earlier PR's:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1391
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1397

However it seems that the update hooks haven't been running successfully on all library sites. We have located 8 libraries where this is the case:
- Fredensborg
- Frederiksberg
- Halsnæs
- Horsens
- Næstved
- Norddjurs
- Tårnby
- Thisted

What they all have in common is that the language and default_language has been changed to danish instead of english. Furthermore webform.settings and webform.webform_options.* settings have been added to the ignored_config_entities list. 
The best guess for this is that the auto_ignore module has added them to the list, before we had the chance to restrict access to the specific configurations.

To solve the above issue, this PR adds 3 new webform update hooks:
1. Sets the langcode and default langcode back to english for websites who are currently set to danish. We do not want the langcode to be danish as it messes up the translation of configurations.

2. Removes unwanted webform configuration from ignored_config list. Specifically it removes webform.settings.yml and webform-webform_options.* from the ignore list. We don't want these config ignored, as it prevent us from fixing earlier configuration translation errors.

3. Reruns some of the earlier update hooks that fixes configuration translation errors. On some sites, they were not run successfully because of config ignore.

### DIFF
--- a/web/modules/custom/dpl_webform/dpl_webform.install
+++ b/web/modules/custom/dpl_webform/dpl_webform.install
@@ -7,6 +7,7 @@
 
 use Drupal\drupal_typed\DrupalTyped;
 use Drupal\locale\StringStorageInterface;
+use function Safe\preg_match;
 
 /**
  * Add all webforms to config ignore.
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_webform/dpl_webform.install
```diff
@@ -107,3 +108,89 @@ function dpl_webform_update_10004(): string {
 
   return implode("\n", $feedback);
 }
+
```
spaceo, line 111: You have a typo in the commit msh:
```

from the ignore list. We don't want these config ignored, as it prvent us

File: web/modules/custom/dpl_webform/dpl_webform.install [hunk 2]
```diff
@@ -107,3 +108,89 @@ function dpl_webform_update_10004(): string {
...
+ * danish, which causes issues when doing configuration translations. To
+ * prevent this in the future, we now set the default language to english.
+ */
+function dpl_webform_update_10005(): string {
+  $feedback = [];
+
+  \Drupal::configFactory()->getEditable('system.site')->set('langcode', 'en')->save();
+  \Drupal::configFactory()->getEditable('system.site')->set('default_langcode', 'en')->save();
+
+  return implode("\n", $feedback);
```
spaceo, line 125: Why the empty feedback here? Maybe just retun empty string if it is expected that the update hooks returns a string?

web/modules/custom/dpl_webform/dpl_webform.install
```diff
@@ -107,3 +108,89 @@ function dpl_webform_update_10004(): string {
...
+ * webform.webform_options.* and webform.settings from the ignore list.
+ */
+function dpl_webform_update_10006(): string {
+  $config_factory = \Drupal::configFactory();
+  $config_ignore_auto_settings = $config_factory->getEditable('config_ignore_auto.settings');
+  $ignored_configs = $config_ignore_auto_settings->get('ignored_config_entities');
+
+  // Remove webform.settings from config ignore list.
+  if (($key = array_search('webform.settings', $ignored_configs)) !== FALSE) {
+    unset($ignored_configs[$key]);
```
spaceo, line 143: if ($key = array_search('webform.settings', $ignored_configs)) {
    unset($ignored_configs[$key]);

web/modules/custom/dpl_webform/dpl_webform.install
```diff
@@ -107,3 +108,89 @@ function dpl_webform_update_10004(): string {
...
+
+  try {
+    $feedback[] = dpl_webform_update_10004();
+  }
+  catch (\Throwable $t) {
+    $feedback[] = "Unable to rerun dpl_webform_update_10004: {$t->getMessage()}.";
+  }
+
+  return implode("\n", $feedback);
+}
```
spaceo, line 196: I do not agree in this approach.
I think we should always move forward. If you need to redo previous actions simply make new update hooks repeating the old action. Update hooks are imperative by nature - I think.

spaceo (changes requested):
In my opinion I think we need to solve the re-running of the failed hooks in a different manner. Please look at my comment.
My other comments are minor.

web/modules/custom/dpl_webform/dpl_webform.install
Dresse: You are right. I have now added a string explaining what happened when running the hook.

web/modules/custom/dpl_webform/dpl_webform.install
Dresse: Fixed now. But I might update commit message after updating the PR.

web/modules/custom/dpl_webform/dpl_webform.install
Dresse: Removed redundant !== FALSE check

Dresse: I decided to re-add the !==FALSE statement after testing the code when removing it. It did not work as intended as array_search does not return TRUE if found, but instead either the string itself. Therefore I have to explicit check that the return is not FALSE.

web/modules/custom/dpl_webform/dpl_webform.install
Dresse: I guess that makes sense and my first approach was also to do that. But when I took a look at how we have done similar things earlier, I decided to copy the same approach.

However I have now updated the update hook, to include the previously run code from earlier update hooks instead of calling the update hook functions.

spaceo (approved):
LGTM 👍

PR #1500, opened 2024-08-21
Changing event recurrance, should not alter ID if possible. DDFFORM-967

rasben:
https://reload.atlassian.net/browse/DDFFORM-967

More info in the upstream issue/patch:

https://www.drupal.org/project/recurring_events/issues/3468521

----

This is a bit out of the ordinary, but as the reviewer, I'd like you to do two things:

### Review the code upstream

https://git.drupalcode.org/project/recurring_events/-/merge_requests/105

I managed to get the unit tests to work. There's warnings, but that's related to PHPStan/PHPCS, **that I'd like you to ignore.**
Please just focus on functionality, rather than 'code-beauty' :)

### Be the tester

This update is pretty technical, and is mostly to solve a problem that the ticket managers have reported.

That also makes it fairly hard to explain how to test for DDF, but, as we have the `skip_test` option, I'd like us to do the testing:

You can test it by altering the recurrance of an eventseries, and check that the IDs of the individual instances that have identical dateranges, have not changed

If in doubt, hit me up :)


composer.json
```diff
@@ -151,7 +151,7 @@
         "drupal/pathauto": "^1.12",
         "drupal/potion": "dev-2.x#e48d6ee336d960b9d2efb5d88e68402db6d229a6",
         "drupal/purge": "^3.6",
-        "drupal/recurring_events": "^2.0@RC",
+        "drupal/recurring_events": "^2",
```
rasben, line 154: The change that updates the module has been reviewed here:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1486

kasperg:
> Review the code upstream

I have reviewed the code upstream.

I agree with your point about coding standards. The current state of the module does not seem to align with the CI pipeline and thus I think the warnings there can be disregarded.

> Be the tester

I have tested and as far as I can tell it works. I have tried to change the start and end dates of an event series. This resulted in instances being created or deleted as expected. If I update the data of an single instance within the series this instance (and data) remains - as expected.

The feedback provided by the module when making a change also makes sense to me:

Monosnap Event series admin | DPL CMS | Logged in 2024-09-03 08-36-03

I would have appreciated a Cypress test which also proved that this worked as expected. Perhaps that is too late at this point.

kasperg (commented):
> Review the code upstream

I have reviewed the code upstream.

I agree with your point about coding standards. The current state of the module does not seem to align with the CI pipeline and thus I think the warnings there can be disregarded.

> Be the tester

I have tested and as far as I can tell it works. I have tried to change the start and end dates of an event series. This resulted in instances being created or deleted as expected. If I update the data of an single instance within the series this instance (and data) remains - as expected.

The feedback provided by the module when making a change also makes sense to me:

Monosnap Event series admin | DPL CMS | Logged in 2024-09-03 08-36-03

I would have appreciated a Cypress test which also proved that this worked as expected. Perhaps that is too late at this point.

rasben:
@kasperg I've updated the merge request in GitLab :)

kasperg:
:+1: I updated the merge request as well. [I have a single remaining comment](https://git.drupalcode.org/project/recurring_events/-/merge_requests/105#note_376859) for you to consider.

rasben:
>👍 I updated the merge request as well. [I have a single remaining comment](https://git.drupalcode.org/project/recurring_events/-/merge_requests/105#note_376859) for you to consider.

I've updated the merge request :)

kasperg:
:+1: This can be pushed along.

kasperg (approved):

PR #1516, opened 2024-08-27
Remove unused, related events view, and block.

rasben:
This block was not displayed on the pages, but
it was still loaded on all pages.
This is a problem, as the view, which also appears to be a remnant pre-dpl_related, loads *all* events.
This is an obvious performance problem on non-cached pages.


xendk (approved):

kasperg:
@dependabot rebase

PR #1522, opened 2024-08-29
Disable caching of static files in Varnish

kasperg:
#### Link

https://reload.atlassian.net/browse/DDFDRIFT-216

#### Description

Static files such as images may consume considerably more space in Varnish and there is no real need to store them there. NGINX will serve them just as quick with little extra effort.

To prevent the caching the NGINX configuration is updated to add the Surrogate-control "no-store" HTTP header to responses for known paths that should contain static files.

This header is supported by Varnish by default and also in the Lagoon Varnish for Drupal Docker image.

#### Screenshot of the result

`surrogate-control: no-store` is added to static file requests leading to `x-varnish-cache: MISS`.

Monosnap Frontpage | DPL CMS 2024-09-09 09-24-35

Monosnap Frontpage | DPL CMS 2024-09-09 09-25-46

Drupal pages are still cached:

Monosnap Frontpage | DPL CMS 2024-09-09 09-27-01


kasperg:
@dependabot rebase

rasben (approved):

PR #1524, opened 2024-08-30
Contextual websearch on tags and categories. DDFFORM-788

rasben:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-788

------------

You can test it by going to an article or event, and click on either a tag or category.
It should lead you to the term page, where related content is shown.

Example:

https://varnish.pr-1524.dpl-cms.dplplat01.dpl.reload.dk/hovedbiblioteket/arrangementer/vi-anbefaler/fars-legestue/2034-01-01
Screenshot 2024-09-12 at 10 34 53

Screenshot 2024-09-12 at 10 34 57

------------

- **Link tags and categories on full displays, to term pages. DDFFORM-788**

- **Aggregrate event+content tags+categories in search API. DDFFORM-788**

This allows us to create facets and filters, that look both in events and contents.

- **Remove core view, showed on term pages. DDFFORM-788**

This is an RSS feed that we do not want.

- **Contextual websearch on tags and categories. DDFFORM-788**

Embedding a contextualized version of the editorial search view directly on 'Categories' and 'Tags' individual term pages.
This allows the user to find content that is associated with this term.


rasben:
I know that the config test is failing, but please ignore that - I have no idea why it's happening, but I hope to find a solution for it eventually.

kasperbirch1 (approved):
I understand if this should be addressed in a separate ticket, but there shouldn't be an underline beneath the link in the tag. This should be corrected in the design system. 

Overall, great work! 👍

kasperg:
For the record: The Drupal Config check fails. This should not be merged until this is addressed. If we do not address it the problem will propagate to all other future work.

kasperg (changes requested):
For the record: The Drupal Config check fails. This should not be merged until this is addressed. If we do not address it the problem will propagate to all other future work.

kasperg (approved):

PR #1528, opened 2024-09-02
Ddfform 822 modal til abningstider

kasperbirch1:
#### Links
https://reload.atlassian.net/browse/DDFFORM-822

#### Depend on
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/713
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1399

#### Description
This pull request adds the opening hours modal. This is accomplished by adding two instances of the same React app, as the dependency on the size prop determines whether the markup should be displayed for the small icon used on the mobile display or the large clock displayed on the desktop.

#### Test
https://varnish.pr-1528.dpl-cms.dplplat01.dpl.reload.dk/

### DIFF
--- a/web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
+++ b/web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
@@ -48,7 +48,11 @@ void {
 }
 
 /**
- * Implements template_preprocess_page().
+ * Implements hook_preprocess_HOOK() for page templates.
+ *
+ * Adds two instances of the Opening Hours sidebar React app to the page.
+ * Since there are two triggers to open the sidebar, two instances of the app
+ * are required. The app will adjust its markup based on the `size` property.
  *
  * @param mixed[] $variables
  *   Theme variables.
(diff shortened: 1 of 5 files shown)


web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -45,3 +46,56 @@ void {
...
+ * @param \Drupal\node\NodeInterface[] $branches
+ *   Array of branch nodes.
+ *
+ * @return array<array{'branch_id': int|string|null, 'name': string|null, 'link': string, 'promoted': bool}>
+ *   Array of branch details.
+ */
+function dpl_opening_hours_build_branches_array(array $branches): array {
+  $branchesArray = [];
+  foreach ($branches as $branch) {
+    if ($branch instanceof NodeInterface) {
```
rasben, line 91: It would be more readable if you did

```suggestion
    if (!($branch instanceof NodeInterface)) {
      continue;
    }
```

rasben (changes requested):

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -77,25 +81,33 @@ function dpl_opening_hours_preprocess_page(array &$variables): void {
...
+      continue;
     }
+
+    $branchesArray[] = [
+      'branch_id' => $branch->id(),
+      'name' => $branch->getTitle(),
+      'link' => $branch->toUrl()->toString(),
+      'promoted' => $branch->hasField('field_promoted_on_lists')
+      && !$branch->get('field_promoted_on_lists')->isEmpty()
+      && (bool) $branch->get('field_promoted_on_lists')->value,
```
rasben, line 109: This looks wrong - is it possible you're trying to do a tenary?
In php, they look like this:

```
(some_condition) ? 'true' : 'false';
```

alternatively, you can be:

```
$something = $value_if_valid ?? $fallback_value
```

rasben (changes requested):

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
kasperbirch1: Okay, I go back to:
`'promoted' => $branch->hasField('field_promoted_on_lists') && !$branch->get('field_promoted_on_lists')->isEmpty() ? (bool) $branch->get('field_promoted_on_lists')->value : FALSE`

xendk: Rather than long boolean statements, use temporary variables to make it more readable:

```php
$hasPromoted = $branch->hasField('field_promoted_on_lists');
$promotedNotEmpty = !$branch->get('field_promoted_on_lists')->isEmpty();
$promoted = (bool) $branch->get('field_promoted_on_lists')->value;

$branchesArray[] = [
    'branch_id' => $branch->id(),
    'name' => $branch->getTitle(),
    'link' => $branch->toUrl()->toString(),
    'promoted' => $hasPromoted && $promotedNotEmpty ? $promoted : FALSE,
];
```

But we can simplify it. `$branch->get('field_promoted_on_lists')->value` returns null if it's empty, which is the same as false when casted to bool:

```php
$hasPromoted = $branch->hasField('field_promoted_on_lists');
$promoted = (bool) $branch->get('field_promoted_on_lists')->value;

$branchesArray[] = [
    'branch_id' => $branch->id(),
    'name' => $branch->getTitle(),
    'link' => $branch->toUrl()->toString(),
    'promoted' => $hasPromoted && $promoted,
];
```

But... We know that branches has the field, it's written to the configuration, and if someone deletes it, we'd rather have thing explode. So:

```php
$branchesArray[] = [
    'branch_id' => $branch->id(),
    'name' => $branch->getTitle(),
    'link' => $branch->toUrl()->toString(),
    'promoted' => (bool) $branch->get('field_promoted_on_lists')->value,
];
```

kasperbirch1: Thanks for the good explanation. I have corrected it  🙏

rasben (approved):

PR #1542, opened 2024-09-04
Automatically pull content into the `hero` paragraph

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-505

#### Description
This pull request adds functionality to automatically pull content into the `hero` paragraph for `page` and `article` content types.

It retrieves the `title`, `teaser-text`, and `teaser-image` from the node. 

**Please note that the node needs to have saved data before the hero section is added.**

#### Test
http://varnish.pr-1542.dpl-cms.dplplat01.dpl.reload.dk/frontpage

### DIFF
--- a/web/modules/custom/dpl_paragraphs/dpl_paragraphs.module
+++ b/web/modules/custom/dpl_paragraphs/dpl_paragraphs.module
@@ -10,6 +10,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
 use Drupal\dpl_react_apps\Controller\DplReactAppsController;
 use Drupal\field\Entity\FieldConfig;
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_paragraphs/dpl_paragraphs.module
```diff
@@ -479,3 +480,60 @@ function dpl_paragraphs_preprocess_paragraph__manual_event_list(array &$variable
   }
   $variables['rendered_field_events'] = $renderedEvents;
 }
+
+/**
+ * Implements hook_form_alter().
```
rasben, line 485: Remember to add a more general comment here, so we know what the hook actually does

rasben (changes requested):

PR #1552, opened 2024-09-10
Keep PHPStan at 1.11

kasperg:
#### Description

Currently PHPStan is an indirect dependency of the project through `mglaman/phpstan-drupal` and others.

However the version of PHPStan used directly affects the results of our CI pipeline and thus unintentional updates of the package performed through the update of other packages can result in CI errors. This notably happens via Dependabot. See #1549 .

To avoid this we move PHPStan to a direct dependency so updates for the package will also be managed directly.

We currently use version 1.11.11 and updates have yielded analysis errors so keep it at 1.11 for now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -191,6 +191,7 @@
         "php-mock/php-mock": "^2.3",
         "phpspec/prophecy-phpunit": "^2",
         "phpstan/extension-installer": "^1.3",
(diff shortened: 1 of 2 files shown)


Dresse (approved):
LGTM 👍

PR #1560, opened 2024-09-12
Port all 1-commit GitLab patches to commit patches.

rasben:
This is to make sure that we only get the relevant code, and not any possible-breaking pushes that may happen to the merge request in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -314,7 +314,7 @@
                 "2708731: Allow redirect to edit form after cloning": "https://www.drupal.org/files/issues/2022-11-15/improve-workflow-of-cloned-entity-2708731-39.patch"
             },
             "drupal/entity_clone_template": {
-                "3424597: Fix error when creating entity of type without templates": "https://git.drupalcode.org/project/entity_clone_template/-/merge_requests/3.patch"
+                "3424597: Fix error when creating entity of type without templates": "https://git.drupalcode.org/project/entity_clone_template/-/commit/ad4964dd51f7f5e814111fcb73e6e472eefccf5c.patch"
             },
             "drupal/field_inheritance": {
                 "3454350: Create seperate permission for adding field inheritance": "https://www.drupal.org/files/issues/2024-06-13/field_inheritance_permission.patch"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -332,18 +332,18 @@
...
-                "3343816: Specification includes disabled resources": "https://git.drupalcode.org/project/openapi_rest/-/merge_requests/4.patch"
+                "3343816: Specification includes disabled resources": "https://git.drupalcode.org/project/openapi_rest/-/commit/6618157334feca93fad1041583375eed660b9085.patch"
             },
             "drupal/paragraphs": {
                 "3027525: Show the paragraph and field label of fields that fail validation": "https://www.drupal.org/files/issues/2023-10-03/3027525_22.patch"
             },
             "drupal/paragraphs_ee": {
                 "3106254: Sorting paragraph list alphabetically in modal": "https://www.drupal.org/files/issues/2023-03-30/paragraphs_ee_sort.patch",
-                "3246408: Arrow buttons instead of drag n drop": "https://git.drupalcode.org/project/paragraphs_ee/-/merge_requests/21.patch"
+                "3246408: Arrow buttons instead of drag n drop": "https://git.drupalcode.org/project/paragraphs_ee/-/commit/4e16e0d79386dcf13db901f5634f2e30f0814137.patch"
```
kasperg, line 346: This merge request has 7 commits on GitLab but there is only 1 patch here. I would have expected it to be 7 commit patches accordingly.

kasperg (changes requested):
:+1: I really appreciate the intention here.

There seems to be one merge request which does not seem to have been ported correctly.

composer.json
rasben: I get the confusion :)
The changes have been merged into the dev version of the module, but has not been released into a stable release yet.

This commit is the merge commit, from here:

https://www.drupal.org/project/paragraphs_ee/issues/3246408#comment-15470073

kasperg: :+1: Thanks for the info.

PR #1561, opened 2024-09-13
Less frequent translation importing on development sites.

rasben:
**Remove reference to inactive PR environment**

**Run translation import less frequently.**

In an attempt to make caching more stable, and to match the rate at
which translations actually happen.

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -139,16 +139,13 @@ environments:
         command: drush cron
         service: cli
       - name: import translations
-        # TODO Consider if translation checking frequency should be reduced.
-        # A high frequency is used to ensure quick turnaround time during
-        # development. Updating (or checking) translations could have side effects on
-        # caching etc. and once things have settled down a lower frequency may be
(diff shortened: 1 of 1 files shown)


spaceo (commented):
I am not against these changes, but there some things to consider:
* The DDF team need to be aware of the fact that they in worst case will have to wait 12 hours before they can will be able to see changes made in the PO editor
* We have locale-check/-update in the post-rollout hook which makes sure that new translations are being imported. Why don't we have config-translations imported as well? This is not connected to your changes, but I just wonder - and we might as well fix that bit

rasben:
About your considerations:

1) I think in reality, this is not a problem for our day-to-day work. I can only speak for myself, but, when it comes to translations, I don't really think about it when they're missing locally, as that is DDF's job to handle. I just need to make sure things are translatable

2) Good point, but, I think it's outside the scope of this PR. If someone actually needs it in the future, I'm sure they can introduce it.

PR #1563, opened 2024-09-13
Optimizing GH actions, to use less minutes. (WIP)

rasben:
We can use `.lagoon.yml` to set up a GH deployment, that we can use to listen for in GH actions.
This is an alternative to us spending a lot of GH minutes just waiting for the site to become available.

As the `@todo`'s say, this needs some work, as I'm unsure how to use secrets and get around `set -e` as part of the deployment.
**I'm looking for input for this :)**

Ontop of that, also setting up `concurrency` rules to `ci-tests`, so if a second push is made, we cancel the old and unrelated workflow.


.lagoon.yml
```diff
@@ -4,9 +4,21 @@ project: dpl-cms-core
...
 api: https://api.lagoon.dplplat01.dpl.reload.dk/graphql
+environment_variables:
+  git_sha: 'true'
 
 tasks:
   post-rollout:
+    - run:
+        name: Create new GH deployment
+        command: |
+          # @TODO Where can i place the token as a secret?
```
achton, line 15: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository
and maybe
https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

?

.lagoon.yml
```diff
@@ -92,7 +104,19 @@ tasks:
           set -e
           drush en -y dpl_example_content
         service: cli
+    # @TODO How do we run this, even if something has gone wrong?
```
achton, line 107: I think you can use a [status check function](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions), probably `if: ${{ !cancelled() }}`

.github/workflows/lagoon-tests.yml
```diff
@@ -0,0 +1,16 @@
+---
+name: "Tests of deployment"
+
+on:
+  deployment_status
+
+# @todo - this action does nothing useful right now, but it is to show that
+# the site URL is available as part of deployment_status.
```
kasperg, line 8: I appreciate the comment but I suggest removing this file entirely.

As I see it we should not run tests on the deployed environment as implied in the workflow name. We do this on the corresponding environment established in CI.

.github/workflows/lagoon.yml
```diff
@@ -1,130 +0,0 @@
...
-        id: environment
-        run: |
-          echo ::set-output name=id::pr-${{github.event.number}}
-      - name: Close environment
-        uses: bobheadxi/deployments@v1.5.0
-        with:
-          step: deactivate-env
-          token: ${{ secrets.GITHUB_TOKEN }}
-          env: ${{ steps.environment.outputs.id }}
-          debug: ${{ runner.debug && 'true' || 'false' }}
```
kasperg, line 108: As I see it the current approach does not handle deactivating environments when the corresponding pull request is closed. I think it is a good idea to do so to avoid cluttering the project with obsolete environments which still show up as active.

As I see it Lagoon does not have a way to run code when closing an environment so I think it makes sense to keep this job as an alternate approach based on the fact that in practice closing a pull request will close the corresponding environment on Lagoon.

.github/workflows/lagoon.yml
```diff
@@ -1,130 +0,0 @@
...
-    name: Send synthetic event to Lagoon
-    runs-on: ubuntu-latest
-    needs: [BranchNameLength]
-    steps:
-    - name: Send pull request event
-      uses: distributhor/workflow-webhook@v3
-      env:
-        webhook_url: ${{ secrets.LAGOON_WEBHOOK_URL }}
-        webhook_secret: ${{ secrets.LAGOON_WEBHOOK_SECRET }}
-        webhook_type: 'json-extended'
```
kasperg, line 130: Please reconsider deleting this job. 

This is the one that informs Lagoon about changes to pull requests so without this new pull request environments will not be created.

.lagoon.yml
kasperg, line 15: Consider mentioning that we do not use `set -e` as we do not want the build process to stop if deployment handling fails.

.github/workflows/ci-tests.yml
```diff
@@ -5,7 +5,27 @@ env:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Ensure branch name length
+        uses: lekterable/branchlint-action@2.1.0
+        if: github.ref_type == 'branch' || github.ref_type == 'pull_request'
+        with:
+          allowed: |
+            /^.{1,100}$/
+          errorMessage: 'Branch name too long. This cannot be deployed to Lagoon.'
```
kasperg, line 27: If we keep the Lagoon workflow around then I suggest moving this job back.

.lagoon.yml
```diff
@@ -92,7 +104,19 @@ tasks:
...
+    - run:
+        name: Setting Deployment status
+        command: |
+          # @TODO How do we set this to 'failure', if something has gone wrong?
+          DEPLOYMENT_STATUS="success";
```
kasperg, line 112: Good question. Perhaps this video helps? https://docs.lagoon.sh/resources/faq/#how-can-i-check-the-status-of-a-build

You could also [ask the Lagoon team on Discord](https://docs.lagoon.sh/community/discord/).

kasperg (commented):
I appreciate the effort and I have tried to help out (a little) and provide feedback where I can.

PR #1565, opened 2024-09-13
Merged logic from duplicate novel_theme_suggestions_form_element_alte…

Dresse:
#### Description

Merged logic from duplicate novel_theme_suggestions_form_element_alter hook.


kasperbirch1 (approved):
I don't know how that could happen. But good that it was discovered so quickly and fixed now 👍

PR #1569, opened 2024-09-16
DDFHER-39 - deletion of opening hours category

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-39

#### Description
This pull request enables the administrator, local administrator, and editor roles to delete opening hour categories. Upon deletion, the user will be notified of the number of associated opening hours. Additionally, a pre-delete hook has been implemented to automatically remove all related opening hours when a category term is deleted.


web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -17,8 +23,7 @@ use Drupal\paragraphs\ParagraphInterface;
  * @param array &$variables
  *   Variables for the paragraph template, including the 'paragraph' entity.
  */
-function dpl_opening_hours_preprocess_paragraph__opening_hours(&$variables):
-void {
+function dpl_opening_hours_preprocess_paragraph__opening_hours(&$variables): void {
```
kasperg, line 26: This change is not related to the change at hand. I think it is nice but next time please at least keep it as a sepatate commit.

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -47,6 +52,79 @@ void {
...
+ *   The ID of the form being altered.
+ */
+function dpl_opening_hours_form_taxonomy_term_opening_hours_categories_delete_form_alter(array &$form, FormStateInterface $form_state, string $form_id): void {
+  $form_object = $form_state->getFormObject();
+  if (!$form_object instanceof EntityFormInterface) {
+    return;
+  }
+
+  $term = $form_object->getEntity();
+  if (!$term instanceof Term) {
```
kasperg, line 75:
```suggestion
  if (!$term instanceof TermInterface) {
```

When we want to assert something from a value [we generally try to use the interfaces instead of the class](https://www.drupal.org/docs/develop/standards/php/object-oriented-code#interfaces) when an appropriate interface exists.

Remember to update your imports accordingly.

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -47,6 +52,79 @@ void {
...
+  ]);
+
+  $opening_hours_repository = DrupalTyped::service(OpeningHoursRepository::class, 'dpl_opening_hours.repository');
+  $opening_hours_in_category = $opening_hours_repository->loadMultiple(categoryId: (int) $term_id);
+  $opening_hours_count = count($opening_hours_in_category);
+
+  $form['description']['#markup'] = t('@term_label is associated with @count opening hours. Deleting @term_label will also delete these opening hours. Are you sure you want to continue?', [
+    '@term_label' => $term_label,
+    '@count' => $opening_hours_count,
+  ]);
```
kasperg, line 97: Please remember to add a context to translations.

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -47,6 +52,79 @@ void {
...
+    '@term_label' => $term_label,
+  ]);
+
+  $opening_hours_repository = DrupalTyped::service(OpeningHoursRepository::class, 'dpl_opening_hours.repository');
+  $opening_hours_in_category = $opening_hours_repository->loadMultiple(categoryId: (int) $term_id);
+  $opening_hours_count = count($opening_hours_in_category);
+
+  $form['description']['#markup'] = t('@term_label is associated with @count opening hours. Deleting @term_label will also delete these opening hours. Are you sure you want to continue?', [
+    '@term_label' => $term_label,
+    '@count' => $opening_hours_count,
```
kasperg, line 96:
```suggestion
  $form['description']['#markup'] = t('@term_label is associated with @count opening hours. Deleting @term_label will also delete these opening hours. Are you sure you want to continue?', [
    '@term_label' => $term_label,
    '@count' => count($opening_hours_in_category),
```

Consider not extracting everything to a separate variable.

Especially if extracting it clearly shows what the value contains. If a value is used multiple times (like `$term_label`) or your lines become very long then I understand it but that is not the case here.

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -47,6 +52,79 @@ void {
...
+
+/**
+ * Implements hook_taxonomy_term_predelete().
+ *
+ * Deletes associated opening hours when a category term is deleted.
+ *
+ * @param \Drupal\taxonomy\TermInterface $term
+ *   The term being deleted.
+ */
+function dpl_opening_hours_taxonomy_term_predelete(TermInterface $term): void {
```
kasperg, line 108: :+1: This is appropriate usage of interfaces instead of classes.

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -47,6 +52,79 @@ void {
...
+  $term_id = $term->id();
+  if (!is_int($term_id) && !is_string($term_id)) {
+    return;
+  }
+
+  $opening_hours_repository = DrupalTyped::service(OpeningHoursRepository::class, 'dpl_opening_hours.repository');
+  $opening_hours_in_category = $opening_hours_repository->loadMultiple(categoryId: (int) $term_id);
+
+  foreach ($opening_hours_in_category as $instance) {
+    if ($instance->id && $instance->repetition->id) {
```
kasperg, line 122:
```suggestion
    if ($instance->id) {
```

I do not think you need the check for the repetition id.

The `delete()` method for the repository allows the repetition id to be `null` so we should be able to pass this value as well if the repetition of the instance does not have an id.

web/modules/custom/dpl_opening_hours/src/Model/OpeningHoursRepository.php
```diff
@@ -78,6 +78,10 @@ public function loadMultiple(int $branchId = NULL, \DateTimeInterface $fromDate
       $query->condition('repetition_id', $repetitionId);
     }
 
+    if ($categoryId) {
+      $query->condition('category_tid', $categoryId);
+    }
+
```
kasperg, line 84:
```suggestion
    if ($categoryId) {
      $query->condition('category_tid', $categoryId);
    }

```

This is really just nitpicking but I do not think we need a space between this check and the others which perform a similar function/purpose.

config/sync/user.role.local_administrator.yml
```diff
@@ -152,6 +152,7 @@ permissions:
   - 'delete page revisions'
   - 'delete terms in breadcrumb_structure'
   - 'delete terms in categories'
+  - 'delete terms in opening_hours_categories'
```
kasperg, line 155: :+1: Great that you remembered to add this for both roles.

kasperg (changes requested):
:+1: Well done! I think this needs a few more tweaks before this can be merged.

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
```diff
@@ -47,6 +51,78 @@ void {
...
+  $term_id = $term->id();
+  if (!is_int($term_id) && !is_string($term_id)) {
+    return;
+  }
+
+  $term_label = $term->label();
+
+  $form['#title'] = t('Are you sure you want to delete all "@term_label" opening hours?', [
+    '@term_label' => $term_label,
+  ]);
```
kasperg, line 87: You forgot to add a context here 😅 .

kasperg (approved):
:+1: but please remember to add the context everywhere before you merge.

web/modules/custom/dpl_opening_hours/dpl_opening_hours.module
kasperbirch1: Sorry 😅

PR #1584, opened 2024-09-18
Creating a legacy Redia RSS feed for events. DDFHER-60

rasben:
This feed is needed for some legacy apps to continue working. It will not be updated in the future - rather, this is a 1-1 recreation of the old feed.

As I've made a custom redia legacy module, i've also moved the legacy opening hours REST api into this module.

#### Link to issue

https://reload.atlassian.net/browse/DDFHER-60

**Old feed:**

https://www.aakb.dk/ding-redia-rss/event

**New feed:**

https://varnish.pr-1584.dpl-cms.dplplat01.dpl.reload.dk/ding-redia-rss/event


web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,234 @@
...
+      $changed_date = DrupalDateTime::createFromFormat('U', strval($event->getChangedTime()));
+
+      $items[] = [
+        'title' => $event->label(),
+        'description' => $this->getEventDescription($event),
+        'author' => $event->getOwner()->get('field_author_name')->getString(),
+        'id' => $event->id(),
+        'date' => $changed_date->format('r'),
+        'promoted' => FALSE,
+        'subtitle' => $event->get('event_description')->getString(),
```
kasperg, line 92: This is a bit confusing. Our event instances have a `description` field. This controller also has an `getEventDescription()` function.

I think I understand why this is needed but the current state of the code does not help me much.

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,234 @@
...
+        'start_time' => $event_dates[0]['value'] ?? NULL,
+        'end_time' => $event_dates[0]['end_value'] ?? NULL,
+        'media' => $this->getEventImageFields($event, 'redia_feed_large'),
+        'media_thumbnail' => $this->getEventImageFields($event, 'redia_feed_small'),
+        'branch' => [
+          'label' => $branch ? $branch->label() : NULL,
+          'id' => $branch ? $branch->id() : NULL,
+        ],
+      ];
```
kasperg, line 101: Consider replacing these array structures with value objects.

The current approach makes it hard to ensure the validity of the datastructure.

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,234 @@
...
+    if (!($file instanceof FileInterface)) {
+      return NULL;
+    }
+
+    $file_uri = $file->getFileUri();
+    $style = $this->entityTypeManager()->getStorage('image_style')->load($image_style);
+
+    if (empty($file_uri) || !($style instanceof ImageStyleInterface)) {
+      return NULL;
+    }
```
kasperg, line 139: Consider moving (some of) this logic to the `EventWrapper`.

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,234 @@
...
+   */
+  public function getFeed(): Response {
+    $items = $this->getItems();
+
+    $rss_content = $this->buildRss($items);
+
+    $response = new Response();
+    $response->setContent($rss_content);
+    $response->headers->set('Content-Type', 'application/rss+xml');
+    return $response;
```
kasperg, line 55: Please update this to return a cacheable response.

I do not know how often the feed is requested but I do not see a reason why we should not be able to cache it in Varnish.

config/sync/image.style.redia_feed_large.yml
```diff
@@ -0,0 +1,15 @@
...
+dependencies: {  }
+name: redia_feed_large
+label: 'Redia feed large'
+effects:
+  71f3eb2a-c6f8-4a85-a16f-651a6e43239e:
+    uuid: 71f3eb2a-c6f8-4a85-a16f-651a6e43239e
+    id: image_scale
+    weight: 1
+    data:
+      width: 2000
```
kasperg, line 13: :+1: for adding a max width. It seems as if the current implementation does not really set one but I think it is a good thing that we do now.

kasperg (changes requested):
Lots of code, lots of comments here 😅 .

I think it is a good idea to create a separate module for the legacy integrations and as I see it you have managed to recreate the feed 1-1 compared to the original.

There are a couple of places where we disagree on the overall code structure:

- I would like to see the `EventWrapper` used more.
- I would prefer more proper objects and less arrays and strings.

web/modules/custom/dpl_event/src/EventWrapper.php
```diff
@@ -113,6 +124,22 @@ public function getAddressField(): ?FieldItemListInterface {
     return $branch->get($branch_address_field);
   }
 
+  /**
+   * Getting the description, from the first available text paragraph.
+   */
+  public function getDescription(): ?string {
```
rasben, line 130: I've moved this, from EventRestMapper, as now we use it in two places.

web/modules/custom/dpl_event/src/EventWrapper.php
```diff
@@ -86,6 +86,22 @@ private function getDate(string $value): \DateTimeInterface {
     return new DateTimeImmutable($event_date_values[$value], new \DateTimeZone('UTC'));
   }
 
+  /**
+   * Getting an events branches.
+   *
+   * @return array<NodeInterface>|null
+   *   The matching branches.
+   */
+  public function getBranches(): ?array {
```
rasben, line 95: I've created a single method, that can be used both by Redia and the RestMapper.

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,234 @@
...
+  private function buildRss(array $items): string {
+    $config = $this->config('system.site');
+    $site_title = $config->get('name');
+    $site_url = $this->requestStack->getCurrentRequest()?->getSchemeAndHttpHost();
+    $feed_url = Url::fromRoute('dpl_redia_legacy.events');
+    $feed_url->setAbsolute();
+    $feed_url = $feed_url->toString();
+
+    $current_date = new DrupalDateTime();
+    $date = $current_date->format('r');
```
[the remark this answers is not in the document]
rasben: Uh, why? The time service only gives me unix timestamp, that I'll need to parse as a date anyway?
Isnt it better just to use DrupalDateTime directly?

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,234 @@
...
+   * @param array<mixed> $items
+   *   See $this->getItems();.
+   */
+  private function buildRss(array $items): string {
+    $config = $this->config('system.site');
+    $site_title = $config->get('name');
+    $site_url = $this->requestStack->getCurrentRequest()?->getSchemeAndHttpHost();
+    $feed_url = Url::fromRoute('dpl_redia_legacy.events');
+    $feed_url->setAbsolute();
+    $feed_url = $feed_url->toString();
```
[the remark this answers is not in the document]
rasben: I'd prefer not to get it from the request, as then the feed would include any possible parameters set, which is a bit.. dirty.
By using the route, I can be sure that the URL is correct.

rasben:
@kasperg This is ready for you again. All the relevant changes have been added in the latest commit :) ("Refactor...")

web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -208,3 +209,10 @@ function dpl_update_update_10016(): string {
 function dpl_update_update_10017(): string {
   return _dpl_update_install_modules(['dpl_search']);
 }
+
+/**
+ * Install dpl_search module.
```
kasperg, line 214:
```suggestion
 * Install dpl_redia_legacy module.
```

Wee bit of copy/paste here? 😄

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,142 @@
...
+    protected FileUrlGeneratorInterface $fileUrlGenerator,
+  ) {}
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container): static {
+    return new static(
+      $container->get('file_url_generator')
+
```
kasperg, line 33:
```suggestion
```

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,234 @@
...
+    $events = EventInstance::loadMultiple($ids);
+
+    $items = [];
+
+    foreach ($events as $event) {
+      /** @var \Drupal\node\NodeInterface[] $branches */
+      $branches = $event->get('branch')->referencedEntities();
+      $branch = reset($branches);
+      $event_dates = $event->get('date')->getValue();
+      $changed_date = DrupalDateTime::createFromFormat('U', strval($event->getChangedTime()));
```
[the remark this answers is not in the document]
kasperg: As I see it the code remains. Is that intentional?

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -0,0 +1,142 @@
...
+    $storage = $this->entityTypeManager()->getStorage('eventinstance');
+    $query = $storage->getQuery()
+      ->condition('status', TRUE)
+      ->condition('date.value', strtotime('today'), '>=')
+      ->accessCheck(TRUE)
+      ->sort('date.value');
+    $ids = $query->execute();
+
+    /** @var \Drupal\recurring_events\Entity\EventInstance[] $events */
+    $events = $this->entityTypeManager()->getStorage('eventinstance')->loadMultiple($ids);
```
kasperg, line 68:
```suggestion
    $events = $storage->loadMultiple($ids);
```

As I see it we already have the storage available. Is there a reason to retrieve it again?

web/modules/custom/dpl_redia_legacy/src/RediaEventMedia.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    $image_url = $style->buildUrl($file_uri);
+    $image_sizes = getimagesize($file_uri);
+    $file_size = filesize($file_uri);
+
+    $image_type = exif_imagetype($file_uri);
+
+    if ($image_type) {
+      $this->type = image_type_to_mime_type($image_type);
```
kasperg, line 56: Consider if calling `exif_imagetype()` is necessary.

As I read the documentation for `image_type_to_mime_type()` index 2 of the return value for `getimagesize()` already contains the image type.

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
kasperg: As I see it you have changes the implementation to return a cacheable response but please also add some cacheable dependencies to make it behave right :smile:

web/modules/custom/dpl_event/src/EventWrapper.php
kasperg: :+1: Excellent! I appreciate you updating it both places.

web/modules/custom/dpl_redia_legacy/src/RediaEvent.php
```diff
@@ -0,0 +1,78 @@
...
+use Drupal\Core\Datetime\DrupalDateTime;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\dpl_event\EventWrapper;
+use Drupal\node\NodeInterface;
+use Drupal\recurring_events\Entity\EventInstance;
+
+/**
+ * An event object, containing the properties the RSS feed needs.
+ */
+class RediaEvent extends ControllerBase {
```
kasperg, line 15: Please reconsider if this class and `RediaEventMedia` needs to extend `ControllerBase`.

I do not understand why this is necessary. As I see it the class is not a controller and it does not utilize any methods provided by the parent class.

web/modules/custom/dpl_redia_legacy/src/RediaEvent.php
```diff
@@ -0,0 +1,78 @@
...
+      $this->media = new RediaEventMedia($media, 'redia_feed_large');
+      $this->mediaThumbnail = new RediaEventMedia($media, 'redia_feed_small');
+    }
+
+    $this->branch = $branch;
+
+    // In the old system, there was a way for editors to mark content a
+    // promoted. However, this does not exist in the new CMS, so we wil
+    // just hardcode it.
+    $this->promoted = 'Falsk';
```
kasperg, line 75: :+1: Nice level of commenting.

web/modules/custom/dpl_redia_legacy/src/RediaEvent.php
```diff
@@ -0,0 +1,78 @@
...
+use Drupal\node\NodeInterface;
+use Drupal\recurring_events\Entity\EventInstance;
+
+/**
+ * An event object, containing the properties the RSS feed needs.
+ */
+class RediaEvent extends ControllerBase {
+  // We'll disable the documentation rules for the member properties, as
+  // they are pretty self-explanatory.
+  // phpcs:disable
```
kasperg, line 18: Consider if is it possible to only [disable the related rules for these lines](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file) instead of PHPCS entirely.

web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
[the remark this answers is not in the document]
kasperg: Getting the current time is a dependency which should be injected instead of accessed directly.

web/modules/custom/dpl_redia_legacy/src/RediaEvent.php
rasben: They both utilize EntityTypeManager, that comes from that :)

web/modules/custom/dpl_redia_legacy/src/RediaEvent.php
rasben: I did try, and it just refused to accept it, so this is me going nuclear.

PR #1599, opened 2024-09-25
Fix url to React components release 2024.39.0

kasperg:
#### Description

The file currently mentioned does not exist. The appended SHA differs from what is attached to the release.

Why this error occurred is unknown.

The current changes are the result of running
RELEASE=2024.39.0 VERSION=2024.39.0 task dev:composer:update-react

Note that this also rolls back a lot of changes made to the lock file when design system and React components where added in 6f7a3f40c1f4f5f5951fb3c2dccdf5b07d4fbc45

Note that this only occurs in develop because the dependencies where added in different commits for main.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
                 "version": "2024.39.0",
                 "type": "drupal-library",
                 "dist": {
-                    "url": "https://github.com/danskernesdigitalebibliotek/dpl-react/releases/download/2024.39.0/dist-2024-39-0-ffbef7fb6bc88407e932f9fb981278fac9d5bf45.zip",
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍 
**Note:**
I do not understand this sentence:
_Note that this also rolls back a lot of changes made to the lock file when design system and React components where added in https://github.com/danskernesdigitalebibliotek/dpl-cms/commit/6f7a3f40c1f4f5f5951fb3c2dccdf5b07d4fbc45_
As far as I can see the changes in the lock file  is mainly change of how the lines are formatted - typically arrays

kasperg:
> As far as I can see the changes in the lock file is mainly change of how the lines are formatted - typically arrays

Agreed. It is just line formatting. It caught me somewhat off guard because it is not something I have seen when preparing other releases.

PR #1600, opened 2024-09-25
Revert changing event reoccurence

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-16

#### Description

This reverts  #1500.

When rolling out these changes to existing sites this causes problems
during the configuration import part when running drush deploy:

```
> Error: Call to a member function getLabel() on null in /app/web/modules/contrib/scheduler/src/SchedulerManager.php on line 1171 #0 /app/web/modules/contrib/scheduler/scheduler.module(1450): Drupal\scheduler\SchedulerManager->entityUpdate()
> https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1 /app/web/core/lib/Drupal/Core/Config/ConfigImporter.php(574): _scheduler_config_import_process_entity_definitions()
> https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2 /app/vendor/drush/drush/src/Commands/config/ConfigImportCommands.php(259): Drupal\Core\Config\ConfigImporter->doSyncStep()
```

git bisect has traced the error back to this change. Reverting it
removes the problem.

Based on this we cannot tell if it is the update of the module
which was introduced as a part of the updated event reoccurrence
handling which causes the problem or the actual event reoccurrence
handling.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -151,7 +151,7 @@
         "drupal/pathauto": "^1.12",
         "drupal/potion": "dev-2.x#e48d6ee336d960b9d2efb5d88e68402db6d229a6",
         "drupal/purge": "^3.6",
-        "drupal/recurring_events": "^2",
+        "drupal/recurring_events": "^2.0@RC",
         "drupal/redirect": "^1.9",
         "drupal/redis": "^1.7",
         "drupal/restui": "^1.21",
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #1618, opened 2024-10-02
Optimizing GH actions, to use less minutes.

rasben:
We can use `.lagoon.yml` to set up a GH deployment, that we can use to listen for in GH actions.
This is an alternative to us spending a lot of GH minutes just waiting for the site to become available.

By using bash traps, we can also send back error codes when `drush deploy` (or other stuff) goes wrong.
This means we have to remove the `set -e`, but we
get around that by calling exit in the catcher.

Ontop of that, also setting up `concurrency` rules to `ci-tests`, so if a second push is made, we cancel the old and unrelated workflow.

With this set up, a new GH Action can be created, if you need a link to the environment:

```yml
on:
  deployment_status

jobs:
  tests:
    if: github.event.deployment_status.state == 'success'
    runs-on: ubuntu-latest
    steps:
      - run: echo "$URL"
        env:
          URL: ${{ github.event.deployment_status.target_url }}
```

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -5,6 +5,12 @@ env:
   PHP_VERSION: 8.1
   COMPOSER_VERSION: v2
 
+# Detect if this action is already running, and cancel it.
+# This most likely happened because a second push has been made to a branch.
+concurrency:
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   ValidateComposer:
     name: Validate Composer
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -4,13 +4,31 @@ project: dpl-cms-core
 
 ssh: 20.238.147.183:22
 api: https://api.lagoon.dplplat01.dpl.reload.dk/graphql
+environment_variables:
+  git_sha: 'true'
 
 tasks:
   post-rollout:
     - run:
+        name: Create new GH deployment
+        command: |
+          GH_DEPLOYMENT=$(curl -L -X POST -H "Authorization: Bearer $GH_DEPLOYMENT_TOKEN" \
+          https://api.github.com/repos/danskernesdigitalebibliotek/dpl-cms/deployments \
+          -d "{\"ref\":\"$LAGOON_PR_HEAD_BRANCH\", \"environment\":\"$LAGOON_PR_HEAD_BRANCH\", \
+              \"description\":\"Triggered by Lagoon\", \"transient_environment\": true, \
+              \"auto_merge\": false, \"required_contexts\": []}")
+
+          echo "$GH_DEPLOYMENT"
+          GH_DEPLOYMENT_ID=$(echo "$GH_DEPLOYMENT" | grep -m 1 '"id":' | sed -E 's/.*"id": ([0-9]+),.*/\1/')
+
+          echo "Created GH deployment with ID '$GH_DEPLOYMENT_ID'"
+          echo "$GH_DEPLOYMENT_ID" > /tmp/gh_deployment_id
+        service: cli
+    - run:
         name: If drupal is not installed
         command: |
-          set -e
+          source dev-scripts/lagoon-error-handling.sh
+
           if tables=$(drush sqlq "show tables like 'node';") && [ -z "$tables" ]; then
             # Install and set the admin password to a Lagoon variable if it exists.
             if [[ -n $PR_DRUPAL_PWD ]]; then
@@ -29,7 +47,8 @@ tasks:
     - run:
         name: drush deploy
         command: |
-          set -e
+          source dev-scripts/lagoon-error-handling.sh
+
           if [[ -f config/sync/system.site.yml ]]; then
             echo "Config detected, doing a drush deploy"
             drush deploy
@@ -49,7 +68,8 @@ tasks:
         # it will be gone.
         name: Create module upload directory in public files
         command: |
-          set -e
+          source dev-scripts/lagoon-error-handling.sh
+
           if [[ ! -d "web/sites/default/files/modules_local" ]]; then
             echo "Creating directory for module uploads"
             mkdir web/sites/default/files/modules_local
@@ -58,14 +78,16 @@ tasks:
     - run:
         name: Import translations
         command: |
-          set -e;
+          source dev-scripts/lagoon-error-handling.sh
+
           drush locale-check
           drush locale-update
         service: cli
     - run:
         name: Create test users
         command: |
-          set -e
+          source dev-scripts/lagoon-error-handling.sh
+
           # Only create test users if they do not exist already.
           if editor_user=$(drush sqlq 'select * from users_field_data where name = "editor"') && [ -z "$editor_user" ]; then
             drush user:create editor --password="$PR_DRUPAL_PWD"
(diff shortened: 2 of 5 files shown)


.lagoon.yml
```diff
@@ -89,9 +111,27 @@ tasks:
...
+          # Read the deployment ID from the file
+          GH_DEPLOYMENT_ID=$(cat /tmp/gh_deployment_id)
+          echo "Setting GH deployment status '$GH_DEPLOYMENT_ID': '$DEPLOYMENT_STATUS'"
+
+          # Setting the deployment status, on the GH deployment created earlier.
+          curl -L \
+            -X POST \
+            -H "Authorization: Bearer $GH_DEPLOYMENT_TOKEN" \
+            https://api.github.com/repos/danskernesdigitalebibliotek/dpl-cms/deployments/$GH_DEPLOYMENT_ID/statuses \
+            -d "{\"environment\":\"$LAGOON_PR_HEAD_BRANCH\",\"state\":\"$DEPLOYMENT_STATUS\", \"target_url\":\"$DRUPAL_URL\"}"
```
kasperg, line 133: Monosnap Enable opening links in a new tab for `simple_links` paragraph + Update default content by kasperbirch1 · Pull Request #1617 · … 2024-10-03 09-53-56

You probably need to reconsider your use of [`target_url` (or `log_url`) and `environment_url`](https://docs.github.com/en/rest/deployments/statuses?apiVersion=2022-11-28#create-a-deployment-status) when creating an environment status.

.lagoon.yml
```diff
@@ -4,13 +4,31 @@ project: dpl-cms-core
...
 tasks:
   post-rollout:
+    - run:
+        name: Create new GH deployment
+        command: |
+          GH_DEPLOYMENT=$(curl -L -X POST -H "Authorization: Bearer $GH_DEPLOYMENT_TOKEN" \
+          https://api.github.com/repos/danskernesdigitalebibliotek/dpl-cms/deployments \
+          -d "{\"ref\":\"$LAGOON_PR_HEAD_BRANCH\", \"environment\":\"$LAGOON_PR_HEAD_BRANCH\", \
+              \"description\":\"Triggered by Lagoon\", \"transient_environment\": true, \
+              \"auto_merge\": false, \"required_contexts\": []}")
```
kasperg, line 19: It would be nice if we could set a deployment status with a link to the deployment log.

This is supported in the current setup. I think it would be even better to have it now that progressing from a pending to a successful (or failed) deployment status is 100% depenent of the Lagoon deployment.

kasperg (commented):
I think this is really nice! Moving from 15 minutes of wait time to a couple of seconds is good.

It would be nice to be able to work with the official GH CLI but without support for deployments we are left as is.

I have some questions and comments that I would like you to consider.

The bash changes are not my strong suit. Perhaps @xendk or @ITViking has something to say about this?

.lagoon.yml
```diff
@@ -4,13 +4,31 @@ project: dpl-cms-core
...
+  git_sha: 'true'
 
 tasks:
   post-rollout:
+    - run:
+        name: Create new GH deployment
+        command: |
+          GH_DEPLOYMENT=$(curl -L -X POST -H "Authorization: Bearer $GH_DEPLOYMENT_TOKEN" \
+          https://api.github.com/repos/danskernesdigitalebibliotek/dpl-cms/deployments \
+          -d "{\"ref\":\"$LAGOON_PR_HEAD_BRANCH\", \"environment\":\"$LAGOON_PR_HEAD_BRANCH\", \
```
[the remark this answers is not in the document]
rasben: Turns out I could get it from the lagoon name:

`$LAGOON_ENVIRONMENT`

.github/workflows/ci-tests.yml
```diff
@@ -5,6 +5,12 @@ env:
   PHP_VERSION: 8.1
   COMPOSER_VERSION: v2
 
+# Detect if this action is already running, and cancel it.
+# This most likely happened because a second push has been made to a branch.
+concurrency:
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
spaceo, line 12: I am not fond of canceling runs, because if a ci-test starts to fail we wont be able to trace exactly which commit hash that introduced the error.
It's not a biggie and I guess it is also about weighing cloud-resources/environmental awareness against the forensic possibility for pinpointing where an error occured.

.lagoon.yml
```diff
@@ -4,13 +4,36 @@ project: dpl-cms-core
...
+  git_sha: 'true'
 
 tasks:
   post-rollout:
+    - run:
+        name: Create new GH deployment
+        command: |
+          GH_DEPLOYMENTS=$(curl -H "Authorization: Bearer $GH_DEPLOYMENT_TOKEN" \
+          "https://api.github.com/repos/danskernesdigitalebibliotek/dpl-cms/deployments?ref=$LAGOON_PR_HEAD_BRANCH&environment=$LAGOON_ENVIRONMENT")
+          GH_DEPLOYMENT_ID=$(echo "$GH_DEPLOYMENTS" | grep '"id":' | head -n 1 | sed 's/[^0-9]*\([0-9]*\).*/\1/')
```
spaceo, line 17: Could the deployment request be done with `gh` instead of a curl request?
I think the "ttps://api.github.com/repos/danskernesdigitalebibliotek/dpl-cms/deployments, or at least the GH org/repo should be defined in a repository variable

.lagoon.yml
```diff
@@ -4,13 +4,36 @@ project: dpl-cms-core
...
+
+          echo "$GH_DEPLOYMENTS"
+          echo "Found GH deployment with ID '$GH_DEPLOYMENT_ID'"
+
+          # The only way to keep a value between tasks, is saving it in a
+          # file. Environment variables are killed between tasks, apart from
+          # the global ones.
+          echo "$GH_DEPLOYMENT_ID" > /tmp/GH_DEPLOYMENT_ID
+          LAGOON_DEPLOYS_LOG_URL="https://ui.lagoon.dplplat01.dpl.reload.dk/projects/$LAGOON_PROJECT/dpl-cms-$LAGOON_ENVIRONMENT/deployments"
```
spaceo, line 26: I think the start of the deployments url "https://ui.lagoon.dplplat01.dpl.reload.dk/projects" should be defined in a repository variable

dev-scripts/lagoon-set-gh-deploy-status.sh
```diff
@@ -0,0 +1,22 @@
...
+DRUPAL_URL=${2:-""}
+
+GH_DEPLOYMENT_ID=$(cat /tmp/GH_DEPLOYMENT_ID)
+LAGOON_DEPLOYS_LOG_URL=$(cat /tmp/LAGOON_DEPLOYS_LOG_URL)
+
+echo "Setting GH deployment status '$GH_DEPLOYMENT_ID': '$STATE'"
+
+# GitHub API request to update deployment status
+curl -L -X POST -H "Authorization: Bearer $GH_DEPLOYMENT_TOKEN" \
+  "https://api.github.com/repos/danskernesdigitalebibliotek/dpl-cms/deployments/$GH_DEPLOYMENT_ID/statuses" \
```
spaceo, line 20: I think the start of the deployments url "https://ui.lagoon.dplplat01.dpl.reload.dk/projects" should be an argument

spaceo (approved):
Approved with considerationz...

.lagoon.yml
rasben: Yes it can, but, then we need to have the `gh` command on lagoon, which we don't
I'd prefer not to introduce a new package :)

rasben: It would have been much easier with `gh` as I'd also be able to use jq, but oh well :)

PR #1619, opened 2024-10-03
Update reccurring_events to stable 2.0 + Changing event recurrance, should not alter ID if possible. DDFFORM-967 

rasben:
I found out what the issue was, that was reverted in https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1600

For some reason, I've accidentally removed the scheduler patch in the original commit.
In this version, I've made sure to include it.


kasperg (approved):
:+1: LGTM

PR #1629, opened 2024-10-03
Speed up ci:reset by passing along a pre-built DB.

rasben:
This speeds up the CI actions, spending ~2 minute instead of ~4 minutes on just setting up the basic site.

Ideally, we could speed it up even more by somehow caching the result of `docker compose {{ .DOCKER_COMPOSE_FILES }} pull`, but I could not get it working.
If we managed to get that working, I think we could get it all the way down to 30 seconds.


Taskfile.yml
```diff
@@ -382,7 +386,9 @@ tasks:
   ci:reset:
     desc: Create CI setup in a clean state
     cmds:
-      - task dev:reset
+      - task dev:reset:common
+      - task dev:cli -- drush sqlq --file=../tools/site-reset.sql.gz
```
spaceo, line 390: What is contained in site-reset.sql.gz? 
It would be nice if it was documented somewhere what is happening here

Taskfile.yml
```diff
@@ -125,8 +125,8 @@ tasks:
     cmds:
       - docker compose {{ .DOCKER_COMPOSE_FILES }} pull
 
-  dev:reset:
-    desc: Create local development setup in a clean state
+  dev:reset:common:
```
spaceo, line 128: What is the purpose of dev:reset:common? Would be nice to have in the description what the intention is.
Also consider which of the newly introduced dev:rest* cmds private so it is clear what is meant to be run as seprate task cmds and what is used internally

spaceo (commented):
It is a bit difficult to wrap my head around what is happening here.
I get the concept of resusing af prebulit db, but dont undestand how it is effectuated here.
A have some more questions that it would be nice if you take a look at.

PR #1631, opened 2024-10-04
DDFBRA-76 - Opsætning af next.js og graphQl modules i dpl-cms

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-76

#### Description

This is meant to be a POC branch, and as such should not be merged into develop yet. I
Its supposed to be something we can checkout locally and try out and fiddle around with.


Taskfile.yml
```diff
@@ -224,6 +224,9 @@ tasks:
       - task dev:cli -- drush user:create patron --password="test"
       - task dev:cli -- drush user:role:add 'patron' patron
 
+      - task dev:cli -- drush user:create graphql_consumer --password="test"
+      - task dev:cli -- drush user:role:add 'graphql_consumer' graphql_consumer
```
spaceo, line 228: This only happens locally, right? :)

dpl-go.json
```diff
@@ -0,0 +1 @@
..."skip","description":"Directs the executor to skip this field or fragment when the `if` argument is true.","locations":["FIELD","FRAGMENT_SPREAD","INLINE_FRAGMENT"],"args":[{"name":"if","description":"Skipped when true.","type":{"kind":"NON_NULL","name":null,"ofType":{"kind":"SCALAR","name":"Boolean","ofType":null}},"defaultValue":null}]},{"name":"include","description":"Directs the executor to include this field or fragment only when the `if` argument is true.","locations":["FIELD","FRAGMENT_SPREAD","INLINE_FRAGMENT"],"args":[{"name":"if","description":"Included when true.","type":{"kind":"NON_NULL","name":null,"ofType":{"kind":"SCALAR","name":"Boolean","ofType":null}},"defaultValue":null}]},{"name":"deprecated","description":"Marks an element of a GraphQL schema as no longer supported.","locations":["FIELD_DEFINITION","ENUM_VALUE"],"args":[{"name":"reason","description":"Explains why this element was deprecated, usually also including a suggestion for how to access supported similar data. Formatted using the Markdown syntax (as specified by [CommonMark](https://commonmark.org/).","type":{"kind":"SCALAR","name":"String","ofType":null},"defaultValue":"\"No longer supported\""}]}]}}
```
spaceo, line 1: This file is rather big. Is it recommended to commit it? Would be nice to not having it lying in the repository. Is it possible to generate it when we build the application?
I guess it is meant for being able to serve the schema quickly, without hitting code?

dpl-go.sdl
```diff
@@ -0,0 +1,978 @@
+"""The schema's entry-point for queries."""
```
spaceo, line 1: Same comment as the previous one, here...

spaceo (commented):
I have some questions that would be nice to clarify.
Also: Is the graphql consumer user/role still needed after nextjs has been removed?

Taskfile.yml
Dresse: Yes :)

dpl-go.json
Dresse: Right now I don't think we are using the file at all actually. But I think you are right about it being used for serving the schema quickly. Maybe we should just omit the files for now and generate them if we need them in the future. We would also have to update the files everytime we make changes to the schema.

Dresse:
> I have some questions that would be nice to clarify. Also: Is the graphql consumer user/role still needed after nextjs has been removed?

I still think it makes sense to keep the consumer user/role, as we still want to somehow restrict what can be done when calling the graphql endpoint. 

I have however removed the openapi patch and the dpl-go schema files for now. We can re-introduce them at a later point if needed.

spaceo (approved):
LGTM :+1
We agreed that the graphql consumer role should have the execute arbitrary graphql queries as wel...

Dresse:
> LGTM :+1 We agreed that the graphql consumer role should have the execute arbitrary graphql queries as wel...

I added the permission to the graphql_consumer role. But I also removed it for the anonymous role. Even though we do not have oauth authentication setup right now, we can still use the graphql_consumer by adding the UUID as part of the headers as "X-Consumer-ID". This will make sure that only people who have access to the created consumer will be able to make requests.

PR #1632, opened 2024-10-04
Add initial translation for paragraph error messages. DDFHER-21

rasben:
Usually, this would be done by DDF through the POEditor, but apparantly it wont show up on it's own.
We'll add it ourselves, and then leave DDF to update it if relevant.

https://reload.atlassian.net/browse/DDFHER-21


Dresse (approved):
LGTM 👍

PR #1635, opened 2024-10-07
Re-factor how to treat and display expired events in frontend. DDFHER-88

rasben:
### How to test:

Here is an example of an event, that no longer shows the expired events on the list:

https://varnish.pr-1635.dpl-cms.dplplat01.dpl.reload.dk/arrangementer/en-serie-der-bade-har-events-i-fortiden-og-fremtiden

Ontop of that, if you manage to find an expired event, it actually says so on the instance level:

https://varnish.pr-1635.dpl-cms.dplplat01.dpl.reload.dk/arrangementer/en-serie-der-bade-har-events-i-fortiden-og-fremtiden/2024-08-05

Screenshot 2024-10-09 at 08 47 04

Finally, if a series only has one upcoming event (but 1+ expired events), the auto-redirect now redirects to the one active eventinstance.
In the past, you would stay on the series page.

https://varnish.pr-1635.dpl-cms.dplplat01.dpl.reload.dk/arrangementer/series-3-expired-and-1-active-instances

---------

### Suggestions for the reviewer:

1) Review this, commit by commit
2) Make sure to hide whitespace - especially the last commit has a lot of pointless whitespace updates, review-wise :)

Screenshot 2024-10-07 at 13 30 39

--------

### Code description:

**Display if expired date on EventInstance full view. DDFHER-88**

This also means that we add the date as a field in the data overview.
This has been cleared with DDF.

**Expose functionality that finds upcoming event IDs. DDFHER-88**

By exposing it as part of the service, we can look up more correctly in
different parts of the code, where we want to see if there are relevant
eventinstances.

**Update EventRedirect code, to only respect upcoming events. DDFHER-88**

We have some code that redirects to the instance, if you visit an
eventseries with only one eventinstance.
However, the existing code only looked at related instances, and not if
the instances are date-expired.
By using the ReoccuringDateFormat service, we can both simplify the
code, and also redirect if there is only one, non-expired instance
available.

**Simplify logic that lists upcoming events. DDFHER-88**

The functionality that filters out expired events from the eventseries
view, can now be removed and replaced by the ReoccuranceDateFormatter
service.
This makes everything much simpler, whilst keeping the code easier to
maintain.

-----

https://reload.atlassian.net/browse/DDFHER-88


web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
```diff
@@ -42,15 +50,23 @@ public function checkEventSeriesRedirect(RequestEvent $event): void {
...
-    if (count($eventInstances) === 1 && $eventInstances[0] instanceof EventInstance) {
-      $response = new RedirectResponse($eventInstances[0]->toUrl()->toString());
+    $upcoming_ids = $this->reoccurringDateFormatter->getUpcomingEventIds($event_series);
+
+    if (count($upcoming_ids) > 1) {
+      return;
+    }
+
+    $event_instance_id = reset($upcoming_ids);
+    $event_instance = $this->entityTypeManager->getStorage('eventinstance')->load($event_instance_id);
```
kasperg, line 66: I am unsure about this change.

1. Code-wise: What happens if there are no upcoming instances? `$upcoming_ids` will be empty, `reset()` on an empty array with return `false` but what happens when `false` is passed to `load()`?
2. Functionally I still think it is appropriate to redirect from series with a single instance to the actual instance even if it is in the past.

web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
```diff
@@ -16,6 +18,14 @@
  */
 class EventSeriesRedirect implements EventSubscriberInterface {
 
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(
+    private EntityTypeManagerInterface $entityTypeManager,
+    private ReoccurringDateFormatter $reoccurringDateFormatter,
```
kasperg, line 26: Consider making these properties camel case.

I know this can be a mixed bag but in general [the Drupal code guidelines](https://www.drupal.org/docs/develop/standards/php/php-coding-standards#s-functions-and-variables) says:

> Be consistent; do not mix camelCase and snake_case variable naming inside a file.

web/modules/custom/dpl_event/dpl_event.module
```diff
@@ -182,30 +182,24 @@ function dpl_event_preprocess_eventseries(array &$variables): void {
...
-  });
+  $variables['formatted_date'] = $service->getSeriesDateString($event_series);
 
-  $eventinstances_in_series_filtered = array_values($eventinstances_in_series_filtered);
+  if (count($upcoming_ids) < 2) {
+    return;
+  }
 
+  $instance_storage = \Drupal::entityTypeManager()->getStorage('eventinstance');
+  $event_instances = $instance_storage->loadMultiple($upcoming_ids);
```
kasperg: :+1: This is an excellent refactor!

web/themes/custom/novel/templates/layout/eventinstance--card.html.twig
```diff
@@ -1,12 +1,14 @@
...
-  media: content.event_teaser_image,
-  variant: 'x-large',
-  has_media: (content.event_teaser_image.0),
-  teaser_text: content.event_teaser_text,
-  type_tag: 'Event'|trans,
-  date_tag: content.date,
-  title: content.title,
-  url: url('entity.eventinstance.canonical', {'eventinstance': eventinstance.id()})
-} only %}
+{% if not expired %}
```
kasperg, line 1: I understand your point but I do not think this is a good solution to the problem to place this logic in the template layer.

You suggest this solution to avoid having to implement the logic in every view but now we have to remember to put it in every template. I consider this is just as hard to maintain.

It may be easier to implement but on the other hand it messes with other parts of views dealing with the number of items in the list.

A view which is configured to show 5 events may now only show 4. A view with a pager with 10 events per page may now only show 8 per page even though it says that it is showing 10.

web/modules/custom/dpl_event/src/ReoccurringDateFormatter.php
```diff
@@ -90,19 +90,12 @@ public function getSeriesDateString(EventSeries $event_series): string|null {
...
-   *   An array containing the following keys:
-   *   - start: The start date of the upcoming event as a DrupalDateTime object.
-   *   - end: The end date of the upcoming event as a DrupalDateTime object.
-   *   - upcoming_ids: An array of upcoming event instance IDs.
-   *   Returns NULL if the start and end dates are not found.
+   * @return array<int, string>
+   *   An array of matching eventinstance IDs.
    */
-  public function getUpcomingEventDetails(EventSeries $event_series): array|null {
+  public function getUpcomingEventIds(EventSeries $event_series): array {
```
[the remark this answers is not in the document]
kasperg: In general the method would be easier to use for callers if it returned an array of `EventInstance` objects.

One way to achieve this would be to look at `$event_series->get('event_instances')->referencedEntities()` and filter away the ones with a date in the past.

Since this is a field value I would expect this to be at least as efficient.

web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
rasben: The reason I'm making them camelcase is because it's autowiring (is that the name?) - if i make it snake case, it'll end up being used as `$this->entity_type_manager->yada..` which seems inconsistent with how we normally do it

web/modules/custom/dpl_event/src/ReoccurringDateFormatter.php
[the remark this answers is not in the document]
rasben: I'd like to avoid doing that, because technically, it could end up loading a *lot* of entities, for no reason, that we're gonna filter out anyway.
referencedEntities loads the whole entity

web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
rasben: 1) Good point - I've updated line 62 to: `if (count($upcoming_ids) == 1) {`
2) I'm not sure I agree with you here - the whole point of the task is that expired events should be treated as second-class content - they should not show up in various lists, and only accessed if the user has a direct link to them. 
If an eventseries only has expired events, the instances wont show up in the list, and then i wouldnt expect a redirect to happen

web/themes/custom/novel/templates/layout/eventinstance--card.html.twig
rasben: Great point here, and I'm glad you called me out on it.
We've talked about this in person, and I've talked about it with Stefan.

This update (commit) has been reverted.

We came to the conclusion that we'll instead rely on automatic unpublication for events to not show up in various lists.

However, this will become a lot of changes, which I'll add in a seperate, new pull request.

web/themes/custom/novel/templates/layout/eventinstance--full.html.twig
```diff
@@ -1,5 +1,9 @@
 {% set date = eventinstance.get('date').get(0) %}
 {% set description_items = [
+  {
+    label: "Date"|trans,
+    value: content.date
+  },
```
kasperg, line 6: Please remember to add the translation context like you did for "Expired".

web/modules/custom/dpl_event/src/ReoccurringDateFormatter.php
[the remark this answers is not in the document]
kasperg: OK. Let's leave it as is for now.

web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
kasperg: I think you are referring to dependency injection. As far as I know this is not related to casing. The service does not use autowiring at the moment as you defined the arguments in `service.yml`.

kasperg (approved):
:+1: I have a comment for you to consider before you merge.

PR #1646, opened 2024-10-10
DDFHER-99 - Fix issue caused by preserved opening hours referencing deleted branches

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-99

#### Description
This fix addresses a problem where preserved opening hours with references to deleted branches caused `array_filter` to process invalid values.

The `loadMultiple` method has been updated to reindex array keys after filtering out `NULL` values. Without this update, `array_filter` retained the original keys, which resulted in the array being treated as an object.


kasperg (approved):
:+1: LGTM

PR #1649, opened 2024-10-13
Update and expand events Cypress tests. DDFHER-30

rasben:
The test works locally (when setting the test users to prefered language english..) but fails in Github. Looks like a timezone issue, but I dont have time to fix it.

### DIFF
--- a/cypress/e2e/events.cy.ts
+++ b/cypress/e2e/events.cy.ts
@@ -1,97 +1,306 @@
 import * as dayjs from "dayjs";
+import 'dayjs/locale/da'
 import "cypress-if";
 
-const events = {
-  singleEvent: {
-    title: "Single event",
-    subtitle: "A subtitle",
-    recurType: "Custom/Single Event",
-    start: dayjs("2030-01-01T10:00:00"),
-    end: dayjs("2030-01-01T16:00:00"),
+dayjs.locale('da')
+
+// The fields of the eventseries entity, that we use when creating and comparing.
+export interface EventType {
+  uuid?: string,
+  title?: string,
+  subtitle: string,
+  recurType: 'custom' | 'weekly_recurring_date',
+  ticketManagerRelevance: boolean,
+  state: string,
+  start: dayjs,
+  end: dayjs,
+  status: boolean,
+}
+
+// This interface only has what we currently use - ideally, it could be
+// expanded to follow the Swagger spec.
+interface EventApiType {
+  url: string,
+  uuid: string,
+  title: string,
+  description: string,
+  ticket_manager_relevance: boolean,
+  state: string,
+  date_time: {
+    start: string,
+    end: string,
   },
+  external_data: {
+    url: string,
+    admin_url: string,
+  }
+}
+
+export const eventBase: EventType = {
+  // Creating a random title, that we can use as a makeshift ID.
+  subtitle: Math.random().toString(36).slice(2, 7),
+  // Generating a random boolean.
+  ticketManagerRelevance: true,
+  recurType: "custom",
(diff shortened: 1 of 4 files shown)


cypress/e2e/events.cy.ts
```diff
@@ -1,97 +1,306 @@
...
+        url: "https://event.local",
+        admin_url: "https://admin.local",
+      }
+    }
+
+    findEventsInAPI(event).then((events) => {
+      const apiEvent = events[0];
+      // Resetting the logged-in session, for calling API.
+      cy.clearCookies();
+      cy.clearAllSessionStorage();
```
kasperg, line 278: You can set the state of Cypress to be an anonymous user using [`cy.anonymousUser()`](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/d2c5ea26f8db0b2fa1ab3469a30ef97c5046621e/cypress/support/commands.ts#L98-L98).

cypress/e2e/events.cy.ts
```diff
@@ -1,97 +1,306 @@
...
-  singleEvent: {
-    title: "Single event",
-    subtitle: "A subtitle",
-    recurType: "Custom/Single Event",
-    start: dayjs("2030-01-01T10:00:00"),
-    end: dayjs("2030-01-01T16:00:00"),
+dayjs.locale('da')
+
+// The fields of the eventseries entity, that we use when creating and comparing.
+export interface EventType {
```
kasperg, line 8: Please reconsider exporting this type.

It does not seem to be used elsewhere.

cypress/e2e/events.cy.ts
```diff
@@ -1,97 +1,306 @@
...
+  uuid?: string,
+  title?: string,
+  subtitle: string,
+  recurType: 'custom' | 'weekly_recurring_date',
+  ticketManagerRelevance: boolean,
+  state: string,
```
kasperg, line 14: Consider if this should be a set of allowed strings like you provided for `recurType`.

cypress/e2e/events.cy.ts
```diff
@@ -1,97 +1,306 @@
...
+  start: dayjs,
+  end: dayjs,
+  status: boolean,
+}
+
+// This interface only has what we currently use - ideally, it could be
+// expanded to follow the Swagger spec.
+interface EventApiType {
```
kasperg, line 22: :+1: Appreciate the comment.

I understand why you have opted not to include this.

cypress/e2e/events.cy.ts
```diff
@@ -1,97 +1,306 @@
...
-    title: "Single event",
-    subtitle: "A subtitle",
-    recurType: "Custom/Single Event",
-    start: dayjs("2030-01-01T10:00:00"),
-    end: dayjs("2030-01-01T16:00:00"),
+dayjs.locale('da')
+
+// The fields of the eventseries entity, that we use when creating and comparing.
+export interface EventType {
+  uuid?: string,
```
kasperg, line 9: I wonder why you have added this optional property.

It does not seem to be used anywhere.

kasperg (changes requested):
I appreciate the effort put into this but I think there are some significant issues that needs to be addressed before this is ready.

They primarily revolve around:
- Code structure
- Cypress practices
- Usage of existing structures

We previously talked about how the test of the Event API was related to the test of the Event UI and how code could be structured to be reused across both.

I have provided some notes. Let's talk it through in person.

As I see it this drops the existing test of the Event UI entirely to replace it with a test of the UI. I think that is a shame. I wonder how it turned out this way.

PR #1659, opened 2024-10-21
DDFHER-101- Allow retrieval of opening hours for multiple branches

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-101

#### Description

This pull request enhances the opening hours API (redia_legacy) by enabling clients to retrieve opening hours for multiple branches within a single request. Previously, the API allowed retrieval for only one branch at a time.

#### Test

**Redia** API (supports multiple branches):
- https://varnish.pr-1659.dpl-cms.dplplat01.dpl.reload.dk/opening_hours/instances?nid=10&from_date=2024-10-23&to_date=2024-10-23
- https://varnish.pr-1659.dpl-cms.dplplat01.dpl.reload.dk/opening_hours/instances?nid=18&from_date=2024-10-23&to_date=2024-10-23
- https://varnish.pr-1659.dpl-cms.dplplat01.dpl.reload.dk/opening_hours/instances?nid=10,18&from_date=2024-10-23&to_date=2024-10-23

**Cms API:** 
- https://varnish.pr-1659.dpl-cms.dplplat01.dpl.reload.dk/api/v1/opening_hours?branch_id=10&from_date=2024-10-23&to_date=2024-10-23
- https://varnish.pr-1659.dpl-cms.dplplat01.dpl.reload.dk/api/v1/opening_hours?branch_id=18&from_date=2024-10-23&to_date=2024-10-23

### DIFF
--- a/web/modules/custom/dpl_opening_hours/src/Model/OpeningHoursRepository.php
+++ b/web/modules/custom/dpl_opening_hours/src/Model/OpeningHoursRepository.php
@@ -59,14 +59,25 @@ class OpeningHoursRepository {
   /**
    * Load a collection of opening hours.
    *
+   * @param int[] $branchIds
+   *   The branch ID or an array of branch IDs.
+   * @param \DateTimeInterface|null $fromDate
+   *   The start date.
+   * @param \DateTimeInterface|null $toDate
+   *   The end date.
+   * @param int|null $repetitionId
+   *   The repetition ID.
+   * @param int|null $categoryId
+   *   The category ID.
+   *
    * @return OpeningHoursInstance[]
    *   Opening hours instances which match the provided criteria.
    */
-  public function loadMultiple(int $branchId = NULL, \DateTimeInterface $fromDate = NULL, \DateTimeInterface $toDate = NULL, int $repetitionId = NULL, int $categoryId = NULL): array {
+  public function loadMultiple(array $branchIds = [], \DateTimeInterface $fromDate = NULL, \DateTimeInterface $toDate = NULL, int $repetitionId = NULL, int $categoryId = NULL): array {
     $query = $this->connection->select(self::INSTANCE_TABLE, self::INSTANCE_TABLE)
       ->fields(self::INSTANCE_TABLE);
-    if ($branchId) {
-      $query->condition('branch_nid', $branchId);
+    if ($branchIds) {
+      $query->condition('branch_nid', $branchIds, 'IN');
     }
     if ($fromDate) {
       $query->condition('date', $fromDate->format('Y-m-d'), '>=');
--- a/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
+++ b/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
@@ -89,7 +89,7 @@ final class OpeningHoursResource extends OpeningHoursResourceBase {
 
     try {
       $openingHoursInstances = $this->repository->loadMultiple(
-        $typedRequest->getInt('branch_id'),
+        $typedRequest->getInts('branch_id'),
         $typedRequest->getDateTime('from_date'),
         $typedRequest->getDateTime('to_date')
       );
--- a/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
+++ b/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
@@ -89,7 +89,7 @@ final class OpeningHoursResource extends OpeningHoursResourceBase {
 
     try {
       $openingHoursInstances = $this->repository->loadMultiple(
-        [$typedRequest->getInt('branch_id')],
+        $typedRequest->getInt('branch_id') ? [$typedRequest->getInt('branch_id')] : [],
         $typedRequest->getDateTime('from_date'),
         $typedRequest->getDateTime('to_date')
       );
--- a/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
+++ b/web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
@@ -89,7 +89,7 @@ final class OpeningHoursResource extends OpeningHoursResourceBase {
 
     try {
       $openingHoursInstances = $this->repository->loadMultiple(
-        $typedRequest->getInts('branch_id'),
+        [$typedRequest->getInt('branch_id')],
         $typedRequest->getDateTime('from_date'),
         $typedRequest->getDateTime('to_date')
       );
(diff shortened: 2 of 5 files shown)


web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
```diff
@@ -37,8 +37,8 @@ public function getPluginDefinition(): array {
           Request::METHOD_GET => [
             'branch_id' => [
               'name' => 'branch_id',
-              'type' => 'integer',
-              'description' => 'The id of the branch for which to retrieve opening hours.',
+              'type' => 'string',
+              'description' => 'The id(s) of the branch(es) for which to retrieve opening hours. Can be a single id or a comma-separated list of ids.',
```
kasperg, line 41: :+1: for updating the description.

You have misunderstood which resource needs to be updated.

This updates the standard resource for the opening hours at `/api/v1/opening_hours` but the actual API that needs to be updated is the legacy API used by the Redia application. This resides in [`OpeningHoursResource.php`](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/54690dd61703eea71ebeb397afdf2f3573fc3d5d/web/modules/custom/dpl_redia_legacy/src/Plugin/rest/resource/OpeningHoursResource.php#L1-L1) at `/opening_hours/instances`.

web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
```diff
@@ -88,22 +88,59 @@ public function get(Request $request): Response {
...
+   * Parses the 'branch_id' parameter into an integer, array(integers), or NULL.
+   *
+   * @param string|null $branchIdParam
+   *   The 'branch_id' query parameter from the request.
+   *
+   * @return int|int[]|null
+   *   Returns an integer if a single valid ID is provided, an array of integers
+   *   if multiple valid IDs are provided, or NULL if no valid IDs are found.
+   */
+  private function parseBranchIds(?string $branchIdParam): int|array|null {
```
kasperg, line 121: Please reconsider the signature of this method.

As I see it, instead of returning three different types you might as well just return an array of ints.

This return types handles all the other options implicitly. An array can be empty or can contain one or more values.

web/modules/custom/dpl_opening_hours/src/Model/OpeningHoursRepository.php
```diff
@@ -59,14 +59,26 @@ public function load(int $id): ?OpeningHoursInstance {
...
    *   Opening hours instances which match the provided criteria.
    */
-  public function loadMultiple(int $branchId = NULL, \DateTimeInterface $fromDate = NULL, \DateTimeInterface $toDate = NULL, int $repetitionId = NULL, int $categoryId = NULL): array {
+  public function loadMultiple(int|array $branchId = NULL, \DateTimeInterface $fromDate = NULL, \DateTimeInterface $toDate = NULL, int $repetitionId = NULL, int $categoryId = NULL): array {
     $query = $this->connection->select(self::INSTANCE_TABLE, self::INSTANCE_TABLE)
       ->fields(self::INSTANCE_TABLE);
     if ($branchId) {
-      $query->condition('branch_nid', $branchId);
+      $operator = is_array($branchId) ? 'IN' : '=';
+      $query->condition('branch_nid', $branchId, $operator);
```
kasperg, line 81: Consider how the signature of this function needs to be updated.

One approach as you have implemented is to accept multiple types. If you follow this approach I think it would be preferable to convert the single value to an array and then always pass this with the the `IN` operator as the condition.

Another approach would be to change the signature to accept `array $branchIds = NULL` and then update callers to pass single value arrays.

web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
kasperg, line 41: Consider if the specification can be updated to properly specify the format.

The specification actually supports `array` types with a separate entry `items` for the value and `collectionFormat` for the delimiter.

web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
```diff
@@ -88,22 +88,59 @@ public function get(Request $request): Response {
...
+    if ($branchIdParam === NULL) {
+      return NULL;
     }
+
+    // Split the parameter into an array using commas as delimiters.
+    $branchIdsArray = explode(',', $branchIdParam);
+    // Trim whitespace and filter out empty values.
+    $branchIdsArray = array_filter(array_map('trim', $branchIdsArray), fn($value) => strlen($value) > 0);
```
kasperg, line 129: You could also achieve the same thing using [`preg_split()`](https://www.php.net/manual/en/function.preg-split.php).

kasperg, line 129: Consider moving the request parameter to array parsing to the `RequestTyped` class.

It seems quite generic/reusable to me if you leave out the non-positive integer part. I do not think it is that important.

kasperg (changes requested):
Appreciate the effort but I think this needs another round on multiple levels.

web/modules/custom/drupal_typed/src/RequestTyped.php
```diff
@@ -65,4 +65,30 @@ public function getDateTime(string $key, ?\DateTimeInterface $default = NULL) :
...
+   *
+   * @return int[]
+   *   An array of integers.
+   */
+  public function getInts(string $key, array $default = [], string $separator = ","): array {
+    $value = $this->request->get($key);
+    if ($value === NULL) {
+      return $default;
+    }
+    $strings = \Safe\preg_split("/\s*{$separator}\s*/", $value, PREG_NO_ERROR);
```
kasperg, line 90:
```suggestion
    $strings = \Safe\preg_split("/\s*{$separator}\s*/", $value, PREG_SPLIT_NO_EMPTY);
```

As far as I can tell [`PREG_NO_ERROR` is not a known flag for `preg_split()`.](https://www.php.net/manual/en/function.preg-split.php). Did you mean `PREG_SPLIT_NO_EMPTY`?

web/modules/custom/dpl_opening_hours/src/Plugin/rest/resource/v1/OpeningHoursResource.php
```diff
@@ -89,7 +89,7 @@ public function get(Request $request): Response {
 
     try {
       $openingHoursInstances = $this->repository->loadMultiple(
-        $typedRequest->getInt('branch_id'),
+        $typedRequest->getInts('branch_id'),
```
kasperg, line 92:
```suggestion
        [ $typedRequest->getInt('branch_id') ],
```

If you change it to `getInts()` the resource will accept multiple ints even though the specification now only accepts one.

I think it would be more correct to use `getInt()` here as this will fail if the parameter does not contain a single integer.

Then we can always wrap it in an array to comply with the updated `loadMultiple()` signature.

kasperg (approved):
:+1: LGTM

PR #1672, opened 2024-10-22
Also create deployments when PRs are synchronized

kasperg:
#### Description

Currently we only create deployments when pull requests are created or reopened. As a result the environment will be shown as outdated if there are additional commits pushed to the branch.

Update the workflow to also run on synchronisation to address this.

This was the same approach as used previously.

#### Screenshot of the result

Before:

dpl-cms 2024-10-22 22-34-56

After: See the deployments of this PR.


spaceo (approved):
LGTM 👍

PR #1676, opened 2024-10-23
DDFBRA-76 - setup GraphQL modules and configuration

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-76

#### Description

This is a re-commit of a previously reversed PR. It re-adds the graphql, graphql_compose and comsumers module and install them both in core.exstension but also install them in update hooks in dpl_update.install.

Lastly I have also cleaned up some unneeded dependency modules for the next module, as we no longer use it.


web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -239,3 +240,10 @@ function dpl_update_update_10019(): string {
 
   return $return;
 }
+
+/**
+ * Install typed_data, graphql, graphql_compose and consumers modules.
+ */
+function dpl_update_update_10020(): string {
+  return _dpl_update_install_modules(['typed_data', 'graphql', 'graphql_compose', 'consumers']);
```
spaceo, line 248: I have approved the PR. But is it necessary to list all those modules here?
Wouldn't it be enough to eg. include the `graphl_compose` module? - then the rest would follow (didn't double check my assumption).

spaceo (approved):
LGTM 👍

web/modules/custom/dpl_update/dpl_update.install
Dresse: I am actually not sure. I added them all to be on the safe side, as I had problems with this before.

PR #1541, opened 2024-09-04
DDFFORM-66 - Add fulltext search to `articles` and `events` views

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-66

#### Description
This pull request implements a full-text search filter for the `articles` and `events` views, allowing users to perform searches across the `title`, `description`, and `teaser` fields. The search API has been enhanced to index these fields, and case sensitivity has been disabled to make the search more user-friendly.

image
image

image
image

#### Test
https://varnish.pr-1541.dpl-cms.dplplat01.dpl.reload.dk/artikler
https://varnish.pr-1541.dpl-cms.dplplat01.dpl.reload.dk/arrangementer

PR #1678, opened 2024-10-24
DDFHER-98 - Event list jumps to top of page

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-98
<s>https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/750</s>

#### Description

This pull request fixes the problem in the Event view where it scrolls to the top when navigating to the next page.

Because we use the infinite scroll module in Drupal, AJAX is needed for the view. We turned that off in a previous pull request when we added full-text search in [this commit](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1541/files).

The solution is to re-enable AJAX, turn off auto-submit, and hide the “Apply” button with CSS 

#### Test
https://varnish.pr-1678.dpl-cms.dplplat01.dpl.reload.dk/arrangementer

### DIFF
--- a/config/sync/views.view.events.yml
+++ b/config/sync/views.view.events.yml
@@ -342,7 +342,7 @@ display:
           preserve_facet_query_args: false
           query_tags: {  }
       relationships: {  }
-      use_ajax: false
+      use_ajax: true
       header: {  }
       footer: {  }
       display_extenders: {  }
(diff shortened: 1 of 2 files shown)


kasperbirch1:
There are some problems with deployment for this PR 
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-1678/deployments

kasperg (approved):

PR #1682, opened 2024-10-25
DDFHER-107 - Upgrade `linkit` to version 6.1.6.

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-107

#### Description
Upgrade `linkit` to version 6.1.6.
Tested all paragraphs utilizing our custom `dpl_link_options` extension to ensure compatibility.


composer.json
```diff
@@ -136,7 +136,7 @@
         "drupal/jsonlog": "^4.0",
         "drupal/language_neutral_aliases": "^3.1",
         "drupal/libraries": "^4.0",
-        "drupal/linkit": "^6.1.2",
+        "drupal/linkit": "^6.1",
```
kasperg, line 139: Please reconsider this version constraint change.

This changes the specification from allowing any version from 6.1.2 and up until 7.0.0 to any version from 6.1.0 and up until 7.0.0.

I am not aware how you activated this change but I would suggest something like:

[`composer update drupal/linkit --with-all-dependencies`](https://www.drupal.org/docs/updating-drupal/updating-modules-and-themes-using-composer#s-install-updates)

kasperg (changes requested):
:+1: I appreciate your walkthrough of tested changes.

Please reconsider how the change has been carried out with Composer.

composer.json
kasperbirch1: My mistake, I executed the wrong command. Instead of running `dce composer update drupal/linkit --with-all-dependencies`, I accidentally used `dce composer require drupal/linkit --update-with-dependencies`

I have reset and force-pushed and tested it again.

kasperg (approved):

PR #1698, opened 2024-10-29
Update `drupal/openapi` to "2.2.0" and remove patch `3313521`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-113

#### Description
The patch [3313521](https://www.drupal.org/node/3313521), contributed by Kasper Garnes, is now included in `openapi 2.2.0`. After installation, the specification is validated and downloaded to `openapi.json`. The change to `x-tokenUrl` will not cause any issues as we do not use it. Refer to `web/modules/custom/dpl_opening_hours/src/EventSubscriber/ApiRouteSubscriber.php` for details.


kasperg (approved):
:+1: LGTM

PR #1710, opened 2024-11-01
Added dpl_consumers module that handles the creation of a new graphql…

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-133

#### Description

Added dpl_consumers module that handles the creation of a new graphqq_consumer user as well as a new consumer called GraphQL Consumer. 

The plan is that we are gonna use those for authentication when calling the GraphQL API.


spaceo (approved):
LGTM 👍

PR #1725, opened 2024-11-05
Redia XML fixes

kasperg:
#### Link to issue

#### Description

The current approach is somewhat frail in regards to ensuring
correct and consistent formatting.

In practice it also has challenges given the fact that it requires us
to ensure that our values are properly encoded. We have currently done
this for the description but in practice we also need to do this for
all other values.

To avoid this we update the implementation to use core PHP XML Writer
instead.

It has a number of advantages over alternate solutions:

- SimpleXML does not allow the same amount of control of the output.
Example; I could not get it to add an explicit encoding attribute on
the root XML element.
- Third party solutions add to our dependency tree. We do not want to
do this to maintain a legacy solution.

The new implementation has been checked by matching it with the
existing one using semantic diff'ing with difftastic.

It matches 1-to-1 except for the following exception: When data is
missing e.g. branch or media then the related elements are omitted
instead of returning empty elements.

By doing this we can drop the encoding of the description we already
added.

Additionally more fields from the original feed have been added.

I recommended reviewing this commit by commit.


rasben (approved):
I've looked through, commit by commit, and i dont have any comments in regards to the code.

I havent checked the actual feed, as I know it'll be tested by Redia

PR #1726, opened 2024-11-05
DDFBRA-138 - Create task for fetching UUID of the graphql_consumer

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-138

#### Description

Added a new task for fetching the UUID of the graphql_consumer. This UUID is used while calling the GraphQL API.

The task is:

`task dev:dpl-go:get-graphql-credentials` 

It fetches the UUID which should be set in the request headers as X-Consumer-ID

### DIFF
--- a/web/modules/custom/dpl_consumers/dpl_consumers.module
+++ b/web/modules/custom/dpl_consumers/dpl_consumers.module
@@ -1,0 +1,44 @@
+<?php
+
+/**
+ * @file
+ * DPL consumers module.
+ */
+
+use Drupal\dpl_consumers\DplGraphqlConsumersConstants;
+
+/**
+ * Function is used for printing out the consumer credentials to the console.
+ */
+function dpl_consumers_print_consumer_credentials(): void {
+  $graphql_consumer_client_id = DplGraphqlConsumersConstants::GRAPHQL_CONSUMER_CLIENT_ID;
+  $consumer_uuid = dpl_consumers_get_consumer_uuid($graphql_consumer_client_id);
+  print_r('Consumer UUID: ' . $consumer_uuid);
+}
+
+/**
+ * Get consumer by Client ID.
+ *
+ * @return string
+ *   The consumer UUID.
+ */
+function dpl_consumers_get_consumer_uuid(string $client_id): string {
+  try {
+    $consumer = \Drupal::entityTypeManager()
+      ->getStorage('consumer')
+      ->loadByProperties(['client_id' => $client_id]);
+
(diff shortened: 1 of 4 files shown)


web/modules/custom/dpl_consumers/dpl_consumers.module
```diff
@@ -0,0 +1,44 @@
+<?php
```
spaceo, line 1: I think you should turn this into a drush cmd.
It is actualley quite [easy](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/228c256beaa05cedb2eca60242c12ae9b5eaede0/web/modules/custom/collation_fixer/src/Commands/CollationFixerCommands.php)

Can we maybe make the cmd a bit smaller, that can be convenient if we are going to use it a lot.
Eg.: `dev:go:graphql-credentials`

spaceo (commented):
I think we should make the Drupal part as a drush cmd

web/modules/custom/dpl_consumers/dpl_consumers.module
Dresse: Ah smart! I'll convert it into a drush cmd. 👍 

Yes, I wasn't sure if it was more important for the command names to be descriptive or "convenient" to use. But I agree, I'll change the cmd name.

web/modules/custom/dpl_consumers/dpl_consumers.module
```diff
@@ -0,0 +1,33 @@
...
+      ->loadByProperties(['client_id' => $client_id]);
+
+    if (!empty($consumer)) {
+      $consumer = reset($consumer);
+      $uuid = $consumer->uuid();
+      return $uuid ?? 'Consumer UUID not found.';
+    }
+    else {
+      return 'Consumer not found.';
+    }
```
spaceo, line 27: i dont like the behaviour of here. So the contract of the function is that it will always return a string. The string can either be:
* A uuid string
* "Consumer UUID not found."
I think the return should only be a uuid string. If you cannot resolve the uuid you should throw an exception.

spaceo (approved):
Approved, but I think you should use 15 min in changing: `dpl_consumers_get_consumer_uuid ()` so it throws an error if you cannot resolve the uuid

web/modules/custom/dpl_consumers/dpl_consumers.module
Dresse: We now throw exceptions when we do not get the expected result.

PR #1706, opened 2024-10-30
DDFHER-115 - Cypress test for paragraphs module

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-115

#### Description

This pull request verifies the functionality of the `paragraph` module, along with key submodules like `Linkit`, allowing some paragraphs to open in a new tab. It also includes tests for different `CKEditor` editors and our media gallery.

Skærmbillede 2024-11-04 kl  10 51 18

PR #1727, opened 2024-11-06
DDFHER-123 - Update `dynamic_entity_reference` to "3.2.0"

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-123

#### Description
This update includes patch `3405253`, which renames the feild to `Dynamic entity reference`, making it more distinguishable from core entity reference field types that use the `Reference` label.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -311,7 +311,6 @@
                 "3200212: Overwrite files on import": "https://www.drupal.org/files/issues/2021-12-08/default_content-3200212.patch"
             },
             "drupal/dynamic_entity_reference": {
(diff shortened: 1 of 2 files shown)


kasperg:
@kasperbirch1 Does it make sense to add a Cypress test for this?

kasperbirch1:
> @kasperbirch1 What do you think about adding a Cypress test for this?

Yes, I can do that, but is it okay to add it in the [paragraphs test](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1706)? The paragraph card_grid_manual uses dynamic_entity_reference.

kasperg:
:+1: Let's do that.

kasperg (approved):

kasperg:
:+1: This is good to merge.

PR #1728, opened 2024-11-06
Temporarily disable the RSS feed for exposing events to Redia

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-127

#### Description

Problems with the current feed has led to events appearing in problematic ways e.g. by being listed as free to attend but actually having non-free ticket categories registered.

To avoid this we temporarily disable the feed to avoid exposing data.

Provide a cacheable response with a HTTP error code stating that the feed is not available currently.

404 seems like an appropriate error code given the situation even though this is strictly not a client issue. 503 indicates a problem which is more temporary than is the case there. 501 indicates problems with the HTTP method which is not the case.

You can test this locally by trying to access `/ding-redia-rss/event`. It should return a 404.

### DIFF
--- a/web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
+++ b/web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
@@ -2,7 +2,6 @@
 
 namespace Drupal\dpl_redia_legacy\Controller\RssFeeds;
 
-use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Cache\CacheableResponse;
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\Core\Datetime\DateFormatterInterface;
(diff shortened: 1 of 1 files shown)


web/modules/custom/dpl_redia_legacy/src/Controller/RssFeeds/EventsController.php
```diff
@@ -40,6 +40,9 @@ public static function create(ContainerInterface $container): static {
    * Getting the RSS/XML feed of the items.
    */
   public function getFeed(Request $request): CacheableResponse {
+    // Disable the feed while we wait for validation by Redia.
+    return new CacheableResponse("Feed disabled temporarily", Response::HTTP_NOT_FOUND);
```
kasperbirch1, line 44: Great with a description for users :)

kasperbirch1 (approved):
LGTM 👍

PR #1732, opened 2024-11-06
DDFHER-126 - Update `thecodingmachine/safe` to "^2.5.0"

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-126

#### Description

This pull request updates the `thecodingmachine/safe` package to version ^2.5.0.

The following functions, previously part of `thecodingmachine/safe`, are now included in PHP 8.0 and later versions:
	•	usort
	•	array_combine
	•	sprintf
	•	sort

Please review the changes in this commit: [7bb9890](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1732/commits/7bb9890ec7c311fde7779b19aea78aa9b5d2c29b). I ensured that if the value is not a string, it is converted to an empty string. However, I am unsure whether I should handle arrays in this context.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -184,7 +184,7 @@
         "spatie/color": "^1.5",
         "symfony/property-access": "^6.2",
         "symfony/property-info": "^6.2",
-        "thecodingmachine/safe": "^1.3",
+        "thecodingmachine/safe": "^2.5.0",
         "zaporylie/composer-drupal-optimizations": "1.2.0"
     },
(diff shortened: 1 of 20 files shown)


rasben:
I'm gonna remove myself, as I'm the one you talked with, when you made this.
I think @kasperg is better suited to review

kasperg (approved):

PR #1737, opened 2024-11-07
DDFBRA-134 - Setup simple_oauth module

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-134

#### Description

This PR adds the simple_oauth module, which we want to use for authenticating when requesting the GraphQL endpoints.

This PR also fixes 2 issue we found during testing of the module:
1. The openapi module was generating security definitions for oauth2 using v3 specification instead of v2. This resulted in errors when trying to generate API SDK files as v2 does not support multiple oauth2 flows. I created an issue on Drupal.org and created a patch to fix this. The patch is added in this PR.
2. The simple_oauth modules AuthenticationProvider had an issue where in its `applies` function was using a function (`isOauth2Request`) for checking wether Requests was OAuth2 requests. The problem however, was that it was just checking to see if a Bearer token was present. This causes problems with the current REST API for fetching opening hours, as those requests would sometime contain the Library token as a Bearer token, and thus the simple_oauth module would try to authenticate the request.
To solve this issue, this PR extends the simple_oauth modules functionality by decorating the service that provides the logic for determining if a request is a Oauth2 request. The added logic simple check to see if the request path stats with `/graphql`. If this is not the case, we return FALSE. 

The PR also updates the openapi.json file with the new specification.

Lastly this PR makes sure to generate the public and private keys that the `simple_oauth` modules needs to use when creating tokens. This is done in an update hook, but we also make sure to call the same update hook in the install hook function, as we want to make sure that the keys are generated on both old and new sites. I am unsure if this is the best approach, but we do something similar in `dpl_update` module, and I did not think it was worth using more time on figuring a better way of doing this. We also update the configuration for where the module should look for the generated keys. 

#### Other comments

We had a talk about where the public and private keys should be generated. We ended up deciding on placing them in folder called `simple_oauth_keys` placed in the` /web/sites/default/files` directory.


config/sync/simple_oauth.settings.yml
```diff
@@ -0,0 +1,11 @@
+_core:
+  default_config_hash: KsPFWSp6mgXIQgjBJEShfKUGn6VLRlbpIJ2EysXvXWM
+access_token_expiration: 300
+authorization_code_expiration: 300
+refresh_token_expiration: 1209600
+token_cron_batch_size: 0
+public_key: ../public.key
+private_key: ../private.key
```
spaceo, line 8: Are these keys that we need to ensure being present?

openapi.json
```diff
@@ -12,6 +12,12 @@
     "basic_auth": {
       "type": "basic"
     },
+    "oauth2": {
+      "type": "oauth2",
+      "flow": "accessCode",
+      "authorizationUrl": "http://varnish:8080/oauth/authorize",
+      "tokenUrl": "http://varnish:8080/oauth/token"
```
spaceo, line 19: Are these internal urls, and if yes, is it working as suspected?

spaceo (commented):
Nice addition!
I have a couple of answers I would like to have clarified

config/sync/simple_oauth.settings.yml
Dresse: Yes, we need those 2 keys to be present. 
However I am not sure what the best process for doing so is. Also I am a bit unsure on which directory they might best be stored.

openapi.json
Dresse: These are internal yes. So for instance on my local environment it would be:
http://dapple-cms.docker/oauth/token

And the PR environment it is:
https://varnish.pr-1737.dpl-cms.dplplat01.dpl.reload.dk/oauth/token

config/sync/simple_oauth.settings.yml
Dresse: I have now added functionality that generates the keys and updated the configuration to look for the keys in the right direcotry.

web/modules/custom/dpl_graphql/dpl_graphql.install
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Generates public and private keys used for the simple_oauth module.
+ */
+function dpl_graphql_update_10001(): string {
+  $file_system = DrupalTyped::service(FileSystemInterface::class, 'file_system');
+  $key_generator = DrupalTyped::service(KeyGeneratorService::class, 'simple_oauth.key.generator');
+
+  $path = '../web/sites/default/files/simple_oauth_keys';
+  $file_system->prepareDirectory($path, FileSystemInterface::CREATE_DIRECTORY);
```
spaceo, line 37: The files need to be written to a private directory so the are not accessible publicly.
The best think would be that we could write in a mount point that is not accesible at all by the web.
We could look into that later.
But for now we need to make sure that the directory that we create in `sites/default/files` is protected.
You need to investigate if the path to the private dir is set in settings, otherwise you shoul be able to set it via:
```
$settings['file_private_path'] = SOME_ABSOLUTE_PATH_TO_THE_DIRECTORY
````
When you need the path to the dir you should be able to do so by calling:
```
$private_files_path = \Drupal::service('file_system')->realpath('private://');

spaceo (approved):
.

spaceo (changes requested):
Vi need to have the keys properly secured in a private files dir

spaceo (approved):
LGTM 👍

PR #1741, opened 2024-11-08
Update core to 10.3.7. DDFNEXT-396

rasben:
We're pinning the version specifically to 10.3.7, rather than running `composer update`. We do this, both because we prefer it this way, but also because we have a problem when running update, that for some reason, the DPL setup wants to enable a dev version of Drupal core.
We're not sure why, but, for now, it does not matter, as we are pinning.

#### Link to issue

https://reload.atlassian.net/browse/DDFNEXT-396


kasperg (approved):
:+1: LGTM

PR #1743, opened 2024-11-11
Expand event API to include `partners` as array string. DDFHER-119

rasben:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-119

/api/v1/events

Screenshot 2024-11-11 at 15 20 39


web/modules/custom/dpl_event/src/Services/EventRestMapper.php
```diff
@@ -63,6 +63,7 @@ public function getResponse(EventInstance $event_instance): EventsGET200Response
       'branches' => $this->getBranches(),
       'address' => $this->getAddress(),
       'tags' => $this->getTags(),
+      'partners' => $this->getMultiValue('event_partners'),
```
kasperg, line 66: Consider what happens with this value if there are no partners.

The spec requires an optional array of strings.`getMultiValue()` returns `null` or an array of strings. Will the response leave out the `partners` property if the value is `null` or automatically convert it to an empty array?

Personally I would prefer `getMultiValue()` to always return an array of strings. The value might then be empty.

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
```diff
@@ -252,6 +253,22 @@ private function getDateField(string $field_name): ?DateTime {
...
+   *   An array of string values.
+   */
+  private function getMultiValue(string $field_name): ?array {
+    $strings = $this->getValue($field_name);
+
+    if (empty($strings)) {
+      return NULL;
+    }
+
+    return explode(', ', $strings);
```
kasperg, line 269: Please consider if this is an appropriate way to handle multivalue fields. 

As I read it, if one of the value contain `, ` then it will be treated as a multivalue. That does not seem right. Is there a better way to go about this?

If we go with this approach I would appreciate a reference to why `, ` is an appropriate separator.

`event_partners` is a multivalue field and to me it is not obvious that multiple values are separated by `, ` by Drupal Core/`$field->getString()`.

kasperg (changes requested):
:+1: but I think the handling of multivalue fields needs a some more consideration.

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
rasben: Fair point. I've updated it so it'll always return an array, even if it is empty.

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
rasben: It was an attempt to reuse the custom getValue() method, but, i didnt consider that strings might actually contain `, `. 
I've updated it, to use the Drupal `getValue` method instead, and using `array_column` to turn it into a simple array.

kasperg (approved):
:+1: LGTM

PR #1749, opened 2024-11-12
Uninstall database logging module on Lagoon sites

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-131

#### Description

We see considerable strain on the database in some situations in the production environment. Having the dblog module enabled will introduce many writes by default which causes strain. Using the JSNLog module only increases this.

Log data storage and access for sites running on Lagoon is currently duplicated between the dblog module and the JSON Log module, Loki and Grafana.

Thus to avoid both duplication and strain it should be reasonable to uninstall the dblog module on Lagoon sites.

It is nice to have it available on local development environment so we enable the module through the dev:enable-dev-tools task (this already happens) but exclude the configuration from configuration exports by adding the module to config_exclude_modules in settings.php.


web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -104,7 +104,8 @@ function dpl_update_install(): string {
   $messages[] = dpl_update_update_10022();
   $messages[] = dpl_update_update_10023();
   $messages[] = dpl_update_update_10024();
-
+  // Do not uninstall dblog module on install. It is not installed by default.
+  // $messages[] = dpl_update_update_10025()
```
Dresse, line 108: I'm not sure I understand this part. If we don't want to uninstall the dblog module on install, why are we adding the function for doing it in the first place?

Maybe I misunderstand this?

Dresse (approved):
I have a single question regarding the uninstall function in dpl_update. Other than that it looks good, so I'll approve 👍

web/modules/custom/dpl_update/dpl_update.install
kasperg: Fair question. It seems like a pattern has emerged where every update hook in `dpl_update` is also added to `dpl_update_install()`.

Going through the changes I do not understand why it is needed but everyone (including myself) has kept up this pattern for a while now.

I wanted to show that in this situation I had thought about the situation and it is intentional not to add the update hook here - even if `dpl_update_update_10026()` is added at a later point in time.

PR #1755, opened 2024-11-13
Add `retries` in `cypress.config`

kasperbirch1:
#### Description

<s>I have increase `defaultCommandTimeout` to the `cypress.config` because we observed failing tests due to the React app not loading quickly enough. I have tested this across six runs, and it seems to resolve the issue. </s>

`defaultCommandTimeout` alone wasn't enough. Cypress in GitHub Actions tends to be flaky, so we need to add `retries` to avoid spending too much time trying to fix sporadic issues.

I have kept the`scrollIntoView` in the opening hour test because if the test fails again, it will make debugging easier.
I also retained the increased `defaultCommandTimeout` because it does help, although it wasn’t sufficient on its own.

#### Screenshot of the result

image

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -4,6 +4,7 @@ module.exports = defineConfig({
   e2e: {
     // baseUrl is set using environment variables because it differs between
     // development and CI setups.
+    defaultCommandTimeout: 10000,
   },
   env: {
     // This is intentionally left empty.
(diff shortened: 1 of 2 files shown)


cypress.config.js
```diff
@@ -4,6 +4,11 @@ module.exports = defineConfig({
   e2e: {
     // baseUrl is set using environment variables because it differs between
     // development and CI setups.
+    defaultCommandTimeout: 10000,
```
kasperg, line 7: Consdier removing the `defaultCommandTimeout`.

If it turned outthe timeout does not make a difference then I suggest removing it again.

kasperg (approved):
:+1: LGTM

PR #1762, opened 2024-11-14
DDBRA-221  - Update consumer creation after adding simple oauth

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-221

#### Description

This PR seeks to update the way we create a GraphQL Consumer and GraphQL User on deploy. This consumer is used when making the first authentication request to the GraphQL endpoint. After an successful authentication, the request will return an `access_token` and an `refresh_token` which we will be using when making requests to our GraphQL API. 

When initially creating the consumer, we set a consumer secret, which locally can be set in `.env` file and on PR and production environments, it can be set as an env variable in Lagoon.

The PR adds an task and drush command for setting the consumer secret when working locally.

My guess is that some work are required in dpl-go to implement this way of calling the GraphQL endpoint. 

#### Screenshot of the result

Example of an authentication request done in insomnia and the response:
Screenshot 2024-11-14 at 14 49 40

Example of a request to the graphql endpoint using the `access_token` as authentication (Bearer token):
Screenshot 2024-11-14 at 14 51 02

This PR is dependent on the work done in another PR which ideally should be merged to develop first: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1737


web/modules/custom/dpl_consumers/src/Commands/DplConsumersCommands.php
```diff
@@ -0,0 +1,73 @@
...
+
+        return 'Consumer secret set successfully.';
+      }
+
+      return 'Consumer secret was not found.';
+    }
+    catch (\Exception $e) {
+      return 'Error fetching consumer.';
+    }
+  }
```
spaceo, line 60: We should not mix errors and result.
If errors occur we should throw an exception

web/modules/custom/dpl_consumers/dpl_consumers.module
```diff
@@ -0,0 +1,33 @@
...
+    }
+    else {
+      return 'Consumer not found.';
+    }
+  }
+  catch (\Exception $e) {
+    \Drupal::logger('dpl_consumers')->error($e->getMessage());
+    return 'Error fetching consumer.';
+  }
+}
```
spaceo, line 33: We should not mix errors and result.
If errors occur we should throw an exception

spaceo (changes requested):
Looks good but please look at comment about mixing results an errors...

web/modules/custom/dpl_consumers/dpl_consumers.module
Dresse: Please take another look, I have exchanged the "faulty" return strings for exceptions instead.

web/modules/custom/dpl_consumers/src/Commands/DplConsumersCommands.php
Dresse: Please take another look, I have exchanged the "faulty" return strings for exceptions instead.

spaceo (approved):
LGTM 👍

PR #1763, opened 2024-11-15
DDFBRA-215 - Updated 'drupal/honeypot' from version '2.1.4' to '2.2.0'.

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-215

#### Description

This PR updates `drupal/honeypot` module from `2.1.4` to `2.2.0`. 

I went through the [release notes](https://www.drupal.org/project/honeypot/releases/2.2.0) and [issue queue](https://www.drupal.org/project/issues/honeypot?text=&status=Open&priorities=All&categories=All&version=2.2.0&component=All).

The only issue I found was that upgrading to `2.2.0` gives problems, if you are not upgrading to `2.1.4` beforehand. We were already using `2.1.4` so this was no issue for us.

Tested the standard contact webform on `/kontakt` to see if the Honeypot was still working as intended.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -133,7 +133,7 @@
         "drupal/graphql_compose": "^2.2",
         "drupal/handy_cache_tags": "^1.4",
         "drupal/health_check": "^3.1",
-        "drupal/honeypot": "^2.1",
+        "drupal/honeypot": "^2.2",
         "drupal/job_scheduler": "^4.0",
         "drupal/jsnlog": "^1.3",
         "drupal/jsonlog": "^4.0",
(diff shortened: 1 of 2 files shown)


kasperg (approved):
:+1: Appreciate the thorough explation.

As I have mentioned to others: If you think it is interesting and feasible to implement a Cypress test to show that our usage of these contrib modules work I would appreciate this.

Dresse:
> 👍 Appreciate the thorough explanation of your considerations in this regard.
> 
> As I have mentioned to others: If you think it is interesting and feasible to implement a Cypress test to show that our usage of these contrib modules work I would appreciate this.

I have thought about making cypress test for testing the webform functionality (check that the default "kontaktformular" is is working). Here it would make sense to also test that things like Honeypot and Antibot modules are working as intended. 

I am looking into updating the webform module now.

PR #1766, opened 2024-11-18
Re-factor event scheduler setup. DDFHER-88

rasben:
Notice that this builds ontop of https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1635

Please review this, commit by commit. This is how you get the best context.

In short terms (summed up by ChatGPT :D ):

### Pull Request Description

**Redirect unpublished event instances**:
   - Implements a redirection from unpublished event instances (403) to their associated event series.
   - Resolves issue [DDFHER-88](#).

**Optimize unpublish logic for event series**:
   - Simplifies the logic by replacing a resource-intensive iteration through event instances with an efficient entity query for a count check.

**Enhance event scheduler settings**:
   - Introduces new configuration options for event instance and series unpublishing:
     - **Default**: Automatically unpublish event instances 1 hour after expiration.
     - **Optional**: Unpublish the event series when all associated event instances are unpublished (unrecommended).
   - Refactors the scheduler service to respect these new settings.
   - Migrates existing library settings to the updated configuration format.

[DDFHER-88]: https://reload.atlassian.net/browse/DDFHER-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
```diff
@@ -31,7 +36,14 @@ public function __construct(
...
     return [
-      KernelEvents::REQUEST => 'checkEventSeriesRedirect',
+      KernelEvents::REQUEST => [
+        ['checkEventSeriesRedirect'],
+      ],
+      // Only target users that end up on an unpublished eventinstance page,
+      // without being allowed to see it.
+      KernelEvents::EXCEPTION => [
+        ['checkEventInstanceRedirect'],
+      ],
```
kasperg, line 46: Please reconsider the structure of this addition.

As I see it you are adding new functionality to an existing class that related somewhat to the domain of the code - events and redirects. 

There is also some differences. The current class deals with event series as indicated by the name. The addition deals with event instances. There is no code reuse between the existing and new code.

This in mind I would suggest moving it to a separate class.

web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
```diff
@@ -31,7 +36,14 @@ public function __construct(
    */
   public static function getSubscribedEvents(): array {
     return [
-      KernelEvents::REQUEST => 'checkEventSeriesRedirect',
+      KernelEvents::REQUEST => [
+        ['checkEventSeriesRedirect'],
+      ],
+      // Only target users that end up on an unpublished eventinstance page,
+      // without being allowed to see it.
+      KernelEvents::EXCEPTION => [
```
kasperg, line 44: Consider whether this is is the proper event to react to.

As I read it our listener will react to any exception being thrown within the request. I would have excepted it to be able to react on requests instead. If that is not the case and access control leading to an exception runs first then perhaps we should look into ordering?

web/modules/custom/dpl_admin/assets/dpl_admin.css
```diff
@@ -184,3 +184,12 @@
...
 }
+
+/**
+  A very simple element, used in Drupal forms when displaying warnings.
+ */
+.dpl-form-warning {
+  color: #d80404;
+  font-weight: bold;
+  max-width: 500px;
+}
```
kasperg, line 195: Consider using [something that relates to the Gin admin theme](https://git.drupalcode.org/project/gin/-/blob/8.x-3.x/styles/base/_form.scss?ref_type=heads#L406).

This could make the addition appear more in tune with the remaining admin interface.

web/modules/custom/dpl_event/src/Form/SettingsForm.php
```diff
@@ -98,14 +100,66 @@ public function buildForm(array $form, FormStateInterface $form_state): array {
...
+      '#prefix' => '<div class="dpl-form-warning js-form-wrapper">',
+      '#markup' => $this->t('Notice - if series get unpublished, old instance links will no longer work. If you however keep the series published, expired instances will redirect to the associated series.', [], ['context' => 'DPL event']),
+      '#suffix' => '</div>',
+      '#states' => [
+        'visible' => [
+          ':input[name="unpublish_enable"]' => ['checked' => TRUE],
+          // PHPCS doesn't understand Drupal's weird way of doing states.
+          // phpcs:disable Squiz.Arrays.ArrayDeclaration.NoKeySpecified
+          'and',
+          // phpcs:enable Squiz.Arrays.ArrayDeclaration.NoKeySpecified
```
kasperg, line 159: :+1:

web/modules/custom/dpl_event/src/Form/SettingsForm.php
```diff
@@ -98,14 +100,66 @@ public function buildForm(array $form, FormStateInterface $form_state): array {
...
+          ':input[name="unpublish_enable"]' => ['checked' => TRUE],
+        ],
+      ],
+    ];
+
+    $form['unpublish']['unpublish_series_enable_warning'] = [
+      '#type' => 'container',
+      // js-form-wrapper is important - otherwise, Drupal states will not work.
+      '#prefix' => '<div class="dpl-form-warning js-form-wrapper">',
+      '#markup' => $this->t('Notice - if series get unpublished, old instance links will no longer work. If you however keep the series published, expired instances will redirect to the associated series.', [], ['context' => 'DPL event']),
```
kasperg, line 151: When I think about it I wonder why we ended up with not recommending unpublishing series when the last instance has been unpublished.

If the intention is that lists should not contain irrelevant content then keeping series which only contain expired instances seem to be counterproductive. Perhaps this is what you agreed on with the business?

kasperg (approved):
:+1: LGTM

I have some comments for you to consider before you merge.

You state that this resolves DDFHER-88 but the business also refers to testing DDFNEXT-512 as a part of this ticket. There are no changes in relation to search here. Is that intentional?

kasperg (commented):
As I see it the cache setttings for the view at `/events` does not use caching. Please consider whether it is appropriate to enable this.

web/modules/custom/dpl_event/src/Form/SettingsForm.php
rasben: Yes, this is what we agreed to with Stefan/the business:

For historical purposes, some libraries want to be able to display what has happened in the past.

web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
rasben: I'll leave it as it is, because it is about redirection *to* and *from* series - as i write in the comments :)

web/modules/custom/dpl_event/src/EventSubscriber/EventSeriesRedirect.php
rasben: >If that is not the case and access control leading to an exception runs first then perhaps we should look into ordering?

I did, and no matter what I did, I couldnt catch the request before the exception ran. However, I do actually prefer it this way, as that way, the editors can still preview unpublished instances.

rasben:
>As I see it the cache setttings for the view at /events does not use caching. Please consider whether it is appropriate to enable this.

Done - I've set it to a time of 1 hour.

web/modules/custom/dpl_admin/assets/dpl_admin.css
rasben: I did try that class, but, I prefer my own styling :)

PR #1773, opened 2024-11-19
Update event API, editing series should reflect in changed date. DDFHER-132

rasben:
https://reload.atlassian.net/browse/DDFHER-132

As we use inheritance, we want an updated series to also reflect in the updatedAt API property.
We could implement this, by programmatically saving all instances when the series is saved, but this may have unforseen consequences, as it is running against the Drupal system.
Instead, we'll look up the instance and series changed dates, and take which ever is newer.

### DIFF
--- a/web/modules/custom/dpl_event/src/Services/EventRestMapper.php
+++ b/web/modules/custom/dpl_event/src/Services/EventRestMapper.php
@@ -66,13 +66,44 @@ class EventRestMapper {
       'ticketCapacity' => $this->getValue('event_ticket_capacity'),
       'ticketCategories' => $this->getTicketCategories(),
       'createdAt' => $this->getDateField('created'),
-      'updatedAt' => $this->getDateField('changed'),
+      'updatedAt' => $this->getUpdatedDate(),
       'dateTime' => $this->getDate(),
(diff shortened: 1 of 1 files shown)


kasperbirch1 (approved):
This looks great to me! But since it’s not an area I’m deeply familiar with, I’ll leave it to you to decide if you’d like a second opinion. 

Based on your explanation, though, I’m happy to approve.

web/modules/custom/dpl_event/src/Services/EventRestMapper.php
```diff
@@ -75,6 +75,37 @@ public function getResponse(EventInstance $event_instance): EventsGET200Response
...
+   *
+   * As we use inheritance, we want an updated series to also reflect in the
+   * updatedAt API property.
+   * We could implement this, by programmatically saving all instances when
+   * the series is saved, but this may have unforseen consequences, as it is
+   * running against the Drupal system.
+   * Instead, we'll look up the instance and series changed dates, and take
+   * which ever is newer.
+   */
+  private function getUpdatedDate(): ?DateTime {
```
kasperg, line 89: Consider moving this to the event wrapper.

PR #1779, opened 2024-11-20
DDFBRA-183 - Anonymous and logged in user can try ebook

kasperbirch1:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-183

#### Description
This pull request adds translations for `Reader` as part of the changes in [dpl-react/pull/1555](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1555).

**It is also used to test the implementation of these changes.**


web/themes/custom/novel/templates/layout/page--reader.html.twig
```diff
@@ -0,0 +1,54 @@
...
+ * of the registration form, we have removed the header and footer regions for this
+ * page.
+ *
+ * Available variables:
+ *
+ * General utility variables:
+ * - base_path: The base URL path of the Drupal installation. Will usually be
+ *   "/" unless you have installed Drupal in a sub-directory.
+ * - is_front: A flag indicating if the current page is the front page.
+ * - logged_in: A flag indicating if the user is registered and signed in.
```
kasperg, line 17: Consider dropping all the following comments.

I do not think they are useful in the current context and the take of up a lot of visual space.

I appreciate the comment at the top.

web/themes/custom/novel/templates/layout/page--reader.html.twig
```diff
@@ -0,0 +1,54 @@
...
+ */
+#}
+
+<div>
+  {{ drupal_block('system_messages_block') }}
+
+  <main id="main-content" role="main">
+    {{ page.content }}
+  </main>
+</div>
```
kasperg, line 54: Consider if the `wrapper` div and the system messages block are helpful.

web/modules/custom/dpl_react_apps/dpl_react_apps.routing.yml
```diff
@@ -18,3 +18,12 @@ dpl_react_apps.work:
     _controller: '\Drupal\dpl_react_apps\Controller\DplReactAppsController::work'
   requirements:
     _permission: 'access content'
+
+dpl_react_apps.reader:
+  path: '/reader'
```
kasperg, line 23: Please consider making the identifier and/or order id arguments to the path.

As far as I can tell they are required for the reader to work. We do the same for e.g. works

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -500,4 +511,39 @@ public static function externalApiBaseUrls(): array {
...
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
+   */
+  public function reader(Request $request): array {
+
+    $identifier = $request->query->get('identifier');
+    $orderid = $request->query->get('orderid');
+
+    if (!$identifier && !$orderid) {
+      throw new \InvalidArgumentException('Either identifier or orderid must be provided.');
```
kasperg, line 531: Please consider changing this to a HTTP exception signaling there is something wrong with the request e.g. `BadRequestHttpException`

A normal exception will result in an error code 500 indicating an internal server error 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

web/modules/custom/dpl_react_apps/dpl_react_apps.routing.yml
```diff
@@ -18,3 +18,12 @@ dpl_react_apps.work:
...
+  defaults:
+    _controller: '\Drupal\dpl_react_apps\Controller\DplReactAppsController::reader'
+  requirements:
+    _permission: 'access content'
+  options:
+    no_cache: true
```
kasperg, line 29: Please reconsider if there is a reason to add `no_cache`.

We do not do so for other apps which I consider similar so I would not except it to be necessary here either.

kasperg (changes requested):
:+1: Nice start, but I think this needs some more consideration.

composer.json
```diff
@@ -92,6 +94,8 @@
         "danskernesdigitalebibliotek/cms-api": "*",
         "danskernesdigitalebibliotek/dpl-design-system": "0.0.0-dev",
         "danskernesdigitalebibliotek/dpl-react": "0.0.0-dev",
+        "danskernesdigitalebibliotek/dpl-design-system": "0.0.0-dev",
+        "danskernesdigitalebibliotek/dpl-react": "0.0.0-dev",
```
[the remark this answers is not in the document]
kasperbirch1: They will be deleted. It is only temporary to have the updated race/design system.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -415,6 +425,7 @@ public function work(string $wid): array {
       'rating-is-text' => $this->t('Rating of this item is', [], ['context' => 'Work Page']),
       'rating-text' => $this->t('out of', [], ['context' => 'Work Page']),
       'read-article-text' => $this->t('Read article', [], ['context' => 'Work Page']),
+      'reader-modal-description-text' => $this->t('modal for Reader', [], ['context' => 'Work Page']),
```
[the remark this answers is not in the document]
kasperbirch1: Sorry, this is no longer in use after the reader is on its own page and is no longer in a modal.

PR #1780, opened 2024-11-20
Ddfher 94 update webmaster modules 2

xendk:
#### Link to issue

[DDFHER-94](https://reload.atlassian.net/browse/DDFHER-94)

#### Description

Replaces the "add new module" button with a version that allows for module updates too.

[DDFHER-94]: https://reload.atlassian.net/browse/DDFHER-94?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/modules/custom/dpl_webmaster/src/Form/InstallOrUpdateModule.php
```diff
@@ -0,0 +1,283 @@
...
+      $this->messenger()->addError($this->t('%project is not a module.', ['%project' => $project]));
+      return;
+    }
+
+    if ($updater->isInstalled()) {
+      // Tell UpdateReady form which projects to update.
+      $this->session->set('update_manager_update_projects', [$project => $project]);
+      $form_state->setRedirect('update.confirmation_page');
+      return;
+    }
```
kasperg, line 243: Consider adding a code comments adding "markers" to  explain where the deviation from `UpdateManagerInstall` starts and ends.

This approach makes it hard to determine what has been changed compared to the original - and thus also hard to review.

web/modules/custom/dpl_webmaster/dpl_webmaster.info.yml
```diff
@@ -0,0 +1,8 @@
+name: Webmaster
+type: module
+description: 'Webmaster library support functionality'
+package: 'DPL'
+core_version_requirement: ^10
+
+dependencies:
+  - drupal:update
```
kasperg, line 8: :+1: for remembering to declare dependencies.

web/modules/custom/dpl_webmaster/dpl_webmaster.links.action.yml
```diff
@@ -0,0 +1,6 @@
+dpl_webmaster.install_or_update:
+  route_name: dpl_webmaster.install_or_update
+  title: 'Add new or update module'
+  weight: 25
```
kasperg, line 4: Consider adding a code comment explaining why this weight is necessary.

Weights appear as somewhat arbitrary numbers but they are often set to ensure that something occurs before or after another key point. Understanding this non-obvious relationship can be import in future work.

kasperg (approved):
+1: LGTM.

I have a couple of comments for you to consider before you merge.

In general it would be nice if we could have some form of automated test for this but I recognize that this will be tricky.

web/modules/custom/dpl_webmaster/dpl_webmaster.links.action.yml
xendk: You're right. Actually it's copied directly from update modules actions, but I can't see any obvious reason. I'll try to remove it.

kasperg (approved):

PR #1788, opened 2024-11-25
Upgrade Drupal Core to 10.3.10

kasperg:
#### Description

This update is relevant to address a performance degradation in release 10.3.9.


rasben (approved):
👍

PR #1791, opened 2024-11-25
No CORS + Https

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-226

#### Description

Make sure that we do not have cors restrictions in development
NOTE: We need to find the proper settings in PR/prod as well

Https is also being activated in docker compose setup. Documentation added for how to use it.


Dresse (approved):
LGTM 👍

PR #1806, opened 2024-11-27
Remove Config Filter module

kasperg:
#### Description

It has previous been uninstalled and can now be removed from the project codebase.

See #1798.


xendk (approved):
:+1:

PR #1809, opened 2024-11-28
Clarifying password management forms. DDFHER-149

rasben:
- Hide the core password suggestions, after password policy. DDFHER-149

The password policy module has it's own restrictions, and the core
password suggestions are very confusing.

- Add basic styling on "reset password" page. DDFHER-149

- Re-order the user edit form, clarifying password management. DDFHER-149

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -26,6 +26,8 @@ parameters:
     - '#Function .*_theme_suggestions_.*\(\) has parameter .* with no value type specified in iterable type array\.#'
     # Drupal hook_page_attachments functions uses page array which we cannot provide more detailed typing of.
     - '#Function .*_page_attachments\(\) has parameter \$page with no value type specified in iterable type array\.#'
+    # Drupal hook_module_implements_alter functions uses implementations array which we cannot provide more detailed typing of.
+    - '#Function .*_module_implements_alter\(\) has parameter \$implementations with no value type specified in iterable type array\.#'
     # Drupal hook_requirements() implementation returns array which we cannot provide more detailed typing of.
     - '#Function .*_requirements\(\) return type has no value type specified in iterable type array\.#'
     # Drupal hook_schema() implementation returns array which we cannot provide more detailed typing of.
--- a/web/modules/custom/dpl_admin/assets/dpl_admin.css
+++ b/web/modules/custom/dpl_admin/assets/dpl_admin.css
@@ -177,6 +177,20 @@
 }
 
 /**
+  We use the password policy module, which shows it's own password restrictions.
+  We want to hide the standard drupal password suggestions, as they are
+  confusing, as they may give conflicting information.
+  We need to make sure that inline styling also gets overriden, hence [style]
+  and !important.
+ */
+.user-form {
+  .password-suggestions,
+  .password-suggestions[style] {
+    display: none !important;
+  }
+}
+
+/**
  Hide default action button on confirm form.
  Editors should focus on the date-related changes.
  Selector is convoluted due to Gin DOM structure.
--- a/web/modules/custom/dpl_admin/dpl_admin.module
+++ b/web/modules/custom/dpl_admin/dpl_admin.module
@@ -216,8 +216,35 @@ function _dpl_admin_form_alter_node_page(array &$form, FormStateInterface $form_
     $form['field_hero_title']['#states']['invisible'] = $inactive_state;
     $form['field_subtitle']['#states']['invisible'] = $inactive_state;
   }
+}
 
 /**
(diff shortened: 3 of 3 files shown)


web/modules/custom/dpl_admin/dpl_admin.module
```diff
@@ -216,7 +216,54 @@ function _dpl_admin_form_alter_node_page(array &$form, FormStateInterface $form_
...
+ *
+ * Re-ordering the user edit form, to make the password demands clearer.
+ *
+ * @param array<mixed> $form
+ *   See the $form in dpl_admin_form_alter().
+ */
+function _dpl_admin_form_alter_user_form(array &$form, FormStateInterface $form_state, string $form_id): void {
+  $form['account']['current_pass']['#weight'] = 997;
+  $form['account']['pass']['#weight'] = 998;
+  $form['account']['password_policy_status']['#weight'] = 999;
```
kasperg, line 232: Please consider adding a comment about why these (very high) numbers are set.

When setting weights it is usually in relation to another value. Perhaps you just want these fields to appear last?

web/modules/custom/dpl_admin/assets/dpl_admin.css
```diff
@@ -176,6 +176,20 @@
...
+  confusing, as they may give conflicting information.
+  We need to make sure that inline styling also gets overriden, hence [style]
+  and !important.
+ */
+.user-form {
+  .password-suggestions,
+  .password-suggestions[style] {
+    display: none !important;
+  }
+}
```
kasperg, line 191: Please reconsider this solution.

Hiding the elements via CSS seems somewhat crude to me. Have you considered doing it in the existing form alter?

web/modules/custom/dpl_admin/dpl_admin.module
```diff
@@ -216,7 +216,54 @@ function _dpl_admin_form_alter_node_page(array &$form, FormStateInterface $form_
...
+  $form['#attributes']['class'][] = 'dpl-form';
+  $form['title'] = [
+    '#type' => 'markup',
+    '#prefix' => '<h1>',
+    '#suffix' => '</h1>',
+    '#markup' => t('Reset password', [], ['context' => 'DPL admin']),
+    '#weight' => '-999',
+  ];
+
+  $form['actions']['submit']['#attributes']['class'] = ['btn-primary', 'btn-filled btn-small', 'dpl-button'];
```
kasperg, line 266: Please consider if there is some configuration we have forgotten to add.

We use the [Gin Login module](https://www.drupal.org/project/gin_login) which already "provides a nice login screen based on either Claro or Gin for User Login, User Register, User Forgot Password". 

I think it would be appropriate if this also supports the "Reset password page". If not I think it would be a worthy contribution.

phpstan.neon.dist
```diff
@@ -26,6 +26,8 @@ parameters:
     - '#Function .*_theme_suggestions_.*\(\) has parameter .* with no value type specified in iterable type array\.#'
     # Drupal hook_page_attachments functions uses page array which we cannot provide more detailed typing of.
     - '#Function .*_page_attachments\(\) has parameter \$page with no value type specified in iterable type array\.#'
+    # Drupal hook_module_implements_alter functions uses implementations array which we cannot provide more detailed typing of.
+    - '#Function .*_module_implements_alter\(\) has parameter \$implementations with no value type specified in iterable type array\.#'
```
kasperg, line 30: :+1:

kasperg (changes requested):
:+1: I think there are a few aspects of this that warrants a second thought before merging this as is.

web/modules/custom/dpl_admin/dpl_admin.module
rasben: I get the confusion. `gin_login` does provide styling for /user/password, but what I'm styling is the page you end up on, when you're clicking on the link that Drupal sends to your email:

Screenshot 2024-12-03 at 12 01 05

`/user/reset/1/XXXXXXX`

kasperg: Thanks for clarifying.

I still think it would be a worthy contribution to add support for the "Reset password page" for Gin Login instead of keeping this in the frontend theme.

rasben: I've added an issue here:
https://www.drupal.org/project/gin_login/issues/3491373

Are you saying that I should also create a patch, as part of this pull request/task?

kasperg: That was what I had hoped for compared to the current solution, but I do not know how difficult it would be to achieve :smile:

PR #1813, opened 2024-12-02
Logging custom DDF events, such as updates to users. DDFHER-150

rasben:
When updating a user, we want to log the changes done - this is part of a security policy by DDF.
For most fields, the old and new value gets logged, but others may be ignored, or declared on a meta-level (such as password).

#### Link to issue

https://reload.atlassian.net/browse/DDFHER-150

An example of how the logging looks like:

User account updated: local_administrator3 (6) has been updated with the following values: Foretrukken sprogkode (preferred_langcode): en changed to da | Foretrukken sprogkode til administration (preferred_admin_langcode): da changed to en | Name (name): local_administrator2 changed to local_administrator3 | Password was updated | E-mail (mail): test2@test.dk changed to test3@test.dk | New roles added to user: local_administrator | Old roles removed from user: external_system | Author name (field_author_name): test2 changed to test3

### DIFF
--- a/web/modules/custom/dpl_logging/dpl_logging.module
+++ b/web/modules/custom/dpl_logging/dpl_logging.module
@@ -1,0 +1,112 @@
+<?php
+
+use Drupal\Component\Utility\DiffArray;
+use Drupal\user\Entity\User;
+
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ *
+ * When updating a user, we want to log the changes done - this is part of
+ * a security policy by DDF.
+ * For most fields, the old and new value gets logged, but others may be
+ * ignored, or declared on a meta-level (such as password).
+ */
+function dpl_logging_user_presave(User $user): void {
+  $original_user = $user->original;
+
+  if (!($original_user instanceof User)) {
+    return;
+  }
+
+  $original_values = $original_user->toArray();
+  $values = $user->toArray();
+  $updates = DiffArray::diffAssocRecursive($values, $original_values);
+
+  // Some fields are so meta that we do not need to log their changes.
+  $ignored_updates = ['changed', 'metatag', 'field_password_expiration'];
+  $diffs = [];
+
+  foreach ($updates as $key => $value) {
+    if (in_array($key, $ignored_updates)) {
+      continue;
+    }
+
+    if ($key === 'roles') {
+      $original_roles = $original_user->getRoles();
+      $new_roles = $user->getRoles();
+
+      // This results in the machine names of the roles.
+      // We could also do a look up, to find the translated labels also,
+      // but it is probably overkill for what this log will be used for.
+      $added_roles = array_diff($new_roles, $original_roles);
+      $removed_roles = array_diff($original_roles, $new_roles);
+
+      if (!empty($added_roles)) {
+        $diffs['roles_added'] = t(
+          'New roles added to user: @roles',
+          ['@roles' => implode(', ', $added_roles)],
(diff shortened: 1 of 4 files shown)


web/modules/custom/dpl_logging/dpl_logging.module
```diff
@@ -0,0 +1,113 @@
...
+
+    if ($key === 'roles') {
+      $original_roles = $original_user->getRoles();
+      $new_roles = $user->getRoles();
+
+      // This results in the machine names of the roles.
+      // We could also do a look up, to find the translated labels also,
+      // but it is probably overkill for what this log will be used for.
+      $added_roles = array_diff($new_roles, $original_roles);
+      $removed_roles = array_diff($original_roles, $new_roles);
```
kasperg, line 43: I agree with this. I would argue the same this for the fields but I see you went with both labels and keys there.

web/modules/custom/dpl_logging/dpl_logging.module
```diff
@@ -0,0 +1,113 @@
...
+  $diffs = [];
+
+  foreach ($updates as $key => $value) {
+    if (in_array($key, $ignored_updates)) {
+      continue;
+    }
+
+    if ($key === 'roles') {
+      $original_roles = $original_user->getRoles();
+      $new_roles = $user->getRoles();
```
kasperg, line 37: Another more object-oriented and functional approach to achieving this could be in somewhat pseudo-code

```php
$fields = $user->getFields(include_computed: FALSE);
$ignored_fields = ['changed', 'metatag', 'field_password_expiration'];
$relevant_fields = array_filter($fields, ($field) => !in_array($field->getName(), $ignored_fields));
$updated_fields = array_filter($fields, ($field) => !$field->equals($original->get($field->getName()));
$diffs = array_map($updated_fields, ($field) => {
  return = t(
      '@label (@key): @original_value changed to @value',
      [
        '@label' => $label,
        '@key' => $field->getName(),
        '@original_value' => $original->get($field->getName())->toString(),
        '@value' => $field->toString(),
      ],
      ['context' => 'DPL logging']
    )->render();
})
```

To me we get a more structured approach to data handling by not ejecting from object to arrays so early. This also means that we cannot use `DiffArray` but as I see it `getFields()` and `equals()` on `FieldItemListInterface` achieves the same thing.

In the end it also means that we do not have to rely on `[0]['value'] ?? NULL` for data retrieval.

web/modules/custom/dpl_logging/dpl_logging.module
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      continue;
+    }
+
+    // Getting human-readable label, with machine name as fallback.
+    $label = $user->getFieldDefinition($key)?->getLabel() ?? $key;
+
+    // Make sure the label is a string. Sometimes getLabel returns translation.
+    $label = ($label instanceof TranslatableMarkup) ? $label->render() : $label;
```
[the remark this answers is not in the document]
rasben: ooooh that is awesome!

web/modules/custom/dpl_logging/dpl_logging.module
rasben: I get your point, but generally speaking, I prefer my way, as I feel it's more readable

kasperg: I understand that the functional approach (array_filter, array_map etc.) requires some getting used to. You could use the object oriented features from Drupals classes combined with classic loops just as well.

PR #1827, opened 2024-12-04
Add docs about permissions and webmaster libraries

xendk:
Just some doc updates.

### DIFF
--- a/docs/permissions-and-roles.md
+++ b/docs/permissions-and-roles.md
@@ -1,0 +1,31 @@
+# Permissions and roles
+
+'mediator':
+
+* For users who frequently create dynamic content on the website.
+* Can create and edit events, articles, pages, and campaigns in
+  /admin/content.
+
+'editor':
+
+* For users who, in addition to content creation, are responsible for
+  static content (e.g., opening hours) and the site's structure.
+* Can coordinate and manage the work of mediators.
+* Has access to the same as mediator plus all content types in
+  /admin/content, as well as taxonomy and menus in /admin/structure.
+
+'local_administrator':
+
+* For users who need to configure the website in terms of settings,
+  appearance, integrations, general information, library usage, etc.
+* Has access to the same as editor plus the site's configuration
+  options in /admin/config, user creation in /admin/people, and
+  changing the site's appearance in /admin/appearance.
+* This is the role assigned to trusted users on regular libraries.
+
+'administrator':
+
+* Can modify the site beyond the CMS's standard settings, e.g., by
+  enabling and disabling modules.
+* Has access to everything in the CMS.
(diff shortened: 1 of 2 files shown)


docs/permissions-and-roles.md
```diff
@@ -0,0 +1,31 @@
...
+  /admin/content, as well as taxonomy and menus in /admin/structure.
+
+'local_administrator':
+
+* For users who need to configure the website in terms of settings,
+  appearance, integrations, general information, library usage, etc.
+* Has access to the same as editor plus the site's configuration
+  options in /admin/config, user creation in /admin/people, and
+  changing the site's appearance in /admin/appearance.
+* This is the role assigned to trusted users on regular libraries.
```
kasperg, line 24:
```suggestion
* This is the role assigned to trusted users on Editor libraries.
```

Please consider referring to [the name of the plan used by the business and elsewhere in the docs](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/5bb77ed095e4690871a51fd65694023d62a17a61/docs/configuration-management.md#L14-L14). 

The name overlaps with the user role but in practice I do not think this has been a problem.

docs/permissions-and-roles.md
```diff
@@ -0,0 +1,31 @@
+# Permissions and roles
+
+'mediator':
```
kasperg, line 3:
```suggestion
## `mediator`:
```

Just a suggestion to use headlines as this seems to be what it is. Take it or leave it 😄

docs/webmaster-libraries.md
```diff
@@ -0,0 +1,14 @@
...
+administration pages, and can install custom modules.
+
+See [Permissions and roles](./permissions-and-roles) for details on
+Drupal setup.
+
+In addition to the Drupal setup, the site should use the `webmaster`
+plan when configured in [DPL
+Platform](https://danskernesdigitalebibliotek.github.io/dpl-docs/DPL-Platform/),
+this creates a `moduletest` environment for testing and implies a
+slightly different deployment scheduling.
```
kasperg, line 14: :+1: I appreciate this addition. Consider moving it to `dpl-docs`. 

As I see it the plan handling is handled partly by the CMS and partly with the Platform. Now what we have a general location within the documentation I think this would be a nice place to put this information and then refer to the parts within the individual projects.

With this in mind I also think it would make sense to provide an overview of all the plan types [a bit like we have done elsewhere](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/5bb77ed095e4690871a51fd65694023d62a17a61/docs/configuration-management.md#L14-L14) instead of only mentioning the webmaster libraries.

kasperg (approved):
:+1: I have some some suggestions but I think you can resolve this as you prefer.

PR #1831, opened 2024-12-06
Make EventWrapper into a bundle class instead

xendk:
Using a bundle class skips the middleman.

There's methods on `EventRestMapper` that should really be moved to `EventInstance`, but that's for a followup.


kasperg (approved):
I think this is a very nice change.

I think it would be good to get others to read this as well. Especially @rasben.

PR #1851, opened 2024-12-10
Add translation strings for share button labels

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/KBHBIB-24

#### Description

Added translation strings for Share on Facebook and Copy link buttons


composer.json
```diff
@@ -40,10 +40,10 @@
             "type": "package",
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-react",
-                "version": "2024.50.0",
+                "version": "0.0.0-dev",
```
Adamik10, line 43: These changes in composer.json & composer.lock shouldn't be coming in.

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -437,6 +437,10 @@ public function work(string $wid): array {
       'type-text' => $this->t('Type', [], ['context' => 'Work Page']),
       'we-have-shopped-text' => $this->t('In stock:', [], ['context' => 'Work Page']),
       'you-have-borrowed-text' => $this->t('You have borrowed', [], ['context' => 'Work Page']),
+      'share-on-facebook-text' => $this->t('Share on Facebook', [], ['context' => 'Work Page']),
```
Adamik10, line 440: Just to make sure - these buttons only appear on the work page, right?

Adamik10 (changes requested):
I got something that should be changed :))

composer.json
JacobArrow: It is a temporary commit for for demoing purposes, as you can see in the commit message

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
JacobArrow: For now, yes

PR #1874, opened 2024-12-17
Update BnfImporter to also import basic paragraphs.

rasben:
This feels a bit dirty, but for now, until we've cleared how to deal with GraphQL schemas/codegen, it'll do.
Only text_body paragraphs are imported.

### DIFF
--- a/web/modules/custom/bnf/src/Services/BnfImporter.php
+++ b/web/modules/custom/bnf/src/Services/BnfImporter.php
@@ -4,12 +4,18 @@ namespace Drupal\bnf\Services;
 
 use Drupal\bnf\BnfStateEnum;
 use Drupal\bnf\Exception\AlreadyExistsException;
+use Drupal\Core\Entity\EntityFieldManagerInterface;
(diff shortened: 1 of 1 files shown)


web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -20,46 +26,235 @@
...
     $query = <<<GRAPHQL
     query {
       $queryName(id: "$uuid") {
         title
+        paragraphs {
+
+    GRAPHQL;
+
+    // Loop through allowed paragraphs and add their structures.
+    foreach (self::ALLOWED_PARAGRAPHS as $graphBundle => $drupalBundle) {
```
xendk, line 90: I'd suggest to fold this out in a helper method that deals with the "inside" of the paragraphs section.

I think at the end we'll end up with a `getQuery` that just calls `getFields` which recursively generates the needed GraphQL.

xendk (approved):

PR #1876, opened 2024-12-18
DDFBRA-284 - Create basic page for go and add video paragraph

Dresse:
#### Link to issue

I decided to consolidate 2 JIRA tasks in the same PR as they are closely related.

https://reload.atlassian.net/browse/DDFBRA-284
https://reload.atlassian.net/browse/DDFBRA-277

#### Description

This PR adds a new contenttype called 'GO Page' which we will use for GO specific pages that cannot be categorised into either article or category pages. An example of this could be the frontpage. 
The PR adds permissions similar to what the current "Page" contenttype is using. 

Furthermore this PR adds a new GO specific paragraph called 'go_video', which is an extended copy of the existing 'video' paragraph (adding a title field). It will likely be extended with more fields in the future.
We will be using this paragraph type specifically for GO content pages (such as the GO Page). 

Lastly the PR enables the new 'GO Page' and 'go_video' paragraph in GraphQL Compose.

#### Screenshot of the result

Example of a GraphQL query that fetches an GO Page and its associated GO video paragraph and its fields:

Screenshot 2024-12-18 at 10 51 41


kasperbirch1 (approved):
From my knowledge, I think it looks good. 👍  Have you considered whether it would be an advantage to have a "go-paragraphs" page in the default content? like we have "/paragraph-page"?

I'll leave it up to you if you need someone else's opinion on the grapql part.

Dresse:
> From my knowledge, I think it looks good. 👍 Have you considered whether it would be an advantage to have a "go-paragraphs" page in the default content? like we have "/paragraph-page"?
> 
> I'll leave it up to you if you need someone else's opinion on the grapql part.

I did not think of that, but it could be a good idea to include this. This would make it easier for the FE team to work with. 
Do you know if it is possible to add more paragraphs to the page whenever they become available? Or do you need to create a new default content page each time?

PR #1882, opened 2024-12-20
Improve dev https setup

xendk:
Move `https` service into `docker-compose.yml` and tell Lagoon to ignore it. Remove and clean up after `docker-compose.dev.yml`.

Drop defaulting `COMPOSE_PROJECT_NAME` sporadically. It should be used everywhere to make sense, but we wouldn't want to pretend to support a docker compose so old that it doesn't define the variable.


spaceo (approved):
LGTM 👍

PR #1891, opened 2024-12-25
Handle graphql consumers for both BNF and Go

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-290

#### Description

This PR handles:
* Deletion of duplicate graphq users
* Handling of separate consumers, users and roles  for BNF and Go


web/modules/custom/dpl_consumers/src/Services/ConsumerHandler.php
```diff
@@ -0,0 +1,87 @@
...
+use Drupal\dpl_consumers\ConsumerRole;
+use Drupal\dpl_consumers\ConsumerUser;
+use Psr\Log\LoggerInterface;
+
+/**
+ * DPL consumer handler.
+ *
+ * This class is responsible for creating consumers and users.
+ */
+class ConsumerHandler {
```
xendk, line 17: I'm not fond of this class. What is a "ConsumerHandler" anyway? 

On one hand it's a composition of consumer, user and role. On the other hand `setComponents` allows for setting new consumer, user and role.

In my mind it should either be a "handler", state free, and having `create` and `delete` just take the consumer, user and role as arguments.

Or it should be an object representing *this* combination (`ConsumerSetup` perhaps?), in which case it should take consumer, user and role as constructor arguments (and then you'd need a factory in order to get the container to inject the logger).

web/modules/custom/dpl_consumers/dpl_consumers.deploy.php
```diff
@@ -22,8 +23,28 @@
...
+/**
+ * Expand the users, roles, consumers so they can handle both BNF and GO.
+ *
+ * So we delete previous entities and create new ones.
+ */
+function dpl_consumers_deploy_10002(): void {
+  // Delete consume and users that we want to handle differently.
+  // We want to create consumers and consumer users
+  // specifically for the two known consumers (BNF and Go) and connected users.
+  (new Consumer("graphql_consumer", \Drupal::entityTypeManager()))->delete();
```
xendk, line 40: Maybe `ConsumerHandler` should have methods for getting `Consumer` and `ConsumerUser` so the DI can be passed to it? Having to create a repository service for each of them is a bit overkill, but I think having to pass `entityTypeManager` to each `new` is a bit of a DX wart.

xendk (changes requested):

xendk (approved):
:+1:

PR #1910, opened 2025-01-05
Make sure to use https in drush user login cmd

spaceo:
In the end of the `task dev:reset` cmd it is annoying to get the http version of the one time login url because http is currently not supported.


Dresse (approved):
LGTM 👍

xendk (changes requested):
Think it would be better to tell Drush to use https.

PR #1929, opened 2025-01-07
BNF: send along canonical data as part of import. DDFHER-187

rasben:


web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -302,8 +305,10 @@ public function importNode(string $uuid, string $endpointUrl, string $nodeType =
     $nodeStorage = $this->entityTypeManager->getStorage('node');
     try {
       $nodeData = $this->loadNodeData($uuid, $endpointUrl, $nodeType);
+
       $nodeData['type'] = $nodeType;
       $nodeData['uuid'] = $uuid;
+      $nodeData['field_canonical_url']['uri'] = $nodeData['canonicalUrl']['url'] ?? NULL;
```
xendk, line 311: This just copies an existing canonical URL, how do it get populated in the first place?

xendk (changes requested):

xendk (approved):

PR #1935, opened 2025-01-08
DDFBRA-321 - Bump `drupal/facets` from "2.0.9" to "3.0.0"

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-321

#### Description

Steps taken:
- Ran `composer require drupal/facets:^3.0`
- Executed `drush updatedb`
- Rebuilt cache using `drush cache:rebuild`
- Exported configuration with `drush config:export --diff`

No changes were detected in the database or configuration files as a result of the update.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -120,7 +120,7 @@
         "drupal/entity_clone_template": "^2.0",
         "drupal/entity_reference_revisions": "^1.11.0",
         "drupal/enum_field": "^1.0",
-        "drupal/facets": "^2.0",
+        "drupal/facets": "^3.0",
         "drupal/field_group": "^3.4",
         "drupal/field_inheritance": "^2.0",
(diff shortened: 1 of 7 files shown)


kasperg (commented):
@kasperbirch1 https://www.drupal.org/project/facets has a large notice regarding Facets 3 which seems relevant to our case as we use AJAX for our facets:

> Facets 3.x only supports Search API, and introduces "[Facets exposed filters](https://project.pages.drupalcode.org/facets/exposed_filters/)" next to "Facet blocks".
> We recommend using Facets exposed filters for new projects, when using Facets + Views
> Using "Facets exposed filters" in Facets 3 has full AJAX support.
>
> Facets 3 will not support AJAX for "Facet blocks".

Have you looked into this?

PR #1937, opened 2025-01-08
Ddfher 145 dev setup adjustments

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-161

#### Description

Fixes BNF to use the new GraphQL creds, and makes BNF server URL configurable.


rasben (commented):
Overall looks good, but, how does the endpoint URL get set locally? I dont see any default value, unless I'm blind?

xendk:
> I dont see any default value, unless I'm blind?

web/modules/custom/bnf/bnf_client/config/install/bnf_client.settings.yml

But it'll get overridden by the docker setup when I get to https://reload.atlassian.net/browse/DDFHER-186

rasben (approved):

PR #1953, opened 2025-01-10
BNF: Test environment for JIRA tasks.

rasben:
Setting the BNF details on the bnf_server, so it is easier to discern the client from server.

Also fixes an issue, so pushing to an existing BNF site doesnt cause a deploy fail when the module is not enabled.


rasben:
Something strange has been happening with Lagoon and the reporting of GH deployments.
But, it HAS been deployed:

BNF: https://varnish.pr-1953.dpl-bnf.dplplat01.dpl.reload.dk/frontpage
CMS: https://varnish.pr-1953.dpl-cms.dplplat01.dpl.reload.dk/frontpage

PR #1956, opened 2025-01-10
Add category term name as notice in opening hours legacy API

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-177

#### Description

The Redia app requires opening hours to have a category for them to be displayed. Currently we only return the category id. However new categories are not likely to be the same as previously. Also we do not provide an API for exposing categories with their ids and titles.

To address this in a simple way we add the category name as a notice for each instance.

It turns out that the Redia app will display opening hours with the notice if the notice is defined no matter what the category id is. Consequently this is a rather simple way to communicate the information.


kasperg:
This has been approved by Redia.

xendk (approved):

PR #1968, opened 2025-01-14
DDFBRA-281 - Create go category page

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-281

#### Description

This PR adds a new content type called "GO Category Page" which will be used on the GO site. 

The new content type is very simple and for now only consist of a title and paragraphs reference field. Furthermore a new details sidebar menu has been added called "Category Info" which consist of 4 fields containing the info needed in the "Kategori bånd" (Title, Image, Color and a optional audio field).

To add a new sound field, we also had to enable the "Audio" media type. 

Both the new media type and the "GO Category Page" had to be enabled through GraphQL Compose. 
The color field type is provided by a contrib module called "Color Field". The field type do not have a GraphQL Compose resolver or schema type associated with it, thus I have added one to the "dpl_graphql" module.

We want to use routes when querying content through GraphQL. This has been enabled through the GraphQL Compose Routes module.

A new role "GO Editor" has been added, which will be used for handling GO specific content/settings. A new custom permission has been added to the UniLogin configuration and granted to the new "GO Editor" role. 
We also cleaned up and removed the "Bypass content access control" permission and instead are now using content specific permissions for each content type.

Two additional comments:

1. This task does only enable the querying of the GO Category Page. It does not implement a way of fetching all existing GO Category Pages and its fields. This will be part of another task.

2. There is a separate [task](https://reload.atlassian.net/browse/DDFBRA-341) for setting up a new GO Editor role as well as hiding "GO specific" content/settings from all other roles than the GO Editor role, but I have decided to add the changes to this PR instead of a separate one, as it was the easiest thing to do.


spaceo (commented):
I am sorry if I misinterpret something, because I am reviewing on my phone.
I think the content type should be called:
Label: Go Category
Machine name: go_category
Because of its multiple purposes of both categorizing content and serving a page.

spaceo (changes requested):
Please see my comment

Dresse:
> Please see my comment

I have read your reply and renamed "GO Category Page" into just "GO Category". 
The same goes for filenames, machine names and labels.

I wasn't sure that we landed on the term "GO Category" instead of "GO Category Page", as I think me and Thomas was vouching for the other, but maybe that was when we where also talking about having a separate "GO Categories" vocabulary? Anyways, it has been renamed now. Please take another look! :) 👍

spaceo (approved):
LGTM 👍

PR #1969, opened 2025-01-14
Display metadata on eventseries, about related eventinstances. KBHBIB-47

rasben:
We use LD json to display related eventinstances in a format that Cludo (and others) can use to index correctly.
It gets added as a `<script>` tag on the HTML `<head>` element, following schema as defined by https://json-ld.org/

#### Link to issue

https://reload.atlassian.net/browse/KBHBIB-47

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -167,7 +167,7 @@ services:
       << : *default-environment
 
   node:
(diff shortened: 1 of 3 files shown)


Dresse (approved):
LGTM 👍

PR #1970, opened 2025-01-14
Ddfher 186 bnf dev setup

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-186

#### Description

Add setup for running both a CMS and a BNF site in one environment.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -31,7 +31,9 @@ vars:
   # Where is the docker file(s) we use for our builds residing?
   LAGOON_DIR: "lagoon"
 
-  DOCKER_COMPOSE_FILES_DEFAULT: "-f docker-compose.yml"
+  DOCKER_COMPOSE_FILES_BNF:
+    sh: if [[ -f web/sites/sites.php ]]; then echo "-f docker-compose.bnf.yml"; fi
+  DOCKER_COMPOSE_FILES_DEFAULT: "-f docker-compose.yml {{ .DOCKER_COMPOSE_FILES_BNF}}"
   DOCKER_COMPOSE_FILES: "{{.DOCKER_COMPOSE_FILES | default .DOCKER_COMPOSE_FILES_DEFAULT }}"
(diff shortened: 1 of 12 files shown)


Taskfile.yml
```diff
@@ -101,6 +117,13 @@ tasks:
     cmds:
       - docker compose {{ .DOCKER_COMPOSE_FILES }} run --rm cli sh -c "{{.CLI_ARGS}}"
 
+  dev:bnfcli:
+    summary: dev:cli for BNF site, a noop if BNF not enabled
+    status:
+      - "[[ -z \"{{ .DOCKER_COMPOSE_FILES_BNF}}\" ]]"
```
spaceo, line 123: I think it would be nice with and error to the end user if BNF is not enabled when using this cmd.

spaceo: And I can see now it is because you use it internally in the Taskfile. The thing is that we use "dev:cli" directly quite a lot in our team. So, as it is with these changes, if we call `task dev:bnfcli` nothing would happen and the developer would not know why, right?

assets/sites.php
```diff
@@ -0,0 +1,13 @@
...
+
+$sites = [];
+
+// Add in the host name to sites if it starts with `bnf-`. This way we don't
+// have to worry about what TLD people use (docker or local), nor the composer
+// project name (dpl-cms per default, but could be anything).
+$host = $_SERVER['HTTP_HOST'];
+if (preg_match('/^bnf-/', $host)) {
+  $sites[$host] = 'bnf';
+}
```
spaceo, line 13: I think this prefix rule should be documented somewhere so people know it.

spaceo (commented):
I both like and dislike the approach taken here :)
At one point I think it solves the problem of managing the two sites in develop, nicely.

We keep on bragging about the genius idea about using the same Drupal instance for dpl-cms and bnf and the automatic benefits - which I agree upon.
But here we have two docker compose files that are almost identical and we have two settings.php files. And as far as I see it, there is quite a big risk that we forget something maintaining it in the future

assets/sites.php
xendk: Ah, yes, but where?

PR #1981, opened 2025-01-15
Make sure the Go node app is persistent

spaceo:
#### Description

We have the experience from the Go Demo environment that having a persitent .next dir helps the build process and makes sure that k8s doesn't try to rebuild it again and again.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -170,9 +170,10 @@ services:
   node:
     image: ghcr.io/danskernesdigitalebibliotek/dpl-go-node:0.0.12
     labels:
-      lagoon.type: node
       provenance: false
-
+      lagoon.type: node-persistent
+      lagoon.persistent: /app/.next
+      lagoon.persistent.size: 500Mi
 volumes:
   projectroot:
     driver: local
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -173,6 +173,10 @@ services:
       provenance: false
       lagoon.type: node-persistent
       lagoon.persistent: /app/.next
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -170,9 +170,10 @@ services:
   node:
     image: ghcr.io/danskernesdigitalebibliotek/dpl-go-node:0.0.12
     labels:
-      lagoon.type: node
       provenance: false
-
+      lagoon.type: node-persistent
```
kasperg, line 174: Beware of this change. According to [the Lagoon docs](https://docs.lagoon.sh/concepts-advanced/service-types/): 

> Once a lagoon.type is defined and the environment is deployed, changing it to a different type is not supported and could result in a broken environment.

I am not sure whether it is a problem here.

spaceo: I believe it is only a problem if you introduce it in an existing environment - eg. pushing a commit with the change in an already deployed environment

spaceo: In production environments the service does not exist yet.

docker-compose.yml
```diff
@@ -170,9 +170,10 @@ services:
...
+      lagoon.persistent: /app/.next
+      lagoon.persistent.size: 500Mi
```
kasperg, line 176: Consider adding a comment about why `500Mi` is a reasonable size. 

:+1: for setting a size lower than the default `5Gi` - even though we only pay for what is used. I wonder if there is a reason why you set it at this level. I could not spot anything in the commit comment.

kasperg (approved):

PR #1994, opened 2025-01-21
Add graphql query ro aquire tokens from Adgangsplatformen

spaceo:
#### Link to issue

[DDFBRA-354
](https://reload.atlassian.net/browse/DDFBRA-354)

#### Description
Adds graphql query ro aquire tokens from Adgangsplatformen
The idea is that the Drupal session token controls wether the client can access the library or the user token. Like it is the case with the /dpl-react/user-tokens request used by the dpl-react project.


config/sync/user.role.patron.yml
```diff
@@ -6,11 +6,13 @@ dependencies:
...
     - rest.resource.proxy-url
   module:
+    - graphql
     - rest
 id: patron
 label: Patron
 weight: 6
 is_admin: null
 permissions:
+  - 'execute graphql_compose_server arbitrary graphql requests'
```
Dresse, line 16: Is this permission necessary? Aren't we supposed to only use the created Consumer Users for this?

web/modules/custom/dpl_library_token/graphql/dpl_library_token.base.graphqls
```diff
@@ -0,0 +1 @@
+
```
Dresse, line 1: This is just an empty file, do we need it for something?

Dresse (approved):
A few comments but LGTM 👍 

Should we refactor the DplConfiguration to do something similar to this?

config/sync/user.role.patron.yml
spaceo: This is because when we are executing queries asking for a user token we use the session cookie which is connected to the authenticated patron

web/modules/custom/dpl_library_token/graphql/dpl_library_token.base.graphqls
spaceo: the graphql module demands that a *.base.graphqls file exists the moment you use a *.extension.graphqls

PR #1998, opened 2025-01-21
Log state changes

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-196

#### Description

This PR adds support for logging changes to Drupal State API - most importantly `system.maintenance_mode`.

With this change enabling maintenance mode will be logged like this:

`State changed: system.maintenance_mode changed from NULL to 1`

I recommend going through the change commit by commit.

One might question the addition of [the State Log module](https://www.drupal.org/project/state_log). It is not very widely used nor actively maintained. There is no Drupal 11 support at the moment. 

However, as I see it the code structure of the module looks like I would have developed it and does nothing more than needed. If we need anything we can provide patches and hope for a merge. As such I think it is OK to adopt. If we really need it we could apply for maintainer access for the module. I think it could be helpful for us when managing so many sites.


config/sync/state_log.settings.yml
Dresse, line 1: Why the incrementation of 2 for each ignored state? For better readability? Just curious.

Dresse (approved):
LGTM 👍 Just have a single comment with a question. 

Oh, also; Are we planning on keeping the module after locating the "maintenance mode issue"? I think it could be nice to have en general.

config/sync/state_log.settings.yml
kasperg: I do not know. I bet this is a result of data handling by the State Log Module.

kasperg:
> Are we planning on keeping the module after locating the "maintenance mode issue"? I think it could be nice to have en general.

I agree. I would like to keep it around.

config/sync/state_log.settings.yml
xendk: @Dresse 
The submit handler on the settings form:

```php
  /**
   * {@inheritdoc}
   */
  public function submitForm(array &$form, FormStateInterface $form_state) {
    $ignore_settings_array = preg_split("[\n|\r]", $form_state->getValue('log_ignored_state'));
    $ignore_settings_array = array_filter($ignore_settings_array);

    $this->config('state_log.settings')
      ->set('log_ignored_state', $ignore_settings_array)
      ->save();

    parent::submitForm($form, $form_state);
  }
```

The regexp used in preg_split will split the submitted string on newlines and carriage return. Form input is normalized to CRLF no matter what the local operating system uses, so the split data will include empty "lines" the newline and the carriage return. These are then removed by the `array_filter`, however, it'll preserve the key of the values, which is why they're a power of two.

xendk (approved):

config/sync/state_log.settings.yml
Dresse: Ah okay, guess that makes sense. 👍

PR #1999, opened 2025-01-22
DDFBRA-396 - Added update hook for installing "graphql_compose_routes" module.

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-396

#### Description

Added update hook for installing "graphql_compose_routes" module.

The module was enabled in config core.extensions.yml in a previous [PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1968).


Dresse:
> Do you need to remove it from core.extensions?

We usually have it both places actually. So I don't think it should be removed from core.extensions.yml

spaceo (approved):
LGTM 👍

PR #2004, opened 2025-01-24
Ddfher 182 graphql client

xendk:
The great rewriting (in honour of @rasben).


web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -264,101 +67,40 @@ public function loadData(string $uuid, string $endpointUrl, string $nodeType = '
...
-      $node->save();
+      $node = $this->mapperManager->map($nodeData);
 
       $node->set(BnfStateEnum::FIELD_NAME, BnfStateEnum::Imported->value);
+
+      $node->set('status', NodeInterface::NOT_PUBLISHED);
+
+      // @todo Deal with canonical URL.
+      // CMS -> BNF: Set to site URL.
+      // BNF -> CMS: Copy whatever BNF says.
```
rasben, line 89: I've previously "dealt" with Canonical URLs, and closed it as part of [DDFHER-187](https://reload.atlassian.net/browse/DDFHER-187)

So, either a new jira ticket has to be created, or, this todo needs to be mended as part of this pull request

web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -399,3 +399,10 @@ function dpl_update_update_10035(): string {
 function dpl_update_update_10036(): string {
   return _dpl_update_install_modules(['state_log']);
 }
+
+/**
+ * Adding autowire_plugin_trait for BNF.
+ */
+function dpl_update_update_10037(): string {
+  return _dpl_update_install_modules(['autowire_plugin_trait']);
```
rasben, line 407: Is it a problem that this module is enabled before we go live with BNF?

web/modules/custom/bnf/tests/src/Unit/ParagraphsExposedTest.php
```diff
@@ -0,0 +1,45 @@
...
+   * paragraph type on a node. This results in an unhelpful JSON parsing error
+   * on the receiving side, so we try to catch the problem earlier.
+   */
+  public function testAllParagraphsEnabled(): void {
+    $config = Yaml::parseFile(DRUPAL_ROOT . '/../config/sync/graphql_compose.settings.yml');
+
+    $this->assertArrayHasKey(
+      'entity_config',
+      $config,
+      'Could not find `entity_config` in graphql_compsoe config file'
```
rasben, line 31:
```suggestion
      'Could not find `entity_config` in graphql_compose config file'
```

web/modules/custom/bnf/tests/src/Unit/ParagraphsExposedTest.php
```diff
@@ -0,0 +1,45 @@
...
+    );
+
+    $this->assertArrayHasKey(
+      'paragraph',
+      $config['entity_config'],
+      'Could not find `entity_config` in graphql_compsoe config file'
```
rasben, line 37:
```suggestion
      'Could not find `entity_config` in graphql_compose config file'
```

docs/bnf-import-export.md
```diff
@@ -0,0 +1,123 @@
+# BNF: Content import/export
```
rasben, line 1: Thank you for adding this documentation 👍

rasben (changes requested):

web/modules/custom/bnf/schema/bnf.graphql
```diff
@@ -0,0 +1,1840 @@
+"""Complex address data."""
```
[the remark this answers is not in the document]
xendk: Oh, right, good catch. It's pulled directly from the GraphQL endpoint.

The language looks a bit random, but that's an issue with `graphql_compose` I suppose. But it doen't matter to us.

web/modules/custom/bnf/src/MangleUrl.php
```diff
@@ -0,0 +1,55 @@
...
+
+declare(strict_types=1);
+
+namespace Drupal\bnf;
+
+use function Safe\parse_url;
+use function Safe\preg_match;
+
+/**
+ * Mangle URLs in the development docker setup.
```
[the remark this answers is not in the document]
xendk: Yes. Related to [name mangling](https://en.wikipedia.org/wiki/Name_mangling) and [filename mangling](https://en.wikipedia.org/wiki/Filename_mangling).

web/modules/custom/dpl_update/dpl_update.install
xendk: No, it only provides a PHP trait. Haven't brought it up with @arnested, but it could just have been a composer package (thought that wouldn't provide it any visibility on drupal.org).

web/modules/custom/bnf/tests/src/Unit/ParagraphsExposedTest.php
xendk: :+1:

web/modules/custom/bnf/tests/src/Unit/ParagraphsExposedTest.php
```diff
@@ -0,0 +1,45 @@
...
+    $this->assertArrayHasKey(
+      'entity_config',
+      $config,
+      'Could not find `entity_config` in graphql_compose config file'
+    );
+
+    $this->assertArrayHasKey(
+      'paragraph',
+      $config['entity_config'],
+      'Could not find `entity_config` in graphql_compose config file'
```
kasperg, line 37:
```suggestion
      'Could not find `paragraph` entry within `entity_config` in in graphql_compose config file'
```

web/modules/custom/bnf/tests/src/Unit/Mapper/ParagraphAccordionTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Test text paragraph mapping.
+   */
+  public function testParagraphTextBodyMapping(): void {
+    $this->storageProphecy->create([
+      'type' => 'accordion',
+    ])->willReturn($this->entityProphecy);
+
+    $mapper = new ParagraphAccordion([], '', [], $this->entityManagerProphecy->reveal());
```
kasperg, line 29: Consider adding the mapper `Mapper` suffix to mapper implementations.

I know it is duplicated from the namespace but the current code reads a bit weird. I am a bit conflicted by my suggestion but I guess this is also why it is mentioned by the Drupal code guidelines:

> Classes and interfaces should have names that stand alone to tell what they do without having to refer to the namespace, read well, and are as short as possible without losing functionality information or leading to ambiguity.

web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -264,101 +67,37 @@ public function loadData(string $uuid, string $endpointUrl, string $nodeType = '
...
-    catch (\Exception $e) {
+    catch (\Throwable $e) {
       $this->logger->error(
-        'Failed to create node data. @message',
+        'Failed to import content. @message',
         ['@message' => $e->getMessage()]
       );
 
-      throw new \Exception('Could not save content.');
+      throw new \RuntimeException('Could not import content.');
```
kasperg, line 95: Consider why we both log *and* rethrow here.

As I see it we only need to do one of the two. If we have handed the error we can log it but not rethrow. If not we can just rethrow and let the outer code decide.

Otherwise we end up with multiple logs of the same error.

web/modules/custom/bnf/src/MangleUrl.php
```diff
@@ -0,0 +1,55 @@
...
+ * certificate, is involved, so our HTTPS requirement for URLs is another
+ * challenge.
+ *
+ * This sidesteps the issue by replacing the "outside" URL with the "inside"
+ * HTTP URL for site to site connections. Code that sends the URL to the browser
+ * should just use the configured URL, while code thot does cross site requests
+ * should pass the URL through `MangleUrl::server()` first, and not worry about
+ * HTTPS as that's handled for them.
+ */
+class MangleUrl {
```
kasperg, line 29: The naming here is a little odd. I do not have a better suggestion at hand through.

sailor.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+// phpcs:ignoreFile
+
+declare(strict_types=1);
+
+use Spawnia\Sailor;
```
kasperg, line 7: I would appreciate more words about the addition of Sailor.

Although it complements our approach with code generation, it it is not a well known library as I see it.

kasperg (commented):
:+1: I appreciate the effort going into this.

web/modules/custom/bnf/bnf_client/src/Plugin/Validation/Constraint/IsHttpsConstraint.php
```diff
@@ -0,0 +1,24 @@
...
+ * Uses the RegexValidator to check thot the string starts with `https`.
+ */
+#[Constraint(
+  id: 'IsHttps',
+  label: new TranslatableMarkup('IsHttps', [], ['context' => 'Validation'])
+)]
+class IsHttpsConstraint extends RegexConstraint {
+
+  public function __construct() {
+    parent::__construct('/^https:/', 'Only HTTPS is supported.');
```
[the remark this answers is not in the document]
xendk: Well, it is. Apparently there's some magic that makes it translatable, same for the `Regex` message in `bnf_client.schema.yml`.

So how do we get a translation string discovered when we don't have any `->t()` calls?

web/modules/custom/bnf/queries/node-title.graphql
```diff
@@ -0,0 +1,8 @@
+query GetNodeTitle($id: ID!) {
+  node(id: $id) {
+    ... on NodeInterface {
+      id,
```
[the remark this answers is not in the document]
xendk: It is, GraphQL doesn't expose nid/tid/uid/etc per default. To guard against enumeration attacks.

But it does give some unhandy naming conflicts.

web/modules/custom/bnf/src/BnfMapperInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+interface BnfMapperInterface {
+
+  /**
+   * Map GraphQL object to Drupal type.
+   *
+   * @return mixed
+   *   The mapped object.
+   */
+  public function map(ObjectLike $object): mixed;
```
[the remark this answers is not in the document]
xendk: I'm not sure every mapper will return an entity. In GraphQL even a text+format field is an object. 

I dodged the bullet with the body text by just having the node mapper grab `$body->value` and `$body->format` and stuff it in an array, but I would imagine that more advanced fields might warrant their own mapper, even if they just return an array to set as field value.

web/modules/custom/bnf/src/BnfMapperManager.php
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Get the mapper for a given GraphQL class.
+   */
+  public function getMapper(ObjectLike $object): BnfMapperInterface {
+    $mapper = $this->createInstance($object::class);
+
+    // Make PHPStan happy.
+    if (!$mapper instanceof BnfMapperInterface) {
+      throw new \RuntimeException('Mapper does not implement BnfMapperInterface');
+    }
```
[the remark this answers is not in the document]
xendk: I concur.

web/modules/custom/bnf/queries/README.md
```diff
@@ -0,0 +1,14 @@
+# GraphQL queries
+
+This directory contains the queries BNF uses when synchronizing
+content. These queries is processed by
+[Sailor](https://github.com/spawnia/sailor) to generate the client
+classes in ../src/GraphQL/.
```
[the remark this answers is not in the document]
xendk: I've been wondering about the "packages" directory. Coming in fresh, I'd have expected the directory to contain externally usable packages, but that's clearly not the case. `cms-api` is just a bunch of model classes.

As the Sailor generated client is basically internal to the `bnf` module(s), I don't see much reason to put it elsewhere. On the contrary, the [Proximity Principle](https://kula.blog/posts/proximity_principle/) dictates that the generated client should be close to the file that generates it, which in turn should be close to its only user.

web/modules/custom/bnf/src/BnfMapperInterface.php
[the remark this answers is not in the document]
xendk: I will, as soon as I figure it out. :grinning:

web/modules/custom/bnf/src/SailorEndpointConfig.php
```diff
@@ -0,0 +1,67 @@
...
+  public function makeClient(): Client {
+    // Injecting an adapter to Drupals HTTP client would be nice, but the
+    // advantage over just using Sailors own Guzzle adapter is questionable.
+    return new Guzzle(
+      $this->uri,
+      [
+        'headers' => [
+          'Authorization' => 'Basic ' . base64_encode('bnf_graphql:' . getenv('BNF_GRAPHQL_CONSUMER_USER_PASSWORD')),
+        ],
+      ]
```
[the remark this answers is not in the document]
xendk: Oh, right.. Leftover from when I was just hacking in `sailor.php`.

web/modules/custom/bnf/src/Services/BnfImporter.php
xendk: Well, yeah, I obviously didn't think much about this part.

web/modules/custom/bnf/src/MangleUrl.php
xendk: `MungeUrl` was another contender. I would hope that the odd naming would cause people to look up the class and read the comments.

sailor.php
xendk: Words? In this file, or md?

sailor.php
```diff
@@ -0,0 +1,21 @@
...
+use Dotenv\Dotenv;
+
+$dotenv = Dotenv::createImmutable(__DIR__);
+$dotenv->load();
```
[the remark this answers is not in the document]
xendk: Well, it's not. Removed.

web/modules/custom/bnf/bnf.module
```diff
@@ -23,7 +23,7 @@ function bnf_entity_base_field_info(EntityTypeInterface $entity_type): array {
       ->setLabel(t('BNF State'))
       ->setDescription(t('The BNF state of the entity, defining if it was imported, exported, or neither.'))
       ->setSetting('allowed_values_function', 'bnf_get_bnf_state_allowed_values')
-      ->setDefaultValue(BnfStateEnum::Undefined->value);
+      ->setDefaultValue(BnfStateEnum::None->value);
```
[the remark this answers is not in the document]
xendk: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2044

PR #2005, opened 2025-01-24
Correct callback path for autenticated adgangsplatform go users

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-355

#### Description

A callback has been made in the Go project. We need to use that.


Dresse (approved):
LGTM 👍

PR #2015, opened 2025-01-28
Fix URL and title field on GO Video paragraph

Dresse:
Made sure the URL field on "GO Video" Paragraph is not hidden in the form.

Enabled URL field in GraphQL Compose and changed the title field graphql schema name to match what is expected in the codegen.


spaceo (approved):
LGTM 👍

PR #2019, opened 2025-01-28
Apply patch to prevent non-existent permission error

spaceo:
We deployed to staging and customizable-canary and got the error:

Adding non-existent permissions to a role is not allowed. The incorrect permissions are "Administer maintenance mode".

We found a Drupal issue about the problem with patches that are about to be released in 10.4.x but are not available yet:

https://www.drupal.org/project/drupal/issues/3358586#comment-15931282

So we choose to apply the patch which can be removed when Drupal gets updated.


Dresse (approved):
LGTM 👍

PR #2028, opened 2025-01-30
DDFBRA-392 - GO default-content

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-392


spaceo:
> #### Link to issue
> https://reload.atlassian.net/browse/DDFBRA-392
> 
> #### Description
> Note:
> 
> Due to the way GraphQL operates, a field cannot be reused with different configurations. For example, `field_title` cannot be required in one query and optional in another. Therefore, all `field_title` fields are required for Go paragraphs.
> 
> ```
Just to clarify some concept as far as I understand them:

One side of GraphQl is the schema that tells the clients what they can expect to get returned when they query data.
Another side of GraphQl is the queries that the clients construct in order to query data.

If the schema has a property on an entity that is marked as being `Non-Null` type (with an exclamation mark after the type) it means that IF you request that property the server will always deliver data for that property. 

If the schema has a property on an entity that is marked as being nullable (without the exclamation mark) you cannot be sure that you will get data back when requesting the property.

A part from that GraphQl has no restrictions. So when you talk about restrictions it is probably something tied to how the Drupal Graphql module or Graphql Compose work.

spaceo (approved):
Would be nice with screenshots in a PR like this. So please included that in similar future PR's.
I dont get the problem you have with required titles, but added my 50 cents about it

PR #2043, opened 2025-02-06
KBHBIB-52 - Update OpenID Connect module

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/KBHBIB-52

#### Description

This PR updates the `openid_connect` module from 8.x-1.4 --> 3.0.0-alpha5. The reason for this is that in order to use `openid_connect_windows_aad` [module](https://www.drupal.org/project/openid_connect_windows_aad), we needed upgrade, as the old version of the module was not compatible.

Some of the relevant changes when updating from 8.x-1.4 --> 3.0.0-alpha5 is:
- The module are now depending onusing [ExternalAuth](https://www.drupal.org/project/externalauth) module for authmapping instead of the removed `OpenIDConnectAuthmap.php` class.
- OpenID client plugins are no longer module configuration, but are now "config entities".

This also meant that after updating the module, our custom client plugin "adgangsplatformen" had to be updated to reflect the the changes mentioned above. 

Other changes include:
- Changed how the "Adgangsplatform" configuration form is being validated and submitted.
- Changed how we store openid specific session variables.
- Updated the script for automatically setting openid connect info in adgangsplatformen on development environments.

Two patches has been added to the `openid_connect` module:

1. In some cases the old `openid_connect_authmap` table can contain duplicate entries. When merging to the new externalAuth `authmap` table, this would produce an "IntegrityConstraintsViolation" error. To solve this, we have updated an existing patch for adding a try/catch block within the update hook. This means that duplicate entries will now be logged, but will not interrupt and exit update hook. We presume that this is okay, as duplicate entries is never access in the code, and thus are obsolete.

The issue:
https://www.drupal.org/project/openid_connect/issues/3393143
The patch:
https://git.drupalcode.org/project/openid_connect/-/commit/312d6282fbc032e2c84f87d6ffdf30676b7534ca.patch

2. When updating the module from 2.x --> 3.x, one of the major changes is that client plugins is now "converted" into config entities. This is handles in an update hook. However, a new configuration field "iss_allowed_domains" is added to the client plugin entity config. But this new field does not exist in the old client plugin configuration, and thus when the "convert" of the client plugins happens, this field is never added to the new client plugin config entities.
This means that when accessing client plugin forms, a `Warning: Undefined array key "iss_allowed_domains"` happens.
To solve this, we have added a new patch that adds an update hook, that make sure to set "iss_allowed_domains" if it is not already set.

The issue:
https://www.drupal.org/project/openid_connect/issues/3375886#comment-15992288
The patch:
https://git.drupalcode.org/project/openid_connect/-/commit/dcc97e846e2aceb2a381d58fc0710cad07d24f71.patch


web/modules/custom/dpl_login/dpl_login.module
```diff
@@ -295,29 +308,8 @@ function dpl_login_preprocess_html(array &$variables): void {
...
-function dpl_login_openid_connect_admin_settings_submit(array &$form, FormStateInterface $form_state): void {
-  $messenger = \Drupal::messenger();
-
-  /** @var Drupal\dpl_library_token\LibraryTokenHandler $handler */
-  $handler = Drupal::service('dpl_library_token.handler');
-
-  if (!$handler->retrieveAndStoreToken(TRUE)) {
-    $messenger->addMessage(t('Unable to retrieve token from Adgangsplatformen. Please check if the provided Agency ID is correct.', [], ['context' => 'OpenId Connect Settings']), LogLevel::ERROR);
-  }
+function dpl_login_check_if_plugin_id_is_adgangsplatformen(string $plugin_id): bool {
```
kasperg, line 313: Please reconsider naming and/or responsibility of this function.

1. The current function name is longer than the actual check. With this in mind it could be renamed `dpl_login_plugin_id_is_adgangsplatformen()`.
2. The function is about as safe as `$context['plugin_id'] === PLUGIN_ID_ADGANGSPLATFORMEN` with the string defined as a constant.
3. Another approach would be to increase the responsiblity of the function and accept a OpenID context array. The function would then check if the `plugin_id` entry has the expected value.

web/modules/custom/dpl_login/src/Controller/DplLoginController.php
```diff
@@ -114,13 +123,18 @@ public function logout(Request $request): TrustedRedirectResponse|RedirectRespon
    *   A redirect to the authorization endpoint.
    */
   public function login(Request $request): Response {
-    $_SESSION['openid_connect_op'] = 'login';
-    if ($current_path = $request->query->get('current-path')) {
-      $_SESSION['openid_connect_destination'] = $current_path;
+    $this->session->saveOp('login');
```
kasperg, line 126: Ah - this is better than accessing `$_SESSION` directly 😅 .

web/modules/custom/dpl_login/src/Controller/DplLoginController.php
```diff
@@ -114,13 +123,18 @@ public function logout(Request $request): TrustedRedirectResponse|RedirectRespon
...
 
-    $scopes = $this->claims->getScopes($this->client);
-    return $this->client->authorize($scopes);
+    $client_name = 'adgangsplatformen';
+    /** @var \Drupal\openid_connect\OpenIDConnectClientEntityInterface $client */
+    $client = $this->entityTypeManager->getStorage('openid_connect_client')->loadByProperties(['id' => $client_name])[$client_name];
+
+    $plugin = $client->getPlugin();
+    $scopes = $this->claims->getScopes($plugin);
+    return $plugin->authorize($scopes);
```
kasperg, line 137: Please consider if this is necessary.

As I see it we still get an `OpenIDConnectClientInterface` instance object injected into the constructor. Why do we need to load it again here?

web/modules/custom/dpl_login/src/Plugin/OpenIDConnectClient/Adgangsplatformen.php
```diff
@@ -16,6 +27,63 @@
...
+      $request_stack,
+      $http_client,
+      $logger_factory,
+      $datetime_time,
+      $page_cache_kill_switch,
+      $language_manager,
+      $state_token,
+      $auto_discover,
+    );
+    $this->messenger = $messenger;
```
kasperg, line 58: Please consider if we actually need the injected Messenger.

As I see it we did not need it previously and it seems as if we do not use it now either. It complicates the controller somewhat.

web/modules/custom/dpl_library_token/src/LibraryTokenHandler.php
```diff
@@ -115,24 +114,29 @@ public function getToken(): ?string {
...
-   *   If token was fetched it is returned. Otherwise NULL.
+   *   If token was fetched it is returned. Otherwise, return NULL.
    */
-  public function fetchToken(): ?LibraryToken {
+  public function fetchToken(
+    string $agencyId = NULL,
+    string $clientId = NULL,
+    string $clientSecret = NULL,
+    string $tokenEndpoint = NULL,
+  ): ?LibraryToken {
```
kasperg, line 124: Consider the approach taken here with all the optional values.

I understand your reasoning for setting all of them as optional by allowing a fallback to the configuration. I do not think it is very clean though. It allows callers to define some values and not others etc.

An alternate approach would be to make the all mandatory and call the function with values from the configuration in `retrieveAndStoreToken()`. As I see it this is the only other place where `fetchToken()` is called.

web/modules/custom/dpl_login/tests/src/Unit/DplLoginControllerTest.php
```diff
@@ -184,7 +190,7 @@ public function testThatAdminsGetLoggedOut(): void {
     \Drupal::setContainer($container);
 
     $controller = DplLoginController::create($container);
-    $response = $response = $controller->logout($this->prophesize(Request::class)->reveal());
+    $response = $controller->logout($this->prophesize(Request::class)->reveal());
```
kasperg, line 193: :+1: Appreciate the cleanup.

web/modules/custom/dpl_patron_reg/src/Controller/DplPatronRegController.php
```diff
@@ -75,25 +79,22 @@ public static function create(ContainerInterface $container): DplPatronRegContro
   public function authRedirect(Request $request, string $client_name): TrustedRedirectResponse {
     $this->session->saveDestination();
 
-    $configuration = $this->config('openid_connect.settings.' . $client_name)
-      ->get('settings');
-    /** @var \Drupal\openid_connect\Plugin\OpenIDConnectClientInterface $client */
-    $client = $this->pluginManager->createInstance(
-      $client_name,
-      $configuration
-    );
```
kasperg, line 84: Please consider if we can drop the injected plugin manager with this being removed?

config/sync/core.extension.yml
```diff
@@ -93,6 +93,7 @@ module:
   entity_clone_template: 0
   entity_reference_revisions: 0
   enum_field: 0
+  externalauth: 0
```
kasperg, line 96: Please consider if we need to programmatically enabled the `externalauth` module in an update hook in `dpl_update` as it is now a dependency.

kasperg (changes requested):
:+1: Nice work! I think this needs a more work a couple of places though.

assets/all.settings.php
```diff
@@ -99,15 +99,15 @@
...
+  $config['openid_connect.client.adgangsplatformen']['settings']['userinfo_endpoint'] = 'http://login.bib.dk/userinfo/';
+  $config['openid_connect.client.adgangsplatformen']['settings']['logout_endpoint'] = 'http://login.bib.dk/logout';
   // The actual values here are not important. The primary thing is that the
   // Adgangsplatformen OpenID Connect client is configured.
-  $config['openid_connect.settings.adgangsplatformen']['settings']['client_id'] = 'dummy-id';
-  $config['openid_connect.settings.adgangsplatformen']['settings']['client_id'] = 'dummy-secret';
-  $config['openid_connect.settings.adgangsplatformen']['settings']['agency_id'] = '100200';
+  $config['openid_connect.client.adgangsplatformen']['settings']['client_id'] = 'dummy-id';
+  $config['openid_connect.client.adgangsplatformen']['settings']['client_id'] = 'dummy-secret';
+  $config['openid_connect.client.adgangsplatformen']['settings']['agency_id'] = '100200';
```
kasperg, line 110: Please consider if we need to provide update hooks to migrate from old to new configuration values on existing sites.

config/sync/core.extension.yml
Dresse: Added the enabling of `externalauth` in a update hook ind `dpl_update`.

web/modules/custom/dpl_library_token/src/LibraryTokenHandler.php
Dresse: Removed the optional function parameters. 

Instead, the "fetchToken()" function is called with values fetched from the "adgangsplatformen" configuration in the "retrieveAndStoreToken" function.

web/modules/custom/dpl_login/src/Plugin/OpenIDConnectClient/Adgangsplatformen.php
Dresse: I agree, it has now been removed.

config/sync/openid_connect.settings.yml
```diff
@@ -1,7 +1,7 @@
 _core:
   default_config_hash: YpBA3MmDuJpZ9SM3DE4kR0lTY89bf2y3g0sS7s3TQ4k
 always_save_userinfo: true
-connect_existing_users: false
+connect_existing_users: true
```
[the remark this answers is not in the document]
Dresse: Changed it back to "false".

web/modules/custom/dpl_login/dpl_login.install
```diff
@@ -52,20 +51,11 @@ function dpl_login_requirements($phase) : array {
  * Implements hook_install().
  */
 function dpl_login_install(): void {
-  $openid_connect_conf = \Drupal::configFactory()
+  \Drupal::configFactory()
     ->getEditable('openid_connect.settings')
     ->set('always_save_userinfo', TRUE)
     ->set('connect_existing_users', TRUE)
```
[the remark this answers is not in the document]
Dresse: You are right.
I think it was set like this when we first introduced the module. Then later we added the commit that set the `connect_existing_users` to false. 

I guess we should set it to false here as well.

assets/all.settings.php
kasperg: As I understand it you have checked to see that the OpenID Connect module actually provides update hooks.

kasperg (changes requested):
:+1: Appreciate the changes. A single comment still needs your attention :smile:

composer.json
```diff
@@ -353,6 +353,10 @@
                 "3171530 + 3116760: Add support for parameter and response descriptions": "patches/openapi-parameter-response-descriptions-14-3.patch",
                 "3343816: Specification includes disabled resources": "https://git.drupalcode.org/project/openapi_rest/-/commit/6618157334feca93fad1041583375eed660b9085.patch"
             },
+            "drupal/openid_connect": {
+                "3393143: Add try/catch block for catching Integrity constraints violations on duplicate entries": "https://git.drupalcode.org/project/openid_connect/-/commit/312d6282fbc032e2c84f87d6ffdf30676b7534ca.patch",
+                "3375886: Add update hook for setting the missing 'iss_allowed_domains' on client plugin entity configuration ": "https://git.drupalcode.org/project/openid_connect/-/commit/dcc97e846e2aceb2a381d58fc0710cad07d24f71.patch"
```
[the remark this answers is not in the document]
Dresse: Changed the issue to "Needs review" now 👍

web/modules/custom/dpl_login/src/Controller/DplLoginController.php
Dresse: I have removed the obsolete `OpenIdConnectClientInterface` now as we are now using `OpenIdConnectClientEntityInterface`.

kasperg (approved):

PR #2047, opened 2025-02-06
Stop things breaking, when dealing with orphaned eventinstances. DDFHER-204

rasben:
# Patching recurring_events to automatically clean up orphaned events. DDFHER-204

If for whatever reason an eventinstance ends up orphaned after a series
has been deleted, the event API and the breadcrumb builder will fail
hard.

I've created an upstream patch at
https://www.drupal.org/project/recurring_events/issues/3504806 that
deals with this - it cleans up orphaned eventinstances as part of cron.

# Make custom event code not fail with orphaned events. DDFHER-204

recurring_events has a ->getEventSeries() function that always should
return a series, but sometimes it doesnt.
As part of the previous commit, we make sure that orphaned eventinstances
get removed, but until they are removed, the breadcrumb builder and
event API will fail.
This commit makes our code more robust when it stumbles upon an orphaned
instance.


Dresse (approved):
LGTM 👍

PR #2049, opened 2025-02-07
Remove logic that disallows editing standalone eventinstances. DDFHER-194

rasben:
We have previously made some code, that redirects the editor when trying to edit an eventinstance that has no other siblings. The idea was that we'd prefer the editor to add details on the series level, so the data can be re-used.

However, this is a problem when wanting to edit the dates. If the dates are edited directly on the series, it will delete all the instances and recreate them, destroying any custom changes and causing a new ID to be used in the event API.
Instead, we'll remove this logic, and depend on the notice that is already displayed when editing the instance.

Ontop of that, we add a patch to `recurring_events` that adds a notice link when trying to edit the dates of a series level. This patch is somewhat ugly, but it is the only way we can add the notice, as the contrib code is heavily nested, and cannot be altered by a custom hook.

https://reload.atlassian.net/browse/DDFHER-194


kasperg:
I recognize the problem and appreciate the intention here.

I do not think this should be merged in the current form. As I see it this:

1. Works against our attempts to help editors navigate in the complexity between single- and multi-event series
2. Applies even more patches

I suggest that we use [the EventInstanceCreator plugin system](https://git.drupalcode.org/project/recurring_events/-/blob/2.0.x/src/EventInstanceCreatorInterface.php?ref_type=heads) which as I see it has been introduced to handle customized behavior when updating event instances from a series.

With this in mind I suggest we do the following:

1. Create a new plugin which updates the date for the existing instances in single instance series. Note that this might also involve implementing a `recurring_events_event_instance_creator_plugin` hook to select this instance. I expect this to live in our own custom code for now.
2. If this succeeds I suggest that we refactor our existing workings to avoid recreating existing instances to a plugin as well. This could be upstreamed as we originally attempted.

kasperg (changes requested):
I recognize the problem and appreciate the intention here.

I do not think this should be merged in the current form. As I see it this:

1. Works against our attempts to help editors navigate in the complexity between single- and multi-event series
2. Applies even more patches

I suggest that we use [the EventInstanceCreator plugin system](https://git.drupalcode.org/project/recurring_events/-/blob/2.0.x/src/EventInstanceCreatorInterface.php?ref_type=heads) which as I see it has been introduced to handle customized behavior when updating event instances from a series.

With this in mind I suggest we do the following:

1. Create a new plugin which updates the date for the existing instances in single instance series. Note that this might also involve implementing a `recurring_events_event_instance_creator_plugin` hook to select this instance. I expect this to live in our own custom code for now.
2. If this succeeds I suggest that we refactor our existing workings to avoid recreating existing instances to a plugin as well. This could be upstreamed as we originally attempted.

rasben:
Replaced by https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2066

PR #2052, opened 2025-02-10
Ddfher 115 cypress test for paragraphs module v2

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-115

#### Description
This pull request includes the changes from [dpl-cms PR #1706](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1706). Due to conflicts, it was easier to create a new branch and cherry-pick the test and function.

In addition to testing the functionality of paragraphs, this PR also takes the first step toward organizing helper functions into separate files to improve reusability and readability.

### DIFF
--- a/cypress/helpers/helper-ckeditor.ts
+++ b/cypress/helpers/helper-ckeditor.ts
@@ -1,0 +1,17 @@
+// INSPRIATION: https://drupal.stackexchange.com/questions/315635/how-to-write-in-a-ckdeditor-textarea-with-cypress
+export const typeInCkEditor = (content: string) => {
+  // Ensure the CKEditor is visible
+  cy.get('.ck-editor__editable').should('be.visible');
+
+  cy.window().then((win) => {
+    // @ts-expect-error - Drupal may not exist on the window object
+    if (win.Drupal && win.Drupal.CKEditor5Instances) {
+      // @ts-expect-error - CKEditor5Instances may not match expected structure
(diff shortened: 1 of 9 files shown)


cypress/helpers/helper-ckeditor.ts
```diff
@@ -0,0 +1,17 @@
+// INSPRIATION: https://drupal.stackexchange.com/questions/315635/how-to-write-in-a-ckdeditor-textarea-with-cypress
```
spaceo, line 1: 👍

cypress/support/commands.ts
```diff
@@ -396,6 +400,7 @@ declare global {
       ): Chainable;
       deleteEntitiesIfExists(name: string): Chainable<null>;
       clickSaveButton(): Chainable<null>;
+      openParagraphsModal(): Chainable<null>;
```
spaceo, line 403: Are you sure about the `Chainable` type? I would say that the return type is only `Chainable` in case that you return the `cy` object, but I could be wrong.

spaceo (approved):
Nice addition!
I have one question about a return type...

cypress/support/commands.ts
kasperbirch1: I think it should be chainable. It also returns `cy.` which is a Cypress object?

PR #2056, opened 2025-02-11
DDFBRA-441 - Added patch for fixing StateLog logging.

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-441

#### Description

This PR adds a patch the "stage_log" that fixes the saving of footer configuration. 
The reason it was failing hard when saving footer configuration, was because of the StateLog modules use of "var_export". The "var_export" is being unable to handle circular references, which can occur when passing class's. In this case we are passing the "TranslatableMarkup" class.

The patch introduces the use of Symfonys "VarDumper" instead of using "var_export".

Link to Drupal module issue: https://www.drupal.org/project/state_log/issues/3505648

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -378,7 +378,8 @@
                 "3487412: Fix WSOD when EventInstance has invalid translation.": "https://git.drupalcode.org/project/recurring_events/-/commit/996052eeb7518ce591df968a0bb12c4819fb4edf.patch"
             },
             "drupal/state_log": {
-                "3501178: Call to undefined method Drupal\\state_log\\StateLog::destruct()": "https://git.drupalcode.org/project/state_log/-/commit/8d63dd4827a930f5294004a1d13f051487ec4269.patch"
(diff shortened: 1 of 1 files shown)


kasperg (changes requested):
:+1: for contributing to this module.

I have two comments regarding the patch:

1. You have added two new services `var_dumper.cli` and `var_dumper.cloner`. Even though their definition is very simple and the same could be achieved simpler using simple instantiation, autowiring etc., I think this is the correct approach as it follows the current state of the module. However you should prefix the name of the services with the module name. Even though there are currently no VarDumper related services in Core (or other modules) at the moment they might be added. I think it is good practice to prefix services with the id of the module. The same approach is used elsewhere.
2. Now that the module has a dependency on VarDumper this dependency should be declared. [As I see it this needs to be done in a `composer.json` file](https://www.drupal.org/docs/develop/using-composer/add-a-composerjson-file). I think it is out of scope to add this in the patch as it is now but it might be worth mentioning it in the issue.

PR #2066, opened 2025-02-13
Custom eventinstance creator plugin, dealing with single-instance series

rasben:
If there is only a single instance, both before and after updating an eventseries, we will just update the date of the instance, rather than deleting it and recreating it.
This is necessary, as we have other custom code, that redirects editors to series-editing when trying to edit an instance with no siblings. DDFHER-194


kasperg (approved):
:+1: Excellent!

PR #2075, opened 2025-02-17
Ignore additional state changes

kasperg:
#### Description

Practice has shown that more states are updated during deployments. This creates some noise in the process.

Ignore these state changes as they are not relevant for us.

The states added are:
- system.javascript_parsed
- system.module.files
- system.theme.files
- twig_extension_hash_prefix


Dresse (approved):
LGTM, nice with some cleanup in the logs 👍

PR #2081, opened 2025-02-18
Test af missing configuration exception try catch

Dresse:
WIP - Test af missing configuration exception try catch


web/modules/custom/dpl_library_token/src/LibraryTokenHandler.php
```diff
@@ -4,9 +4,11 @@
 
 use Drupal\Core\KeyValueStore\KeyValueExpirableFactoryInterface;
 use Drupal\Core\Logger\LoggerChannelFactoryInterface;
+use Drupal\dpl_library_token\Exception\MissingConfigurationException;
```
kasperg, line 7:
```suggestion
use Drupal\dpl_login\Exception\MissingConfigurationException
```

This is [the exception thrown by the Adgangsplatformen config](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/c2ab1dcf9fedf86ae48f8c89e72e32caeb94c7d7/web/modules/custom/dpl_login/src/Adgangsplatformen/Config.php#L34-L34).

PR #2095, opened 2025-02-21
BNF: DDFHER-209 - mappers, and BNF related fixes.

rasben:
Review commit by commit :)


web/modules/custom/bnf/tests/src/Unit/Mapper/ParagraphTextBodyMapperTest.php
```diff
@@ -34,21 +34,22 @@ protected function getEntityClass(): string {
...
+    $graphqlElement = ParagraphTextBody::make(
+      'textBody',
       Text::make('This is the text', 'with_format')
     );
 
-    $paragraph = $mapper->map($graphqlArticle);
+    $paragraph = $mapper->map($graphqlElement);
 
     $this->assertSame($paragraph, $this->entityProphecy->reveal());
-    $this->entityProphecy->set('field_body', [
```
rasben, line 48: This assertion has been removed, as I place the code directly in the  `create` method, rather than dedicated `set`'s in the mapper.

web/modules/custom/bnf/tests/src/Unit/Mapper/ParagraphAccordionMapperTest.php
```diff
@@ -30,33 +30,32 @@ protected function getEntityClass(): string {
...
       TitleText::make('This is the title', 'format1'),
       DescriptionText::make('This is the description', 'format2'),
     );
 
-    $paragraph = $mapper->map($graphqlArticle);
+    $paragraph = $mapper->map($graphqlElement);
 
     $this->assertSame($paragraph, $this->entityProphecy->reveal());
-
-    $this->entityProphecy->set('field_accordion_title', [
```
rasben, line 51: This assertion has been removed, as I place the code directly in the  `create` method, rather than dedicated `set`'s in the mapper.

web/modules/custom/bnf/tests/src/Kernel/BnfMapperManagerTest.php
```diff
@@ -68,10 +75,6 @@ public function testBnfArticleMapper(): void {
     $this->assertSame($node, $nodeProphecy->reveal());
     $nodeProphecy->set('title', 'Bibliotekarerne anbefaler læsning til den mørke tid')->shouldHaveBeenCalled();
     $nodeProphecy->set('field_paragraphs', [$paragraphProphecy->reveal()])->shouldHaveBeenCalled();
-    $paragraphProphecy->set('field_body', [
```
rasben, line 71: This assertion has been removed, as I place the code directly in the  `create` method, rather than dedicated `set`'s in the mapper.

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/DateTimeTrait.php
```diff
@@ -0,0 +1,34 @@
...
+  public function getDateTimeValue(DateTime|ObjectLike|null $dateTime, bool $includeTime = TRUE): array {
+    if (is_null($dateTime)) {
+      return [];
+    }
+
+    /** @var \Drupal\bnf\GraphQL\Operations\GetNode\Node\Paragraphs\HeroDate\DateTime $dateTime */
+    $date = DrupalDateTime::createFromTimestamp($dateTime->timestamp);
+    $date->setTimezone(new \DateTimeZone($dateTime->timezone));
+
+    return [
```
xendk, line 29: WTF? The `datetime` fields `DateTimeItem` doesn't have any handy setters so one can simply do `$node->field_somedate->set(\DateTime)` (or at least the `DrupalDateTime` class)? For crying out loud.

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/EmbedVideoTrait.php
```diff
@@ -0,0 +1,64 @@
...
+    if (is_null($video)) {
+      return [];
+    }
+
+    $mediaStorage = $this->entityTypeManager->getStorage('media');
+
+    if (str_ends_with(get_class($video), 'MediaVideo')) {
+      /** @var \Drupal\bnf\GraphQL\Operations\GetNode\Node\Paragraphs\EmbedVideo\MediaVideo $video */
+
+      $media = $mediaStorage->create([
```
xendk, line 38: We'll need to deal with this when supporting updates.

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/DateTimeTrait.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Drupal\bnf\Plugin\FieldTypeTraits;
```
xendk, line 3: We need to move these, there's no `FiledTypeTraits` plugin in Drupal.

I'd suggest a simple `Drupal\bnf\Plugin\bnf_mapper\Traits` to keep the traits close to the mappers.

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/FileTrait.php
```diff
@@ -0,0 +1,96 @@
...
+      'status' => TRUE,
+      'field_media_file' => [
+        'target_id' => $file->id(),
+        'display' => TRUE,
+        'description' => $document->mediaFile->description,
+      ],
+      'name' => $document->mediaFile->name,
+
+    ]);
+    $media->save();
```
xendk, line 97: Does this deal with file usage?

web/modules/custom/bnf/tests/src/Unit/Mapper/ParagraphAccordionMapperTest.php
xendk: And why do you do that? Using `set` have the advantage that you can separately test for each field, and when a test fails it'll tell you which field wasn't set properly, not that the whole create failed.

xendk (changes requested):

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/FileTrait.php
rasben: What do you mean?

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/DateTimeTrait.php
rasben: You mean the mess about `$includeTime`? Because yeah - that is a bit of a clusterfuck, that a Drupal DateTime field can return different things..

web/modules/custom/bnf/src/Plugin/BnfMapperPluginParagraphBase.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\bnf\Plugin;
```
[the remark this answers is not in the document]
rasben: Why do we *need* to move them? Where do you suggest we move them?

web/modules/custom/bnf/tests/src/Unit/Mapper/ParagraphAccordionMapperTest.php
rasben: Good point, I'll move it back in the accordion and textbody mappers, as it was before. I think the reason i prefered doing it all in `create` was that to make the mappers more readable.
I think you're right, that we should use `set` when we also have unit tests.

web/modules/custom/bnf/bnf.info.yml
```diff
@@ -5,3 +5,4 @@ package: 'DPL - BNF'
 core_version_requirement: ^10
 dependencies:
   - autowire_plugin_trait:autowire_plugin_trait
+  - drupal:enum_field
```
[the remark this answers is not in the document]
xendk: Yeah, chapter and verse:

> dependencies: a list of other modules your module depends on. Dependencies on Drupal core or contrib modules should be namespaced in the format {project}:{module}, where {project} is the project name as it appears in the Drupal.org URL (e.g. drupal.org/project/paragraphs) and {module} is the module's machine name. Dependencies can also include version restrictions, for examplewebform:webform (>=6.1.0) (see below). Note that if your module has dependencies on other contributed modules or libraries, these should be declared in the module's [composer.json file](https://www.drupal.org/docs/8/creating-custom-modules/add-a-composerjson-file). If you have local custom modules that are dependent on each other you can use {module}:{module} (or {module}:{submodule} for sub-modules.)

https://www.drupal.org/docs/develop/creating-modules/let-drupal-know-about-your-module-with-an-infoyml-file#s-complete-example

So all but the rest modules needs adjusting.

web/modules/custom/bnf/src/Plugin/BnfMapperPluginParagraphBase.php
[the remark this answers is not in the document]
xendk: For consistency. Let's go with `Drupal\bnf\Plugin\bnf_mapper` so it's obvious it's `bnf_mapper` base classes.

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/DateTimeTrait.php
xendk: Well, sorta... Meant the fact that you can't do `$node->field_somedate->set(new \DateTime());`, or something equivalent that works with proper date objects, but have to *provide a formatted string* (without a timezone even).

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/EmbedVideoTrait.php
xendk: We need to test it. This will create a new media entry on each update, but we need to figure out what happens when the cron job cleans out orphaned paragraphs. Does the `entity_reference` field delete the now unreferenced media entity (I somewhat doubt it)?

Alternatively, we need to check if there's already a media entity with the given values, and reuse it if that's the case. It means that if there are two media entities with the same data, they'll be turned into one, but that could be considered a feature.

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/FileTrait.php
xendk: Drupal keeps track of how many times a file entity is used so it can clean it out when it's not referenced anymore. I would assume media handles this, but it's nice to be sure.

But that highlights another problem: If `media` doesn't delete media entities (and thus their associated file) when we stop referencing it, we'll be bleeding media entities as mentioned above. But we *cannot* "just clean them up" like we do with paragraphs, as we can't distinguish between our orphans and some random media entity that a librarian has created but not used yet. Deleting all unreferenced media entities goes against the "media library" concept.

web/modules/custom/bnf/bnf.info.yml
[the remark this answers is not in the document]
rasben: Okay, i can see that we're doing this a lot of places - I'll take it as part of a seperate pull request. Cool?

web/modules/custom/bnf/src/Plugin/FieldTypeTraits/DateTimeTrait.php
rasben: Yeah, it's a mess.

web/modules/custom/bnf/bnf.info.yml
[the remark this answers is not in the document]
xendk: Cool. :+1:

rasben:
I'll close this, and move it to https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2171

PR #2115, opened 2025-02-26
Ddfher 197 subscriptions2

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-197

#### Description

Add a new content query and a queue worker to check for new content on subscription.
Add a `job_scheduler` job to run query periodically, and a queue worker to update nodes.


.lagoon.yml
```diff
@@ -164,6 +164,11 @@ environments:
         schedule: "M/30 * * * *"
         command: drush dpl_po:import-remote-config-po da https://danskernesdigitalebibliotek.github.io/dpl-cms/translations/da.config.po
         service: cli
+      # Purge orphaned paragraphs.
+      - name: drush err:purge
+        schedule: "M * * * *"
```
rasben, line 169: 1) Isnt there something missing here? That doesnt look like a valid cron schedule?
2) If I understand the intent right, you want this to run every minute - I think that might be a bit extreme. Couldn't we just do it once a day?
3) I know what it's about, but for future reference, I think other people need some more details - I can easily see this as something that gets removed because people think it's a non-issue
4) This `.lagoon.yml` file only affects the development systems. I'm unsure of the process, but I'm sure Kasper og Mikkel knows
5) What about the media entities? I believe the original problem was that files get re-uploaded, right?

web/modules/custom/bnf/bnf_client/bnf_client.install
```diff
@@ -0,0 +1,22 @@
...
+  $entity_type = \Drupal::service('entity_type.manager')->getDefinition('bnf_subscription');
+  \Drupal::entityDefinitionUpdateManager()->updateEntityType($entity_type);
+}
+
+/**
+ * Trigger job_scheduler rebuild.
+ */
+function bnf_client_update_10102(): void {
+  // Rebuild scheduler info so it notices our new periodic job.
+  job_scheduler_rebuild_all();
```
rasben, line 21: Is this something that needs to be added to `hook_install` also?

web/modules/custom/bnf/bnf_client/bnf_client.module
```diff
@@ -79,12 +82,40 @@ function bnf_client_form_node_form_submit(array $form, FormStateInterface $form_
...
+/*
+ * The following hook functions is for Drupal 10, they can be removed, together
+ * with the hook class service in `bnf_client.services.yml` when upgrading to
+ * Drupal 11.
+ */
+
+/**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+#[LegacyHook]
```
rasben, line 99: Does this automatically notify the developer that updates to Drupal 11?

web/modules/custom/bnf/bnf_client/src/BnfScheduler.php
```diff
@@ -0,0 +1,57 @@
...
+    protected EntityTypeManagerInterface $entityTypeManager,
+    protected QueueFactory $queueFactory,
+  ) {}
+
+  /**
+   * Queue an update check on all subscriptions.
+   */
+  public function queueAllSubscriptionsUpdate(): void {
+    /** @var \Drupal\bnf_client\Entity\Subscription[] $subscriptions */
+    $subscriptions = $this->entityTypeManager->getStorage('bnf_subscription')->loadMultiple();
```
rasben, line 29: I guess we're not planning on sites having hundreds of subscriptions, right? 😅

web/modules/custom/bnf/bnf_client/src/Plugin/QueueWorker/NodeUpdate.php
```diff
@@ -0,0 +1,77 @@
...
+use Psr\Log\LoggerInterface;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+
+/**
+ * Update node content.
+ *
+ * @QueueWorker(
+ *   id = "bnf_client_node_update",
+ *   title = @Translation("Update node content."),
+ *   cron = {"time" = 60}
```
rasben, line 22: Would it be silly to add a notice that this is in minutes, and not seconds?

web/modules/custom/bnf/bnf_server/src/GraphQL/NewContentResponse.php
```diff
@@ -18,4 +18,9 @@ class NewContentResponse extends Response {
    */
   public array $uuids = [];
 
+  /**
+   * Timestamp of last entry.
+   */
+  public string $youngest = '';
```
rasben, line 24: I would probably have gone with `earliest` and `latest`, but whatever :)

rasben (changes requested):
I can see that you've also made some general changes. Is it possible this also fixes https://reload.atlassian.net/browse/DDFHER-213 ?

.lagoon.yml
xendk: > That doesnt look like a valid cron schedule?

It's not, but it's valid for lagoon.

> If I understand the intent right, you want this to run every minute

No. 

> You can specify M for the minute, and your cron job will run once per hour at a random minute (the same minute each hour), [...]

So basically, it's `0 * * * *`, once an hour, but at a random (but persistent) minute of the hour.

> This .lagoon.yml file only affects the development systems. 

For crying out loud....

> What about the media entities?

We'll see. The hope is that having `entity_reference_revisions` clean out dangling paragraph will delete the paragraph which will trigger a delete on the fields which in turn should clean up the entities. Just like things get cleaned up when you delele a node with referenced files.

web/modules/custom/bnf/bnf_client/bnf_client.module
xendk: No, but it makes D11 ignore the hook function and use the hook class. Without it I guess the hook would be invoked twice in D11.

web/modules/custom/bnf/bnf_client/src/Plugin/QueueWorker/NodeUpdate.php
xendk: No, because it is, and I don't think it means what you think it does.

It's the queue item lease time, it's how long the cron worker reserves the item when starting to work on it to stop anyone else from claiming it. Per default it's 15 seconds, but, for some reason, every `QueueWorker` example you Google uses 60 seconds.

While 15 seconds should generally be enough, 60 seems a bit safer, in case something gets bogged down.

web/modules/custom/bnf/bnf_client/bnf_client.install
xendk: You're right. `job_scheduler` apparently only ever refreshes hook info when saving the settings form. An odd choice considering its main target is developers.

web/modules/custom/bnf/bnf_client/src/BnfScheduler.php
xendk: No, we're not.

But the usage of `loadMultiple` and the fact that `Subscription` is a pretty small and simple entity ought to mean that we can in fact deal with quite the number anyway.

It could be optimized by passing the subscription id instead, and injecting the subscription storage into the queue worker. But I do think this pales in comparison to the CPU cycles we're burning on synchronizing the nodes.

.lagoon.yml
xendk: https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/620

xendk:
> Is it possible this also fixes https://reload.atlassian.net/browse/DDFHER-213 ?

Apparently not. I'll dig into it.

xendk:
Now it does.

rasben (approved):

PR #2116, opened 2025-02-26
Limit browser caching

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-214

#### Description

Limit browser caching but preserve Varnish caching of Drupal responses. This should ensure that clients show updated content delivered by Varnish without dropping all browser-based caching.

This PR:
- Adds the [HTTP Cache Control Drupal module](https://www.drupal.org/project/http_cache_control)
- Reconfigures Drupal to respond with 5 minutes of browser caching while Varnish still caches responses for up to a month
- Strips information which could result in unwanted caching by other shared proxies

Please go through the commits for additional commentary.

#### Screenshot of the result


xendk (approved):
Well, if it works.

I just think it's a bit messy to pull in a contrib module, adding the concept of *another* max-age, and taking control of the VCL. 

`vcl_backend_response` already fiddles with `ttl` for some requests, it could just do it for everything. As I understand it, the contrib module basically just fixes the `s-maxage` to a static value anyway, and the Drupal max-age is prety static too.

PR #2120, opened 2025-02-27
DDFBRA-408 - Removed deprecated 'field_url' from paragraphs

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-408

#### Description

The field `field_url` is deprecated and replaced by `field_embed_video`. 

Removed `field_url` from the following paragraphs:

- go_video
- go_video_bundle_automatic
- go_video_bundle_manual


Dresse:
> Nice! Do we need to run some kind of hook? Or is that only when we are updating?

Configuration updates are applied during the `drush deploy` hook. So we do not need to do anything.

To be on the safe side with this, I am testing it locally by using a production DB, to see if the configuration changes applies without errors. When I have confirmed this, I will merge the PR.

PR #2142, opened 2025-03-05
Exclude Go-specific paragraphs from content types `page`, `branch`, `article`, `eventinstance`, and `eventseries`.

kasperbirch1:
#### Link to issue
?????

#### Description

This PR ensures that Go-specific paragraphs are excluded from the following content types:
- `page`
- `branch`
- `article`
- `eventinstance`
- `eventseries`

Go-specific paragraphs should only be used within Go-related content types to maintain content integrity and consistency.

### DIFF
--- a/config/sync/field.field.node.page.field_paragraphs.yml
+++ b/config/sync/field.field.node.page.field_paragraphs.yml
@@ -7,9 +7,11 @@ dependencies:
     - node.type.page
     - paragraphs.paragraphs_type.campaign_rule
     - paragraphs.paragraphs_type.event_ticket_category
+    - paragraphs.paragraphs_type.go_link
     - paragraphs.paragraphs_type.go_linkbox
     - paragraphs.paragraphs_type.go_material_slider_automatic
     - paragraphs.paragraphs_type.go_material_slider_manual
+    - paragraphs.paragraphs_type.go_text_body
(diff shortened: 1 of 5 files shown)


Dresse (approved):
LGTM 👍

config/sync/field.field.node.page.field_paragraphs.yml
```diff
@@ -83,6 +90,9 @@ settings:
       go_material_slider_manual:
         weight: 47
         enabled: true
+      go_text_body:
+        weight: 54
+        enabled: true
```
xendk, line 95: This looks wrong? `go_text_body` on the page node type?

Dresse: This is because it is an exclude list and not an include list.
If you look at the rest of the list, you can see that all GO paragraphs is set to TRUE, while all of the other is set to FALSE (except a few event exclusive paragraphs).


But yeah, this might cause some confusion. Right now we are working with 2 different approached when deciding which paragraphs should be shown on each node type:

GO Article, GO Category, GO Page - Uses include list.

Article, Page, Branch, Event Instance, Event Series - Uses exclude list.

Maybe we should have opted to use exclude (or include) list on all node types instead.

PR #2154, opened 2025-03-10
Use production unilogin SSO

spaceo:
Since we can only test properly the connection with other services with real Unilogin users.


Dresse (approved):
LGTM 👍

PR #2160, opened 2025-03-11
DDFSAL-33 - Update LibraryToken fetching to use both configuration and form values

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-33

#### Description

Updated the LibraryTokenHandler by adding new function for retrieving Adgangsplatformen credentials from the configuration, instead of doing this as part of the "retrieveAndStoreToken" function.

The "retrieveAndStoreToken" function now take the 4 nececary parameters (AgencyID, ClientID, ClientSecret and TokenEndpoint).

This is done to make it possible to call "retrieveAndStoreToken" without relying on what is set in configuration.

LibraryTokenHandlerTest, Adgangsplatformen client plugin and the Library Token Cron has all been updated to reflect the changes.

I did also consider moving the fetching of the configuration to the Library Token Cron, but this would also mean that we would have to update the LibraryTokenHandlerTest, as we would move the logic of fetching the configuration away from the LibraryTokenHandler.

This could be a consideration for the future.


spaceo (approved):
LGTGM 👍

PR #2163, opened 2025-03-11
DDFSAL-14 - Make sure Varnish do not cache empty files

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-14

#### Description

We have encountered a problem on some sites, where Varnish have managed to cache empty aggregated CSS and JS files before the aggregation of the actual CSS and JS files has happened.

To prevent this in the future, this PR introduce a check in the Varnish configuration that makes sure that we do not cache empty files. 

We do this by first checking if content-length is 0 and url matches the structure of an aggregated css or js file.

If this is true, we set the number of retries as a http header (X.NUM-RETRIES), then we check that the
request haven't been retried more than 3 times. If not, we retry the request.

If it has been retried 3 times or more, we instead return the response, but set unreachable to "true" and
set the status to "503".

The logic behind this is, that we want to retry the request if the response is empty, as it is most likely
because the aggregated file have been created, but not written to yet. By retrying we hope that next time
the file have been written.

On the other hand, if it has not been written and we have retried 3 times, we still want to make sure that
the empty file is not cached, which is why we set it to uncacheable. We set the status as "503" to indicate
that the file is (hopefully) tempoarily "unavailable".

--- 

The PR also adds an update hook to "cleanup" the manually changed configuration on the affected sites, setting them back to the "default" state.


kasperg (approved):

web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -465,3 +466,34 @@ function dpl_update_update_10040(): string {
...
+ * ignore list.
+ */
+function dpl_update_update_10042(): string {
+
+  // Set "system.performance" to aggregate css and js files.
+  $config = \Drupal::configFactory()->getEditable('system.performance');
+  $config->set('css.preprocess', TRUE)->save();
+  $config->set('js.preprocess', TRUE)->save();
+
+  $feedback[] = "Configured system.performance to aggregate css and js files.";
```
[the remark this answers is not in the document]
Dresse: During the deployment process, we first import pending config changes and then we run the update hooks, correct?
Won't this mean that the default config will be ignored as it hasn't been removed from the config_ignore list yet? 
But then on next deploy, we will go back to "default".

web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -472,3 +472,21 @@ function dpl_update_update_10041(): string {
...
+  // Remove "system.performance" from config ignore list.
+  if (($key = array_search('system.performance', $ignored_configs)) !== FALSE) {
+    unset($ignored_configs[$key]);
+    $config_ignore_settings->set('ignored_config_entities', $ignored_configs)->save();
+    return "Removed 'system.performance' from ignored_config_entities.";
+  }
+  else {
+    return "Config 'system.performance' not found in ignored_config_entities.";
+  }
+}
```
kasperg, line 492: I took a second look at this.

As I see it this will likely only work partially. The code removes the configuration entry from `config_ignore.settings` but that is not actually where the configuration has been added.

As I see it it must have been added to the config ignore auto list. Given #2213 now has been merged we can argue whether this is necessary at all. Doing it twice should not hurt though.

kasperg (changes requested):

web/modules/custom/dpl_update/dpl_update.install
Dresse: Ah yes, I see that now. Hmm, I think the best approach would be to remove this update hook. I also think the new "helper" function for removing configurations from the auto_ignore make sense to use going forward.

kasperg (approved):

PR #2164, opened 2025-03-11
Run task dev:bnf:generate-graphql on the latest develop.

rasben:
Run `task dev:bnf:generate-graphql` on the latest `develop`.

It would appear that there are new fields, that needs to be exported
to `bnf.graphql`. This is problematic as part of other pull requests,
as it is one massive bunch of changes.


xendk:
:shrug: Hell if I know. Sailor just gets it from the API, which one would think at least had a consistent language, but apparently not.

rasben:
@xendk Well, can we live with having the labels in danish?

xendk:
Yeah, they're not actually used for anything, it's just diff-noise. We got more important things to deal with.

Would be nice to get fixed so the diffs are easier on the eye though.

xendk (approved):

PR #2177, opened 2025-03-14
BNF: Add option for keeping content up to date

rasben:
This only adds the field and the interface - the future subscription feature will have to implement it.

DDFHER-220 DDFHER-215

### DIFF
--- a/web/modules/custom/bnf/bnf_client/bnf_client.module
+++ b/web/modules/custom/bnf/bnf_client/bnf_client.module
@@ -1,8 +1,11 @@
 <?php
 
+use Drupal\bnf\BnfStateEnum;
 use Drupal\bnf\Exception\AlreadyExistsException;
 use Drupal\bnf\Services\BnfImporter;
 use Drupal\bnf_client\Services\BnfExporter;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\drupal_typed\DrupalTyped;
 use Drupal\node\NodeForm;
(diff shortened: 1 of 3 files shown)


web/modules/custom/bnf/bnf_client/bnf_client.module
```diff
@@ -104,3 +124,36 @@ function bnf_client_form_node_form_submit(array $form, FormStateInterface $form_
...
+ *   The field definitions.
+ */
+function bnf_client_entity_base_field_info(EntityTypeInterface $entity_type): array {
+  $fields = [];
+
+  $field_name = 'bnf_keep_updated';
+
+  // Create new fields for node bundle.
+  if ($entity_type->id() === 'node') {
+    $fields[$field_name] = BaseFieldDefinition::create('boolean')
```
xendk, line 143: This is not what I had in mind.

`keep_updated` is not a state distinct from imported/exported, it is intrinsically bound to "imported".

We can just add `case LocallyClaimed = 3;` to `BnfStateEnum`, and the "backend" is basically implemented as the updating code only queues nodes where the state is `Imported`.

The form_alter then just add the checkbox when the state is either `LocallyClaimed` or `Imported`, and toggle between the two.

xendk (changes requested):

web/modules/custom/bnf/bnf_client/bnf_client.module
rasben: What is the pros of doing it that way? Because, a con I see is that it makes it more confusing and complicated when we want to allow the editors to filter imported nodes in various views.

xendk: Why would it make it more complicated? The pro is that we have guaranteed consistent data in one place. The backend doesn't have to check two fields in order to determine if it should sync. There's no risk of inconsistent data (what does "keep in sync" = true mean if status is none/exported?).

rasben: Alright, you've got it your way - I've updated the code :)

web/modules/custom/bnf/bnf_client/bnf_client.module
```diff
@@ -48,31 +49,66 @@ function bnf_client_form_alter(array &$form, FormStateInterface $form_state, str
...
+    [':input[name="bnf_export"]' => ['checked' => TRUE]];
 
-  $form['bnf'] = [
+  // Getting the BNF state, if it exists.
+  $state = BnfStateEnum::None;
+
+  if ($node->hasField(BnfStateEnum::FIELD_NAME) && !$node->get(BnfStateEnum::FIELD_NAME)->isEmpty()) {
+    /** @var \Drupal\enum_field\Plugin\Field\FieldType\EnumItemList $state_field */
+    $state_field = $node->get(BnfStateEnum::FIELD_NAME);
+    $states = $state_field->enums();
```
xendk, line 60: Doesn't `$state_field->enum` work? `FieldItemList` has a magic getter that'll get the named property from the first item.

I would think that `$state = node->get(BnfStateEnum::FIELD_NAME)->enum ?? BnfStateEnum::None;` ought to work.

xendk (approved):
Just one minor question.

web/modules/custom/bnf/bnf_client/bnf_client.module
rasben: Nope, neither `->enum` or `->enum()` works. Only `->enums()`

xendk: Weird, it should work, according to the module page:

https://www.drupal.org/project/enum_field

xendk (approved):

PR #2178, opened 2025-03-14
bnf: subscriptions

xendk:
Adds subscriptions with queueworkers.


web/modules/custom/bnf/bnf_server/src/Plugin/GraphQL/DataProducer/NewContentProducer.php
```diff
@@ -71,6 +72,10 @@ public function resolve(string $termUuid, string $since): NewContentResponse {
       return $result;
     }
 
+    // The `node_list` cache tag is cleared when saving any node, so we'll be
+    // able to catch both changes to existing nodes and new nodes.
+    $fieldContext->addCacheTags(['node_list']);
```
kasperg, line 77: Consider adding the cache information from the actual nodes using `$fieldContext->addCacheableDependency()`.

PR #2197, opened 2025-03-25
Add Drupal as a dependency for Mapp script

kasperg:
#### Description

[Our Mapp integration script uses `Drupal.behaviors`](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/4b949fd7c1706961aa8782408f6bab6ae6f1a5a1/web/modules/custom/dpl_mapp/js/dpl_mapp.js#L17-L17). If this is not defined then errors will be thrown. For whatever reason this has not occurred previously but it does now.

Deactivating JS aggregation fixes this but it is not a long term solution at the moment.

To address this we add the `core/drupal` library as a dependency.

#### Screenshot of the result

Before:

Monosnap | DPL CMS — Privat browsing 2025-03-25 22-16-38


kasperbirch1 (approved):
I have tested it and it works. good finding 👍

PR #2201, opened 2025-03-26
Extend library token with expire information

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-402

#### Description

Since we need the expire information about the library token in the graphql API we do two things:
* Store the expire value we get from Adgangsplatformen in the library-token-key/value-store along the token value
* Expose the token expire time as a DateTime graphql scalar in the same way as other dates are returned eg.: Entity::created time.


Dresse (approved):
I have 1 comment regarding formatting, otherwise it looks good 👍

PR #2203, opened 2025-03-27
Ddfher 221 cowboy button

xendk:
#### Link to issue

[DDFHER-221](https://reload.atlassian.net/browse/DDFHER-221)

#### Description

The cowboy button.

[DDFHER-221]: https://reload.atlassian.net/browse/DDFHER-221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/modules/custom/bnf/bnf_client/src/Form/BnfImportConfirmForm.php
```diff
@@ -45,9 +46,11 @@ public function __construct(
     protected LoggerInterface $logger,
     ConfigFactoryInterface $configFactory,
     EntityTypeManagerInterface $entityTypeManager,
+    TranslationInterface $stringTranslation,
   ) {
     $this->baseUrl = $configFactory->get(SettingsForm::CONFIG_NAME)->get('base_url');
     $this->nodeStorage = $entityTypeManager->getStorage('node');
+    $this->setStringTranslation($stringTranslation);
```
rasben, line 53: Why is this necessary? You dont seem to do anything in relation to translations in the rest of the file?

web/modules/custom/bnf/bnf_client/src/Form/BnfSyncForm.php
```diff
@@ -0,0 +1,83 @@
...
+   */
+  public function getFormId(): string {
+    return 'bnf_import_sync_form';
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state): array {
+    $form['#title'] = $this->t('Refresh all subscriptions?', [], ['context' => 'BNF']);
```
rasben, line 56: Please add a `#description` here also, so the user of the form knows what refresh actually means

Without having the background knowledge, It could also mean that all the existing content gets deleted :)

rasben (changes requested):
Looks good, two minor comments

web/modules/custom/bnf/bnf_client/src/Form/BnfImportConfirmForm.php
xendk: Well, this should have been a seperate fixup commit.

It is the proper way to do dependency injection with the traits. If you don't do this, `StringTranslationTrait` will fall back to getting the service using `\Drupal`, which defeats the purpose of DI.

rasben (approved):

PR #2204, opened 2025-03-28
Patch recurring events module to handle deleted owner users

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-74

#### Description

Currently our `RediaEvent` class expects `getOwner()` on event instances to return a `UserInterface` in accordance with the implemented `EntityOwnerInterface`.

However this is currently not the case if the owner user has been deleted. In this case `getOwner()` will return `NULL` which in turn causes an exception to be thrown in our own code.

The core error here is the fact that the Recurring Events module does not respect the return value specified by Drupal Core. To address this we patch the Recurring Events module to return the anonymous user instead of `NULL`.

This approach is in accordance with similar discussions elsewhere. See https://www.drupal.org/project/drupal/issues/3335025.


Dresse (approved):
LGTM 👍

PR #2207, opened 2025-03-30
BNF: Subscription UI, term-mapping and unpublishing handling

rasben:
# BNF: UI for toggling Delingstjenesten subscriptions.

- Add a button on the server, that acts like the "import content",
except is displayed on terms.
- When redirected back to the client site, a subscription management
form is now displayed
- This form allows the editor to create the linked subscription
- If the subscription already exists, the user is prompted to delete instead

DDFHER-198 DDFHER-222

# BNF: Allow 'mapping' of terms to subscriptions.

Updates the create-subscriptions form on the client site, to define
which categories and tags should be added to content imported.
DDFHER-198 DDFHER-222

# Expand BNF drush command to show dates in friendly format

Example:
```
 ----------------------------- ----------------------------- -------------------------- ------------------------ ------------------ ------------------ 
  UUID                          Subscription UUID             Categories for content     Tags for content         Created            Last update       
 ----------------------------- ----------------------------- -------------------------- ------------------------ ------------------ ------------------ 
  909db5c3-0ea9-4eeb-b9e7-f4b   cf26338e-7760-48e8-afac-63a   Design & Teknologi (15),   dans (14), modern (17)   2025-04-03 09:19   2025-04-03 09:19  
  f4665a772                     24e0f42ba                     Klima (11)                                          (1743664747)       (1743664748)      
 ----------------------------- ----------------------------- -------------------------- ------------------------ ------------------ ------------------ 
 ```

# BNF: Subscriptions to unpublish imported nodes. DDFHER-222
    
- Update the graphQL permissions, so we can see unpublished node data with GraphQL calls
- Update the importer, so it checks if the node it's trying to pull is unpublished, and only pulls it if it already exists on the client site
(meaning: 'dont import new, unpublished nodes')
 - Update the mappers to centralize identical code.

### CODE REVIEW

File: web/modules/custom/bnf/bnf_client/src/Plugin/QueueWorker/NodeUpdate.php [hunk 1]
```diff
@@ -67,7 +67,26 @@ public function __construct(
...
+
+      if ($node->hasField('field_tags') && !empty($data['tags'])) {
+        $tag_values = array_map(function ($term) {
+          return ['target_id' => $term->id()];
+        }, $data['tags']);
+
+        $node->set('field_tags', $tag_values);
+      }
+
+      $node->setPublished();
```
rasben, line 88: this is wrong :)

web/modules/custom/bnf/bnf_server/bnf_server.module
```diff
@@ -67,13 +98,12 @@ function bnf_server_theme(): array {
...
       'variables' => [
         'url' => NULL,
-        'name' => NULL,
         'label' => NULL,
       ],
     ],
     'bnf_server_missing_callback' => [
       'variables' => [
-        'uuid' => NULL,
+        'node_uuid' => NULL,
```
xendk, line 106: Good call. We should fix up `uuid` to `node_uuid`/'subscription_uuid`/`term_uuid` where appropriate, to avoid messing up, in general, but we can fix that up as we go.

web/modules/custom/bnf/bnf_server/src/Plugin/GraphQL/DataProducer/NewContentProducer.php
```diff
@@ -72,16 +71,18 @@ public function resolve(string $termUuid, string $since): NewContentResponse {
     }
 
     $query = $this->nodeStorage->getQuery();
-    $query->condition('created', $since->getTimestamp(), '>')
-      ->condition('status', Node::PUBLISHED);
+    $query->condition('changed', $since->getTimestamp(), '>');
```
xendk, line 74: We're sure that `scheduler` updates `changed` when publishing?

web/modules/custom/bnf/bnf_server/src/Plugin/GraphQL/DataProducer/NewContentProducer.php
```diff
@@ -90,7 +91,7 @@ public function resolve(string $termUuid, string $since): NewContentResponse {
       $result->uuids = array_map(fn ($node) => (string) $node->uuid(), $nodes);
 
       $youngest = array_reduce($nodes, fn ($youngest, $node) => max($youngest, $node->changed->value), 0);
-      $youngest = new DateTime('@' . $youngest);
+      $youngest = new DateTime("@$youngest");
```
xendk, line 94: Nice.

web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -55,23 +57,42 @@ public function getNodeTitle(string $uuid, string $endpointUrl): string {
...
-  public function importNode(string $uuid, string $endpointUrl): NodeInterface {
+  public function importNode(string $uuid, string $endpointUrl, bool $published): NodeInterface {
     $this->setEndpoint($endpointUrl);
 
     try {
       $response = GetNode::execute($uuid);
 
       $nodeData = $response->data?->node;
 
+      // If no nodedata could be find, it is possible that we've been informed
```
xendk, line 68: So I gather that the GraphQL endpoint doesn't return data for unpublished nodes?

So nodes that's created with a future publishing date isn't actually imported before it's published, and will throw an exception on each cron run. And should an error cause `GetNode` to return nothing, we'll happily start unpublishing exsting content.

I don't like it, but I can't see a solution right now (think DST is still messing with me).

xendk (changes requested):

web/modules/custom/bnf/bnf_client/src/Entity/Subscription.php
```diff
@@ -56,13 +57,67 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type): a
...
+      ])
+      ->setDisplayOptions('view', [
+        'label' => 'above',
+        'type' => 'entity_reference_label',
+        'weight' => 10,
+      ])
+      ->setDisplayConfigurable('form', TRUE)
+      ->setDisplayConfigurable('view', TRUE);
+
+    $fields['categories'] = BaseFieldDefinition::create('entity_reference')
```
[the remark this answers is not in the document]
rasben: Yep, I went by this solution, as it was easier to manage with code, and easier for the end user to understand.

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
```diff
@@ -0,0 +1,265 @@
...
+  public function getFormId(): string {
+    return 'bnf_subscription_form';
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state): array {
+    $uuid = $this->routeMatch->getParameter('uuid');
+    $label = $this->requestStack->getCurrentRequest()?->query->get('label');
```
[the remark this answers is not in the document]
rasben: The only reason I added it as a simple ?label=Hejsa is because we only need it here - unlike the node title, which is pulled from the whole graphql endpoint, we don't expose the term label as part of the subscription endpoint

Does that make sense? Do you rather suggest we expand the other functionality, to also save the label? But, then that might end up out-of-sync with BNF reality later on

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
```diff
@@ -0,0 +1,265 @@
...
+      '#title' => $this->t('Subscription label', [], ['context' => 'BNF']),
+      '#type' => 'textfield',
+      '#default_value' => $label ?? NULL,
+      '#disabled' => TRUE,
+    ];
+
+    $existingSubscriptions = $this->subscriptionStorage->loadByProperties(['subscription_uuid' => $uuid]);
+
+    if (!empty($existingSubscriptions)) {
+      $form['#title'] = $this->t('Confirm deletion of BNF subscription', [], ['context' => 'BNF']);
```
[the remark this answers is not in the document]
rasben: Agreed, but, for now it works :)
My idea was that when you/we have made the list-of-subscriptions interface, there should naturally be an edit-subscription endpoint. When that is done, I'll change this code ^ and just redirect the user to there instead.

What do you think?

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
```diff
@@ -0,0 +1,265 @@
...
+  /**
+   * Building a term-reference form element.
+   *
+   * @return mixed[]
+   *   A form element, to be used in $form.
+   */
+  private function getTermFormElement(string $vid): array {
+    $options = [];
+
+    $terms = $this->termStorage->loadByProperties(['vid' => $vid]);
```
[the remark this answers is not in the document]
rasben: What is the problem with loading the entire vocabulary? I would assume that it's limited how many tags and categories a site really has (?)

web/modules/custom/bnf/bnf_server/bnf_server.module
rasben: Yeah, it has also confused me somewhat, that we use `uuid` for everything :)

web/modules/custom/bnf/src/Services/BnfImporter.php
rasben: >So I gather that the GraphQL endpoint doesn't return data for unpublished nodes?

Exactly - it has an `accessCheck(TRUE)` somewhere in the code, which is what I'd also expect.

>So nodes that's created with a future publishing date isn't actually imported before it's published, and will throw an exception on each cron run

Hmm.. Well, I didn't consider that. One solution I can think of, is that we define 'NewContent' as nodes that has been published atleast once in its life time. What do you think of that?

>and will throw an exception on each cron run

..until any other valid content is actually imported - right?


>And should an error cause GetNode to return nothing, we'll happily start unpublishing exsting content.

Let's not make any errors then :D But nah, I think this is actually how I'd like it to work:
If a client site is told that an article is outdated, and it has to be kept up-to-date, then i would assume that it's logical that it gets unpublished

web/modules/custom/bnf/bnf_server/src/Plugin/GraphQL/DataProducer/NewContentProducer.php
rasben: I just double checked - yep, it updates `changed` :)

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
```diff
@@ -0,0 +1,265 @@
...
+  public function createSubscription(array &$form, FormStateInterface $form_state): void {
+    $uuid = $form_state->get('uuid');
+    $existingSubscriptions = $this->subscriptionStorage->loadByProperties([
+      'subscription_uuid' => $uuid,
+    ]);
+
+    // Even though we already have this check as part of the form building,
+    // we want to make sure that even if the form submission is triggered anyway
+    // we don't end up with duplicate subscriptions.
+    if (!empty($existingSubscriptions)) {
```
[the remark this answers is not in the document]
rasben: Nope, because the form can both create and delete (see my other comment :))

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
xendk: > Does that make sense?

Eh not really? Subscription endpoint? Save the label? I was just suggesting moving the label from query arguments to the path (`/admin/bnf/subscribe/{uuid}/{label}`) to not have to mess around with `Request`.

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
xendk: Well, didn't we agree on some meeting that there's no edit? And the most obvious implementation would make a delete link by the subscriptions uuid, not the BNF term ID.

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
xendk: Categories, yes, but terms is pretty much free for all. It might not be a big problem in practice, but it's the kind of sloppy thinking that's costing performance on a project level.

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
xendk: Well, there's another reason why you shouldn't mix creation and deletion functionality in one form.

web/modules/custom/bnf/bnf_server/bnf_server.module
xendk: I've shot ourselves in the foot with `subscription_uuid`. It's not the subscriptions uuid, but the subscribed (ie term) uuid.

web/modules/custom/bnf/src/Services/BnfImporter.php
xendk: > Hmm.. Well, I didn't consider that. One solution I can think of, is that we define 'NewContent' as nodes that has been published atleast once in its life time. What do you think of that?

Meh, sounds sketchy, what does "atleast once in its life time" mean when we consider revisions?

> ..until any other valid content is actually imported - right?

Well, okay, it'll "only" throw an exception whenever the unpublished node is touched on the BNF side and get considered "updated", but we shouldn't throw exceptions in what would be "normal". We could just ignore it, but that might mask other issues.

The "proper" solution would be if the GraphQL endpoint just returned the nodes updated, published or not, but that would require much recoding.

*sigh*

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
```diff
@@ -0,0 +1,265 @@
...
+   */
+  public function getFormId(): string {
+    return 'bnf_subscription_form';
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state): array {
+    $uuid = $this->routeMatch->getParameter('uuid');
```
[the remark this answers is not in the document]
rasben: Uh.. what? Are you sure about that?

I might be completely misunderstanding you, but do you mean like this? 

Screenshot 2025-04-03 at 16 22 49

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
rasben: Oh - well, in that case - it's to make it easier to handle symbols and spaces in the title.
If i wanna do `/admin/bnf/subscribe/{uuid}/{label}` then I need to both convert "Harry Potter" to "harry-potter" (and the other way around)

web/modules/custom/bnf/src/Services/BnfImporter.php
rasben: >The "proper" solution would be if the GraphQL endpoint just returned the nodes updated, published or not, but that would require much recoding.

Wouldn't it be a problem that the GraphQL endpoint could expose unpublished content? Like, I know it's far fetched, but what if Delingstjenesten has an article that they're working on that says "Vi lukker biblioteket på Læsø"

rasben: >eh, sounds sketchy, what does "atleast once in its life time" mean when we consider revisions?

A field that defaults to `false`. We add a `entity_post_save` hook (or whatever it's called), and if the entity is published, we set the custom field to `true`.

Done - that field will forever be true. What do you think of that?

web/modules/custom/bnf/bnf_server/bnf_server.module
rasben: Should we nip it in the bud and fix it now? Before real content is actually added?

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
rasben: I've updated it now, to have two seperate forms for create and delete
The reason I did it this way, was cause I thought Drupal would give some edit/delete forms straight out of the box, but after looking at how other contrib modules do, I can see that they're just basic forms.

So, you're right :)

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
xendk: Quite sure, but the argument must be defaulted.

There's only one existing example in the project: https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/b4839ce034df869a8164c489f815551636b1da89/web/modules/custom/collation_fixer/src/Form/CollationFixerForm.php#L85

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
xendk: That's what `rawurlencode()` is for. :wink:

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionForm.php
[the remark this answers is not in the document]
xendk: Not as such, but I also know if it gets merged it'll never get fixed.

web/modules/custom/bnf/bnf_server/bnf_server.module
xendk: Well, we already have deployed versions, so we're past the "just rename it" phase and at the "remove the old field, add the new", though we can argue that we can still get away with that without a "migrate data" step as there's no real subscription users yet.

web/modules/custom/bnf/src/Services/BnfImporter.php
xendk: > Wouldn't it be a problem that the GraphQL endpoint could expose unpublished content? 

Well, that's a consequence of the selected architecture. It ought to be possible to lock down the BNF related parts but I don't know if we'll have to create a new "server".

> Done - that field will forever be true. What do you think of that?

More hackery. :wink:

rasben: As we've discussed, I'll add an update that means that the graphQL endpoints can pull unpublished nodes, but only imports then if they already exist.

web/modules/custom/bnf/bnf_server/bnf_server.module
rasben: I guess it's annoying, but not critical. We'll leave it for now. Agreed?

web/modules/custom/bnf/bnf_client/src/Entity/Subscription.php
```diff
@@ -56,13 +62,67 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type): a
...
     // Using an 'uuid' would seem obvious, but the UUID field generates a new
-    // UUID as default value, so to avoid actidentially generating an UUID, we
+    // UUID as default value, so to avoid accidentally generating an UUID, we
     // use 'string' instead.
     $fields['subscription_uuid'] = BaseFieldDefinition::create('string')
       ->setLabel('Subscription UUID')
       ->setDescription('The UUID subscribed to.')
       ->setRequired(TRUE);
 
+    $fields['tags'] = BaseFieldDefinition::create('entity_reference')
```
xendk, line 72: You're still missing some code to update Drupals definitions. Try running:

```
task dev:bnfcli -- drush php-eval 'print_r(\Drupal::service(\"entity.definition_update_manager\")->getChangeList());'
```

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionCreateForm.php
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    try {
+      /** @var \Drupal\bnf_client\Entity\Subscription $subscription */
+      $subscription = $this->subscriptionStorage->create([
+        'subscription_uuid' => $uuid,
+      ]);
+
+      if ($form_state->getValue('only_new_content')) {
+        $subscription->setLast(time());
```
xendk, line 205: `time()` is frowned upon. See `Drupal\Component\Datetime\Time`, the `datetime.time` service.

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionDeleteForm.php
```diff
@@ -0,0 +1,171 @@
+<?php
```
xendk, line 1: You could have saved a lot of code by leaning into the entity system.

Adding the "delete-form" path to the links section of the Subscription entity annotation means you don't have to provide route for it. You could then add "delete" to the "form" handler to point at this class, which in turn could inherit from `ContentEntityDeleteForm` in order to add in our special needs.

Take a look at `Node` and `NodeDeleteForm` in core.

web/modules/custom/bnf/bnf_client/src/Plugin/QueueWorker/SubscriptionNewContent.php
```diff
@@ -98,6 +98,11 @@ public function processItem($data): void {
     );
 
     foreach ($newContent['uuids'] as $uuid) {
+      $this->nodeQueue->createItem([
+        'uuid' => $uuid,
+        'categories' => $subscription->getCategories(),
+        'tags' => $subscription->getTags(),
+      ]);
       $this->nodeQueue->createItem(['uuid' => $uuid]);
```
xendk, line 106: Huh? Double create?

web/modules/custom/bnf/bnf_server/bnf_server.module
xendk: Yeah.

xendk (approved):

web/modules/custom/bnf/bnf_client/src/Entity/Subscription.php
rasben: Uh.. what? I don't follow. 

```
task: [dev:bnfcli] docker compose -f docker-compose.yml -f docker-compose.bnf.yml run --rm bnfcli sh -c "drush php-eval 'print_r(\Drupal::service(\"entity.definition_update_manager\")->getChangeList());'"
configured lagoon-php.ini
Array
(
    [eventinstance] => Array
        (
            [field_storage_definitions] => Array
                (
                    [title] => 3
                )

        )

)

web/modules/custom/bnf/bnf_client/src/Plugin/QueueWorker/SubscriptionNewContent.php
rasben: Mistake, after a faulty rebase. Thank you!

web/modules/custom/bnf/bnf_client/src/Form/BnfSubscriptionDeleteForm.php
rasben: I did actually look at it, but I couldn't get it to work.
I think it's fine how it is :)

web/modules/custom/bnf/bnf_client/src/Entity/Subscription.php
rasben: but, maybe this is a non-issue? I guess once we merge it, all of this subscription stuff will be added only on new sites?

xendk: It worked here, I don't have `last` on the `subscriptions2` branch. But I do have `bnf_source_name`. Have you run updb recently?

It should just be: https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/21738031007ae499a26210573b99c7b316e70ce0/web/modules/custom/bnf/bnf_client/bnf_client.install#L19

rasben: Yep, `updb` runs as part of `drush deploy` - i even saw it run.
isnt it because you've run ` task dev:reset` whilst on the branch? cause then it creates the site from scratch

xendk (changes requested):
One last thing.

web/modules/custom/bnf/bnf_client/src/Entity/Subscription.php
rasben: @xendk I've found a solution here:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2207/commits/2e01b4bf23e9d10d078a84233080004dc3d12a11

I don't know if thats what you had in mind when we talked, but, atleast it works now.

rasben:
@xendk I've gotten rid of the custom forms. See the last 3 commits :)

web/modules/custom/bnf/bnf_client/src/Entity/Subscription.php
xendk: That was exactly what I had in mind. I'm sure that someone will point out that it's not correct if the field changes again in the future and all update hooks haven't been run, just as in the old days with `hook_schema` and update hooks, but this is good enough for jazz.

xendk (approved):
Much better.

PR #2213, opened 2025-03-31
Unignore system.performance configuration

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-76

#### Description

Previously we have disabled JS aggregation to fix JavaScript problems
caused by faulty dependency chains. This has caused the
system.performance configuration to be ignored.

Now that we have addressed this  in #2197 we can unignore the configuration.


xendk (approved):
:+1:

PR #2214, opened 2025-03-31
Add `field_tags` to `node.page`.

LasseStaus:
Similarly to the eventseries/instances and articles, these should be displayed in a `tagging` details siderbar form group.

Categories have not been added, only tags.

DDFSAL-73

#### Link to issue

[jira.](https://reload.atlassian.net/jira/software/c/projects/DDFSAL/boards/539?selectedIssue=DDFSAL-73&useStoredSettings=true)

Design system pr: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/876

#### Description

This PR Adds a sidebar group "tagging" with the option to add tags to regular pages. 

Categories are not included. 

#### Screenshot of the result

image

image

### DIFF
--- a/web/themes/custom/novel/templates/layout/node--page--full.html.twig
+++ b/web/themes/custom/novel/templates/layout/node--page--full.html.twig
@@ -6,6 +6,7 @@
       <h1 class="article-header__title">{{ title }}</h1>
 
       {{ content.field_subtitle }}
+
     </header>
   {% else %}
     {# We want the title to still be available for SEO/Accessibility. #}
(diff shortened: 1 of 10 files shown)


web/themes/custom/novel/templates/layout/node--page--full.html.twig
```diff
@@ -6,6 +6,10 @@
       <h1 class="article-header__title">{{ title }}</h1>
 
       {{ content.field_subtitle }}
+
+      {% if (content.field_tags.0) %}
+          {{ content.field_tags }}
+      {% endif %}
```
Dresse, line 12: According to the JIRA task, the tags should be displayed after paragraphs and before the footer, as it it will look weird if the page does not contain header/subh_title (which is possible in some cases).

Dresse (changes requested):
The configuration looks fine, but the placement of the tags should be moved to after paragraphs and before the footer. Otherwise looks good 👍

kasperg (approved):

PR #2224, opened 2025-04-01
BNF: Prepare permissions for BNF. DDFHER-225

rasben:
- Enable the base BNF module
- Move permissions from bnf_client to bnf (This is necessary to set the permissions for the existing roles)
- Create new `bnf_pilot` role, that is allowed to export to Delingstjenesten.


config/sync/administerusersbyrole.settings.yml
```diff
@@ -10,3 +10,4 @@ roles:
   bnf_graphql_client: unsafe
   go_graphql_client: unsafe
   go_editor: unsafe
+  bnf_pilot: unsafe
```
kasperg, line 13: 👌

kasperg (approved):
:+1: LGTM

PR #2241, opened 2025-04-02
Add update hook to make sure LibraryToken has correct format.

Dresse:
#### Link to issue

N/A

#### Description

After updating the signature of the LibraryToken to also contain expire time, it is no longer saved as a string but as a object instead, containing the token and expire time. 

However, this is causing trouble for already existing LibraryTokens saved in the KeyValueStore. To make sure all libraries are using a LibraryToken with the correct format, we add an update hook that checks if the current token is valid, and if not, we delete it and fetch a new one.


spaceo (approved):
LGTM 👍

PR #2244, opened 2025-04-02
Remember to set config_ignore mode

xendk:
Of course it blows up if the mode doesn't match the config data.


arnested (approved):
LGTM :+1:

PR #2258, opened 2025-04-07
Update DateFormatter to match other event filter logic. DDFSAL-27

rasben:
Our various views and other queries, filter out expired events, by the logic of: "We must be past the *end* date".

The logic in ReoccuringDateFormatter instead looked at the *start* date. This is a problem if you have a series with a single active eventinstance, that spans over several days: It'll show as expired.

DDFSAL-27


Dresse (approved):
LGTM 👍

PR #2267, opened 2025-04-08
BNF: Support more content types for export/import. DDFHER-168

rasben:
I would suggest to review this, commit by commit, as it touches on a lot of different sections.

-----

## Add `field_canonical_url` to Page CT in Drupal. DDFHER-168

The field already exists on articles, but not on pages. It is needed, when we start to support exporting pages as part of BNF.

## Require body field as part of text body paragraph.

The GoTextBody paragraph has it required, and we would like to streamline the two, so the GraphQL is simpler.

## Export CT `bundle` as part of GraphQL NodeInterface.

We need it when we want to do Imports, as we need to pass along the bundle.

## ~~Expand GetNodeTitle GraphQL to include bundle also. DDFHER-168~~ See last commit
 
Up until now, we only needed the title as part of the BNF import form - however soon we will need to support more CTs than just article, and for that, we need to also pass along the source bundle.
To do this, we rename GetNodeTitle to GetNodeMetaData, and expand it to also include the bundle.

## ~~Pass along ContentType as part of BNF importer. DDFHER-168~~ See last commit

As an alternative to just hardcoding `article`.

## Re-enable and re-export all GraphQL fields with Sailor

This enables all fields and paragraphs as part of the BNF-supported content types (article, page, go_article, go_page, go_category), and re-exports it using sailor.
Some of the fields have changed (such as name and `required`), so some of the mappers needed to be updated.
I've also updated the `::make`'s in the testers to have named arguments, as we cannot assume the order in `node.graphql`, as it makes it impossible to manage.

## BNF: Support more content types for export/import. DDFHER-168

- Expand the exporter/importer to support `article`, `page`, `go_article`, `go_page`, `go_category`
- Add missing mappers for the new content types
- Add missing mappers for missing GO paragraphs
- Add a SoundTrait for supporting Audio media, as part of `go_category`


web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -37,26 +46,26 @@ public function __construct(
...
-    return $nodeData->title;
+    return $nodeData;
   }
 
   /**
    * Importing a node from a GraphQL source endpoint.
    */
-  public function importNode(string $uuid, string $endpointUrl, string $nodeType = 'article'): NodeInterface {
+  public function importNode(string $uuid, string $nodeType, string $endpointUrl): NodeInterface {
     if (!in_array($nodeType, self::ALLOWED_CONTENT_TYPES)) {
```
xendk, line 69: You've extendend an API, intrduced five classes and spent effort carrying the type through forms to importNode, only for this check, other than that, it's just used for logging that could live without.

But we can't import a node without a mapper, and the mapper manager will throw an exception if it can't find a mapper for the GraphQL object fetched. And the GraphQL class is implicitly bound to the node type. So it'll only map content types we "allow" by actually implementing them.

So this check, and the code supporting it, is not needed.

rasben: jesus christ, you're right.. i think it's because the importer *used* to need the content type, before you did the big overhaul, and i never realized until now

nghh.. i'll fix it.

rasben: Fixed here: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2267/commits/be9d7ef535ea9d03aaa34bb56d8dc36b285ac47a

PR #2270, opened 2025-04-08
BNF: Export from client should be published.

rasben:


web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -103,7 +103,7 @@ public function importNode(string $uuid, string $endpointUrl, string $nodeType =
 
       $node->set(BnfStateEnum::FIELD_NAME, BnfStateEnum::Imported);
 
-      $node->set('status', NodeInterface::NOT_PUBLISHED);
+      $node->set('status', $published);
```
xendk, line 106: `status` isn't `true`/`false` but `0`/`1` (`NodeInterface::NOT_PUBLISHED`/`NodeInterface::PUBLISHED`).

Either we should use a trinary here (`$node->set('status', $published ? NodeInterface::PUBLISHED : NodeInterface::NOT_PUBLISHED);` or make the argument an int and use `$this->importer->importNode($uuid, $callbackUrl, $node_type, NodeInterface::PUBLISHED);` to make things more obvious.

xendk (changes requested):

web/modules/custom/bnf/src/Services/BnfImporter.php
rasben: I'm pretty sure it still worked, but, I guess being explicit is always nice :)
I've updated it.

xendk: Yeah, it's not a question about whether it works, but about code readability.

xendk (approved):

PR #2272, opened 2025-04-08
BNF: Fixing error on export

rasben:
Whenever you export a node from the client to the server, it throws an error, despite actually working.
This has already been fixed as part of subscription, but I'm so tired of seeing it, that I've pulled it out seperately.


rasben:
This is not my code - it's made by @xendk in https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2178 where I've approved it, so I'm gonna bypass the merge rules :)

PR #2275, opened 2025-04-09
Bnf example content

xendk:
Add a BNF example content module for development.

Without any front page, "Login to BNF" will just end up at the login of the BNF site, without any indication that you're in fact "logged in".

### DIFF
--- a/dev-scripts/install-site.sh
+++ b/dev-scripts/install-site.sh
@@ -7,7 +7,7 @@ SKIP_LANGUAGE_IMPORT=""
 while [[ "$#" -gt 0 ]]; do
   case $1 in
     --skip-language-import) SKIP_LANGUAGE_IMPORT="true" ;;
-    --no-content) SKIP_CONTENT="true" ;;
+    --bnf-content) BNF_CONTENT="true" ;;
     *) echo "Unknown parameter passed: $1"; exit 1 ;;
   esac
   shift
@@ -42,8 +42,10 @@ curl --silent --show-error --fail --output /dev/null http://varnish:8080/
 # Enable dev modules (see task dev:enable-dev-tools).
 MODULES=(devel field_ui purge_ui restui uuid_url views_ui dblog)
 
-if [[ $SKIP_CONTENT != "true" ]]; then
+if [[ $BNF_CONTENT != "true" ]]; then
   MODULES=("${MODULES[@]}" dpl_example_content)
+else
+  MODULES=("${MODULES[@]}" bnf_example_content)
 fi
 drush install -y "${MODULES[@]}"
 
--- a/web/modules/custom/bnf/bnf_example_content/bnf_example_content.module
+++ b/web/modules/custom/bnf/bnf_example_content/bnf_example_content.module
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Logger\LoggerChannelFactoryInterface;
+use Drupal\drupal_typed\DrupalTyped;
+
+/**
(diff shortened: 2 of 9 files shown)


web/modules/custom/bnf/bnf_example_content/bnf_example_content.module
```diff
@@ -0,0 +1,28 @@
...
+ * @param string[] $modules
+ *   The names of the modules that were installed.
+ */
+function bnf_example_content_modules_installed(array $modules): void {
+  // Only react to our own installation.
+  if (!in_array('bnf_example_content', $modules)) {
+    return;
+  }
+
+  $logger = DrupalTyped::service(LoggerChannelFactoryInterface::class, LoggerChannelFactoryInterface::class)->get('bnf_example_content');
```
kasperg, line 21: Consider if `DrupalTyped` is necessary here.

When retrieving loggers I think `\Drupal::logger()` is fine.

web/modules/custom/bnf/bnf_example_content/bnf_example_content.module
```diff
@@ -0,0 +1,28 @@
...
+ * Implements hook_modules_installed().
+ *
+ * @param string[] $modules
+ *   The names of the modules that were installed.
+ */
+function bnf_example_content_modules_installed(array $modules): void {
+  // Only react to our own installation.
+  if (!in_array('bnf_example_content', $modules)) {
+    return;
+  }
```
kasperg, line 19: I wonder why this is implemented in `hook_modules_installed()` instead of `hook_install()` 🤔

dev-scripts/install-site.sh
```diff
@@ -42,8 +42,10 @@ curl --silent --show-error --fail --output /dev/null http://varnish:8080/
 # Enable dev modules (see task dev:enable-dev-tools).
 MODULES=(devel field_ui purge_ui restui uuid_url views_ui dblog)
 
-if [[ $SKIP_CONTENT != "true" ]]; then
+if [[ $BNF_CONTENT != "true" ]]; then
   MODULES=("${MODULES[@]}" dpl_example_content)
+else
+  MODULES=("${MODULES[@]}" bnf_example_content)
```
kasperg, line 48: I would appreciate a few more thoughts/commentary here.

As I see it this approach prevents us from not installing any content during installation as you can either pass `--bnf-content` to get `bnf_example_content` or not to get `dpl_example_content`.

I wonder whether you considered an approach where we support all three options with `--dpl-content`, `--bnf-content` or an argument `--content=bnf`, `--content=dpl` etc.

I recognize that we probably do not use the `--no-content` option at all at the moment though.

kasperg (approved):
:+1: I have some thoughts for you to consider before you merge.

web/modules/custom/bnf/bnf_example_content/bnf_example_content.module
xendk: Well, you tell me why you did it in `dpl_example_content`? :grinning:

kasperg: Heh. I initially implemented default opening hours in `hook_modules_installed()` because the code depends on the availability of default content which in itself is created by `hook_modules_installed()` implemented by `default_content`.

Perhaps the situation is the same for the front page since the alias `/delingstjenesten` will only exist if the BNF example content has been enabled.

dev-scripts/install-site.sh
xendk: > I recognize that we probably do not use the --no-content option at all at the moment though.

Exactly my thinking. `--no-content` was added because BNF needed to avoid the standard content, which was mostly because it made it rather confusing when you had two sites that were initially identical.

PR #2284, opened 2025-04-11
Activate the graphql_compose_views module

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-531

#### Description

Activate the graphql_compose_views module
...which is needed to be able to query the go_categories.


Dresse (approved):
LGTM 👍

PR #2290, opened 2025-04-14
Make clearing varnis work in bnf

xendk:
Make `drush cache:rebuild-external` work for `bnfcli`.

### DIFF
--- a/web/sites/bnf/settings.php
+++ b/web/sites/bnf/settings.php
@@ -8,6 +8,12 @@
  * settings.lagoon.php that contains Lagoon-specific settings.
  */
 
+// Override the configured varnish hostname so Drupal will purge the right
+// instance. The hex numbers is the ID of the configured purgers, these should
+// follow the ones in the exported configuration.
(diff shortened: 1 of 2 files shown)


web/sites/bnf/settings.php
```diff
@@ -8,6 +8,9 @@
  * settings.lagoon.php that contains Lagoon-specific settings.
  */
 
+$config['varnish_purger.settings.30cd45a1b1']['hostname'] = 'bnfvarnish';
+$config['varnish_purger.settings.65fc931232']['hostname'] = 'bnfvarnish';
```
rasben, line 12: Where are you getting these values from? (30cd45a1b1 / 65fc931232)

xendk: Ah, I forgot a comment.

rasben: Can these values change? I'm a bit worried that one day, this will break, because of a re-export

Can we set this with a hook instead, or something?

xendk: They change if someone deletes the purgers and create new. Which is not supposed to happen.

rasben (approved):

PR #2294, opened 2025-04-15
BNF: Fallback author

rasben:
Setting 'The library' as a static fallback for imported nodes.

Also, forces show_override_author to always be visible.


web/modules/custom/bnf/tests/src/Unit/Mapper/NodeArticleMapperTest.php
```diff
@@ -71,29 +76,31 @@ public function testNodeArticleMapping(): void {
...
     $this->entityProphecy->set('field_subtitle', 'this is the subtitle')->shouldHaveBeenCalled();
-    $this->entityProphecy->set('field_override_author', 'this is an author')->shouldHaveBeenCalled();
-    $this->entityProphecy->set('field_show_override_author', TRUE)->shouldHaveBeenCalled();
     $this->entityProphecy->set('field_publication_date', ["value" => "2025-01-01"])->shouldHaveBeenCalled();
     $this->entityProphecy->set('field_teaser_text', 'this is a teaser text')->shouldHaveBeenCalled();
+    $this->entityProphecy->set('field_teaser_image', [])->shouldHaveBeenCalled();
+    $this->entityProphecy->set('field_canonical_url', ["uri" => "https://example.dk"])->shouldHaveBeenCalled();
+    $this->entityProphecy->set('field_paragraphs', [])->shouldHaveBeenCalled();
+
+    // Testing that the mapper always sets this to true.
```
xendk, line 99: Ah, but you're missing a test: that it import the override author if there's one on the origin node.

xendk (approved):
You missed the opportunity to actually write a test for new functionality you introduced.

And it would've been a prime example of when TDD works well.

PR #2295, opened 2025-04-15
BNF: Unit tests for all supported paragraphs and CTs.

rasben:


web/modules/custom/bnf/src/Plugin/bnf_mapper/ParagraphMaterialGridAutomaticMapper.php
```diff
@@ -32,6 +32,7 @@ public function map(ObjectLike $object): mixed {
       'field_material_grid_title' => $object->materialGridTitle,
       'field_material_grid_description' => $object->materialGridDescription,
       'field_cql_search' => ['value' => $object->cqlSearch->value],
+      'field_amount_of_materials' => $object->amountOfMaterials,
```
rasben, line 35: The unit tests helped me catch that this was missing :) #winning

xendk (approved):

PR #2297, opened 2025-04-16
BNF: Handle nodes when subscription gets deleted. DDFHER-219

rasben:
- Create new field on the nodes, to reference what subscription (if any) this node originates from
- Fill out field, when NodeUpdate is run
- Add hook for deleting subscription, that removes reference from relevant nodes
- If no subscriptions exists on these nodes anymore, we locally claim them, to stop them from being updated.

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -180,3 +180,9 @@ environments:
         schedule: "M/15 * * * *"
         command: drush cron
         service: cli
+  pr-2297:
+    cronjobs:
+      - name: drush cron
+        schedule: "M/15 * * * *"
+        command: drush cron
+        service: cli


web/modules/custom/bnf/bnf_client/src/Hook/SubscriptionHooks.php
```diff
@@ -38,4 +43,47 @@ public function queueUpdate(EntityInterface $entity): void {
...
+        ->condition($field_key, [$entity->id()], 'IN')
+        // We want all nodes, even if the user does not have access.
+        ->accessCheck(FALSE)
+        ->execute();
+
+    /** @var \Drupal\node\NodeInterface[] $nodes */
+    $nodes = $nodeStorage->loadMultiple($nids);
+
+    // Looping through the nodes, and remove the reference to this subscription.
+    foreach ($nodes as $node) {
```
xendk, line 70: This could potentially be quite a few nodes. Are we sure we'll be able to handle them all in `hook_delete`?

web/modules/custom/bnf/bnf_client/src/Hook/SubscriptionHooks.php
```diff
@@ -38,4 +43,47 @@ public function queueUpdate(EntityInterface $entity): void {
...
+      $subscriptionIds = $node->get($field_key)->getValue();
+
+      $subscriptionIds = array_unique(array_column($subscriptionIds, 'target_id'));
+
+      // Removing any references to our subscription (even if duplicates).
+      $subscriptionIds = array_diff($subscriptionIds, [$entity->id()]);
+      $node->set($field_key, $subscriptionIds);
```
xendk, line 77: This works?  When you get the field data it's an array of arrays with a `target_id`. Here you set the value to be an array of IDs.

web/modules/custom/bnf/bnf_client/src/Plugin/QueueWorker/NodeUpdate.php
```diff
@@ -74,6 +74,20 @@ public function processItem($data): void {
...
+      // Link the subscription to the node, for when we handle deleting of the
+      // subscription in the future.
+      if ($node->hasField('bnf_source_subscriptions')) {
+        $subscriptionIds = $node->get('bnf_source_subscriptions')->getValue();
+        $subscriptionIds[] = $data['subscription_id'];
+
+        // If for whatever reason a node is updated, we don't want to add
+        // the same subscription ID several times.
+        // This line makes sure that the subscription IDs are unique.
+        $subscriptionIds = array_unique(array_column($subscriptionIds, 'target_id'));
```
xendk, line 86: Are you letting ChatGPT take the wheel again? Above you get the field vailue, which should be array of arrays with `target_id`. Then you add in the new ID, but as a value (no `target_id`), and here you then collect the `target_id` values and sets it as the new value.

web/modules/custom/bnf/src/BnfStateEnum.php
```diff
@@ -14,5 +14,11 @@ enum BnfStateEnum: int {
 
   case Exported = 2;
 
+
+  // The content has been imported from a third party source (such as BNF), but
```
xendk, line 18: Well, shouldn't this come with #2177 ?

xendk (approved):

PR #2304, opened 2025-04-22
Ddfher 156 syncing references

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-227

#### Description

Import GO link references.


web/modules/custom/bnf/src/Plugin/GraphQL/DataProducer/LinkedContentUuidProducer.php
```diff
@@ -0,0 +1,81 @@
...
+   * Get the UUID.
+   *
+   * @param array{title: string, url: string, internal: bool} $link
+   *   The link field data.
+   */
+  public function resolve($link): ?string {
+    // While the link field contains the path `internal:node/19`, and it stores
+    // which entity type and ID the link points to for internal links, all we
+    // have to work with here is the external GraphQL representation.
+    if ($link['internal']) {
```
rasben, line 66: what if ['internal'] is not set?

web/modules/custom/bnf/src/Plugin/GraphQL/DataProducer/LinkedContentUuidProducer.php
```diff
@@ -0,0 +1,81 @@
...
+      $path = $this->aliasManager->getPathByAlias($link['url']);
```
rasben, line 67: what if ['url'] is not set?

web/modules/custom/bnf/src/Plugin/bnf_mapper/BnfMapperNodePluginBase.php
```diff
@@ -96,20 +96,7 @@ public function getNode(NodeArticle|NodePage|NodeGoArticle|NodeGoCategory|NodeGo
...
-        $paragraphs[] = $this->manager->map($paragraph);
-      }
-      catch (\Exception $e) {
-        $this->logger->error('Unable to map paragraph of type - skipping: ' . $e->getMessage());
-      }
-    }
-
-    return $paragraphs;
-
+    return $this->manager->mapAll($object->paragraphs ?? []);
```
rasben, line 99: What about the "skiping paragraphs that are not supported" logic, that you're removing?

From what i can see, you've not introduced that as part of mapAll() ?

web/modules/custom/bnf/src/Plugin/bnf_mapper/ParagraphGoLinkboxMapper.php
```diff
@@ -47,33 +48,28 @@ public function map(ObjectLike $object): mixed {
...
       'type' => 'go_linkbox',
       'field_go_color' => $object->goColor,
       'field_go_description' => $object->goDescription,
       'field_go_image' => $this->getImageValue($object->goImage),
-      'field_go_link_paragraph' => [[
-        'target_id' => $goLinkParagraph->id(),
-        'target_revision_id' => $goLinkParagraph->getRevisionId(),
-      ],
+      'field_go_link_paragraph' => [
+        $goLinkParagraph,
```
rasben, line 61: ?? Below you're doing the same, although in a `if ($goLinkParagraph) {`?

web/modules/custom/bnf/src/Services/ImportContextStack.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\bnf\Services;
+
+use Drupal\bnf\ImportContext;
+
+/**
+ * Manages the current import context.
```
rasben, line 10: I know what it does, and you know what it does, but for future readers, this will mean nothing.

Can you expand the documentation?

rasben (changes requested):

web/modules/custom/bnf/src/Plugin/GraphQL/DataProducer/LinkedContentUuidProducer.php
xendk: Why shouldn't it? Have you seen it not be?

web/modules/custom/bnf/src/Services/BnfImporter.php
```diff
@@ -60,7 +62,13 @@ public function getNodeTitle(string $uuid, string $endpointUrl): string {
   /**
    * Importing a node from a GraphQL source endpoint.
    */
-  public function importNode(string $uuid, string $endpointUrl): NodeInterface {
+  public function importNode(string $uuid, string|ImportContext $importContext): NodeInterface {
+    if (!$importContext instanceof ImportContext) {
+      $importContext = new ImportContext($importContext);
```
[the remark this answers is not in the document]
xendk: Compatibility code. The idea is to fix all callers to supply an ImportContext instead.

web/modules/custom/bnf/src/Plugin/bnf_mapper/ParagraphGoLinkboxMapper.php
xendk: The go link mapper can return `null` if we hit the recursion limit, so we should only set the field when we have a value.

web/modules/custom/bnf/src/Plugin/GraphQL/DataProducer/LinkedContentUuidProducer.php
xendk: Can it be, or do you mean empty?

If the URL is empty, `getPathByAlias` won't find an alias and just return the given "path", which then wont be matched by the following `preg_match`.

web/modules/custom/bnf/src/Plugin/GraphQL/DataProducer/LinkedContentUuidProducer.php
rasben: i don't know if i've seen it, but seeing as this is just accessing an loose array, i think it is common to add a check if it actually exists

web/modules/custom/bnf/src/Plugin/GraphQL/DataProducer/LinkedContentUuidProducer.php
rasben: same as other comment

web/modules/custom/bnf/src/Plugin/bnf_mapper/ParagraphGoLinkboxMapper.php
rasben: i don't understand your answer here.

if you look at your code, what you're doing at line 60-62 is identical to what you're doing at line 66-72 - but at 66-72, you're actually checking if the value exists

web/modules/custom/bnf/src/Services/BnfImporter.php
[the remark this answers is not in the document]
rasben: What? I'm asking if you can update this line to instead be 

```
$importContext = new ImportContext(endpointUrl: $importContext);
```

rasben (approved):

rasben (approved):

PR #2305, opened 2025-04-22
DDFSAL-77 - Exclude selected routes from Mapp tracking logic

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-77

#### Description
This pull request disables tracking for the `search_result` and `work` pages. 
Tracking should instead be handled within the React application, where the necessary data is available. This change also helps avoid unnecessary calls.

#### Test
https://varnish.pr-2305.dpl-cms.dplplat01.dpl.reload.dk/

#### Dev links
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-2305

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,10 +40,10 @@
             "type": "package",
             "package": {
                 "name": "danskernesdigitalebibliotek/dpl-react",
-                "version": "2025.18.0",
+                "version": "0.0.0-dev",
                 "type": "drupal-library",
                 "dist": {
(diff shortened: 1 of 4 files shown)


kasperbirch1:
@kasperg is assigned to address the issue where JavaScript is disabled on the search, advanced search, and work pages, causing Cypress tests to fail. 

In the PR environment, we fixed this by avoiding the use of aggregated JS.

I attempted to fix it after discussing the issue with @kasperg and @Dresse, but the change here: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2305/commits/1f684e876d2460e45e8afbed829701abcdc67294 does not resolve the problem.

kasperg (approved):
:+1: LGTM

PR #2307, opened 2025-04-23
Remove entityqueue

kasperg:
#### Description

~~We~~ I accidentally added the module with release 2025.17.0 in #2301 . We do not use it so remove it again.


Dresse (approved):
LGTM 👍

PR #2339, opened 2025-04-29
Ensure that neither library nor user token are cached

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-562

#### Description

Ensure that neither library nor user token are cached in graphql data producers.

### DIFF
--- a/web/modules/custom/dpl_update/dpl_update.install
+++ b/web/modules/custom/dpl_update/dpl_update.install
@@ -10,6 +10,8 @@
  */
 
 use Drupal\Core\Extension\ModuleInstallerInterface;
+use Drupal\dpl_library_token\LibraryTokenHandler;
+use Drupal\dpl_login\Adgangsplatformen\Config;
 use Drupal\drupal_typed\DrupalTyped;
 use Drupal\locale\SourceString;
 use Drupal\locale\StringDatabaseStorage;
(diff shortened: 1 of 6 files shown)


web/modules/custom/dpl_update/dpl_update.install
```diff
@@ -572,3 +574,37 @@ function dpl_update_update_10046(): string {
...
+  }
+  $handler->setToken(token: $newToken);
+
+  $success_output[] = t('New token: Token: @token, properties: @props', [
+    '@token' => $newToken->token,
+    '@props' => implode(', ', array_keys((array) $newToken)),
+  ]);
+
+  return implode(' | ', $success_output);
+}
```
Dresse, line 610: Should this update hook be placed in the dpl_library_token.install file instead like we did last time we needed to generate a new token?

Maybe it doesn't matter that much.

Dresse (approved):
I have a single comment, but it is not a dealbreaker, just a thought.

LGTM otherwise 👍

web/modules/custom/dpl_update/dpl_update.install
spaceo: It's a good question. However because of the time pressure I choose to leave it as is. I have tested it in this state.

PR #2347, opened 2025-05-02
Ddfbra 545 go links

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-545

#### Description

Use a path processor to make intersite links "external".

Best reviewed commit by commit, there's quite a bit of cleanup in the process.

### DIFF
--- a/web/modules/custom/dpl_go/src/GoSite.php
+++ b/web/modules/custom/dpl_go/src/GoSite.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Drupal\dpl_go;
 
+use Drupal\Core\Session\AccountInterface;
 use Drupal\dpl_lagoon\Services\LagoonRouteResolver;
 use function Safe\parse_url;
 
@@ -12,7 +13,21 @@ use function Safe\parse_url;
  */
 class GoSite {
 
-  public function __construct(protected LagoonRouteResolver $lagoonRouteResolver) {}
+  public function __construct(
+    protected LagoonRouteResolver $lagoonRouteResolver,
+    protected AccountInterface $currentUser,
+  ) {}
+
+  /**
+   * Is the current request considered "the Go site".
+   *
+   * This is true when the current user is the Go GraphQL consumer user that the
+   * React front-end uses. Which is the only user given the "rewrite go urls"
+   * permission.
+   */
+  public function isGoSite(): bool {
+    return $this->currentUser->hasPermission('rewrite go urls');
+  }
 
   /**
    * Get the base URL for the Go site.
--- a/web/modules/custom/dpl_go/src/GoSite.php
+++ b/web/modules/custom/dpl_go/src/GoSite.php
@@ -27,9 +27,18 @@ class GoSite {
     if ($mainRoute = $this->lagoonRouteResolver->getMainRoute()) {
       $urlParsed = parse_url($mainRoute);
 
-      // These two parts are required
+      // These two parts are required.
       if (isset($urlParsed['scheme']) || isset($urlParsed['host'])) {
-        $goDomain = sprintf('%s://go.%s%s', $urlParsed['scheme'], $urlParsed['host'], $urlParsed['path'] ?? '');
+        $host = $urlParsed['host'];
+
+        if (str_starts_with($host, 'www.')) {
+          $host = str_replace('www.', 'www.go.', $host);
+        }
+        else {
+          $host = "go.{$host}";
+        }
+
+        $goDomain = sprintf('%s://%s%s', $urlParsed['scheme'], $host, $urlParsed['path'] ?? '');
       }
     }
 
--- a/web/modules/custom/dpl_go/src/GoSite.php
+++ b/web/modules/custom/dpl_go/src/GoSite.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\dpl_go;
+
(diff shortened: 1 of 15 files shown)


web/modules/custom/dpl_go/src/GoSite.php
```diff
@@ -12,7 +13,21 @@
...
+
+  /**
+   * Is the current request considered "the Go site".
+   *
+   * This is true when the current user is the Go GraphQL consumer user that the
+   * React front-end uses. Which is the only user given the "rewrite go urls"
+   * permission.
+   */
+  public function isGoSite(): bool {
+    return $this->currentUser->hasPermission('rewrite go urls');
```
spaceo: I like that it is a permisson. That is a clean solution 👍

web/modules/custom/dpl_go/src/GoSite.php
```diff
@@ -77,4 +78,11 @@ public function getGoBaseUrl(): string {
     return $goDomain;
   }
 
+  /**
+   * Determine if the given node is a Go node.
+   */
+  public function isGoNode(EntityInterface $node): bool {
+    return str_starts_with($node->bundle(), 'go_');
```
spaceo, line 89: Again we talked about detecting a go node otherwise. Maybe we could write a todo about a future implementation here.

web/modules/custom/dpl_go/test/src/Unit/PathProcessor/OutboundPathProcessorTest.php
```diff
@@ -0,0 +1,126 @@
...
+use Drupal\dpl_go\GoSite;
+use Drupal\dpl_go\PathProcessor\OutboundPathProcessor;
+use Drupal\Tests\UnitTestCase;
+use Prophecy\Prophecy\ObjectProphecy;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Test the URL rewriting.
+ */
+class OutboundPathProcessorTest extends UnitTestCase {
```
spaceo, line 20: Nice with a test for it!

web/modules/custom/dpl_go/src/GoSite.php
Dresse: What did you discuss? The creation of a goNodeInterface?

web/modules/custom/dpl_go/test/src/Unit/PathProcessor/OutboundPathProcessorTest.php
```diff
@@ -0,0 +1,126 @@
...
+    $this->pathProcessor = new OutboundPathProcessor(
+      $this->goSite->reveal(),
+      $typeManager->reveal(),
+    );
+  }
+
+  /**
+   * Test that non-node paths are passed through unmolested.
+   */
+  public function testNonNodesArePassedThrouh(): void {
```
Dresse, line 62:
```suggestion
  public function testNonNodesArePassedThrough(): void {
```

Small typo fix

web/modules/custom/dpl_go/src/PathProcessor/OutboundPathProcessor.php
```diff
@@ -0,0 +1,84 @@
...
+      $node = $this->nodeStorage->load($pathParts[2]);
+
+      if ($node) {
+        $isGoNode = $this->goSite->isGoNode($node);
+
+        if ($isGoNode xor $this->goSite->isGoSite()) {
+          $options['base_url'] = $isGoNode ?
+            $this->goSite->getGoBaseUrl() :
+            $this->goSite->getCmsBaseUrl();
+        }
```
Dresse, line 77: After we talked about this part, it made sense to me. But maybe we could add a small comment explaining this bit?
Or would that just be clutter?

Dresse (approved):
Really good work 👍

web/modules/custom/dpl_go/src/GoSite.php
xendk: Yeah, Bundle classes. Which, by the way, was announced october 2021...

Arne found this small module: https://www.drupal.org/project/bca which makes it a matter of adding an annotation instead of implementing a hook.

> write a todo

Some people doesn't believe in todos. :wink: I'm not one of them, but in this case it would probably have poor visibility here.

The thing is, it's not that big of a job jo introduce bundle classes, it's already done for [EventInstance](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/082b34ae99a2de7ef24e92259a7cca88abb7f018/web/modules/custom/dpl_event/dpl_event.module#L29). The hard work is identifying the code that needs to be refactored to the new class in order to actually reap the benefits. For instance, would [`dpl_event_has_free_tickets`](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/082b34ae99a2de7ef24e92259a7cca88abb7f018/web/modules/custom/dpl_event/dpl_event.module#L149C10-L149C36) make sense to move to the bundle class? Well, after some ripgrepping, yes, it's actually only ever called for event instances. But it should be refactored, as the function claims to handle both series and instances.

Which in turn makes me spot [`EventInstance::getState()`](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/082b34ae99a2de7ef24e92259a7cca88abb7f018/web/modules/custom/dpl_event/src/Entity/EventInstance.php#L133), which I now spot is 24 lines that ought to be able to be condensed to:

```
$enums = $this->get('event_state')->enums();
return reset($enums);
```

But who dares to mess with working code?

PR #2349, opened 2025-05-02
Separate goConfiguration graphql into public and private

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-572

#### Description

Separate goConfiguration graphql into public and private to emphasize what is considered to be sensitive and what is not.


Dresse (approved):
LGTM 👍

PR #2368, opened 2025-05-06
Don't rewrite links on admin pages

xendk:
#### Link to issue

DDFBRA-545

#### Description

The PathProcessor messes with the admin experience in two ways:

When re-editing a node with a rewritten link, the link is now the "external" version.

Submitting the node edit page it attempts to redirect to the "external" URL, which causes an error because it doesn't use a TrustedRedirectResponse.

Fix both by disabling rewriting on admin pages.


Dresse (approved):
LGTM 👍

PR #2373, opened 2025-05-07
DDFBRA-594 -  Adjust go content to work with subscriptions

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-594

#### Description

This PR makes sure to add 'field_tags' to all GO specific content. We are currently not using tags on GO content, but we need the field to make subscriptions work (Delingstjenesten). 

Because of this we have 2 scenarios regarding tags:

1. On Delingstjenesten we need to have the field available as this is how content is coupled together to an subscription. This means that the field should be shown on GO content.

2. On local library sites we are not interested in having the tags field visible as it is currently not used as part of the editorial work on GO. Thus we want to hide the field from the editorial view of GO specific content.

To solve the issues above, this PR adds a new permission called 'bnf administer go tags' which is only given to roles that we want to be able to view and edit the tags field on GO content. Furthermore we have a check for this in the node form alter hook. 

To make sure that the above logic is only added on local library sites, we have added it to the 'bnf_client' module, which are only enabled on local library sites. As the module is not enabled on the "Delingstjeneste site", it won't be affected by the logic, and thus the field tags will be visible as normal.

I decided to add a specific permission for handling the visibility of the tags field for several reasons:

1. We talked about maybe only limiting the visibility for editorial roles and not user 1, as it could come in handy to be able to see the field for development usage.
2. Instead of adding a check to see if the current user is user 1, I decided to add the permission, as this gives us the possibility to open up for the visibility for other roles in the future if relevant.

@xendk You suggested that instead of adding the logic in the existing `bnf_client_form_alter` and instead implement it using attribute hooks (used in Drupal 11 and forward). But I did not take this approach as I would have to fiddle with already existing working code and de some refactoring, which would take some extra time that we don't feel we have right now. Thus I decided against it in this PR, but I might take some time on my own time to implement it in a more futureproof way.


xendk (approved):

xendk:
> fiddle with already existing working code and de some refactoring

How do you think the project would look if we only fiddled with non-working code? :wink: 

BTW, I've just done another example in: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2304/commits/dcc85bc60c0b8e451fdcb063e40d2fc50fa8b09e , it should be more transparent.

Anyway, approved.

PR #2374, opened 2025-05-08
Set correct vocabulary for tags field on eventinstances. DDFSAL-111

rasben:
The "field_tags" on eventinstances was set to the categories vocabulary, which must be a mistake.
This commit sets it instead to "tags" vocabulary.
I've tested, and if any existing eventinstances already had data in the field, it is just removed (As expected), without incurring an error.


Dresse (approved):
LGTM 👍

PR #2376, opened 2025-05-08
Include GO nodes in search API indexes.

rasben:
For some reason, one of the two content search API indexes, were set up to only allow some content types.
It appears this was an attempt to get rid of branches and campaigns, but it also meant that it targeted GO content.
Instead, I've reversed the condition, and expliciting stating which content types we do *not* want.
I've also added this condition to the other content index, as it would appear to me that the two should follow the same logic.


config/sync/search_api.index.content.yml
```diff
@@ -117,6 +119,7 @@ tracker_settings:
     indexing_order: fifo
 options:
   cron_limit: 50
+  delete_on_fail: true
```
rasben, line 122: This is a default option, that came just from saving the index. It also exists in the other.

Dresse (approved):
LGTM 👍

PR #2392, opened 2025-05-12
Cache invalidering af dpl cms indhold i go

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-411

#### Description

This PR:
Activates the `drupal/next` module
Configures the module to revalidate cache on the Go site. 
Distinguishes between  development and production settings of the Go site configuration.
Adds an EventSubscriber for returning cache tags for all roles with the permission: `get dpl graphql cache tags`
Adds the permission to the `go_graphql_client`role which makes sure that the cache tags are returned to the go graphql consumer.


web/modules/custom/dpl_go/dpl_go.deploy.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+/**
+ * Make sure that we have properly configured the Next Site.
+ */
+function dpl_go_deploy_0001_create_next_go_site_configuration(): string {
+  if (
+    getenv('LAGOON_ENVIRONMENT_TYPE') === 'production' &&
```
xendk, line 8: Should we collect all the "sniff out lagoon env var" in a `dpl_lagoon` service?

web/modules/custom/dpl_go/dpl_go.deploy.php
```diff
@@ -0,0 +1,55 @@
...
+    'id' => 'go',
+    'label' => 'Go',
+    'base_url' => $base_url,
+    'preview_url' => $preview_url,
+    'preview_secret' => $preview_secret,
+    'revalidate_url' => $revalidate_url,
+    'revalidate_secret' => $revalidate_secret,
+  ];
+
+  // Create the entity programmatically like in dpl_update_john().
```
xendk, line 47: `john`?

web/modules/custom/dpl_graphql/src/EventSubscriber/CacheableResponseSubscriber.php
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @param \Symfony\Component\HttpKernel\Event\ResponseEvent $event
+   *   The event to process.
+   */
+  public function onRespond(ResponseEvent $event): void {
+    if (!$event->isMainRequest()) {
+      return;
+    }
+
+    if (!$this->currentUser->hasPermission('get dpl graphql cache tags')) {
```
xendk, line 44: We ought to find a better way to determine if it's a GraphQL request. Using a permission to enable GraphQL behaviour seems... Wrong.

Perhaps checking the current route if the controller is the `graphql` modules controller?

xendk (approved):
:+1: 

But some points for future consideration.

Dresse (approved):
LGTM 👍

web/modules/custom/dpl_go/dpl_go.deploy.php
spaceo: Hmm. Have to think about that. I'll go with this for now.

web/modules/custom/dpl_go/dpl_go.deploy.php
spaceo: well 🙄

PR #2396, opened 2025-05-13
Implement a cache of node types in the path processor

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-112

#### Description

The consequence of path rewriting is that we have to load each linked node to check its type to determine whether we need to rewrite the URL.

In order to limit the number of entity loads we do, we implement a static cache to keep track of whether a nid is a go node or not. We save this cache to state (as even CACHE_PERMANENT cache items are flushed on cache rebuild) and do it in chunks of a hundred items to keep down the amount of queries when our cache is filled.

### DIFF
--- a/web/modules/custom/dpl_go/src/GoSite.php
+++ b/web/modules/custom/dpl_go/src/GoSite.php
@@ -5,7 +5,10 @@ declare(strict_types=1);
 namespace Drupal\dpl_go;
 
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\State\StateInterface;
 use Drupal\dpl_lagoon\Services\LagoonRouteResolver;
 use function Safe\parse_url;
 
@@ -14,10 +17,26 @@ use function Safe\parse_url;
  */
 class GoSite {
 
+  /**
+   * Node storage.
+   */
+  protected EntityStorageInterface $nodeStorage;
+
+  /**
+   * Static node type cache.
+   *
+   * @var array<int, array<string, bool>>
+   */
+  protected $typeCoche = [];
+
   public function __construct(
     protected LagoonRouteResolver $lagoonRouteResolver,
     protected AccountInterface $currentUser,
-  ) {}
+    EntityTypeManagerInterface $entityTypeManager,
+    protected StateInterface $state,
+  ) {
+    $this->nodeStorage = $entityTypeManager->getStorage('node');
+  }
 
   /**
    * Is the current request considered "the Go site".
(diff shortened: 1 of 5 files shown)


web/modules/custom/dpl_go/src/GoSite.php
```diff
@@ -14,10 +17,26 @@
...
+   * @var array<int, array<string, bool>>
+   */
+  protected $typeCoche = [];
+
   public function __construct(
     protected LagoonRouteResolver $lagoonRouteResolver,
     protected AccountInterface $currentUser,
-  ) {}
+    EntityTypeManagerInterface $entityTypeManager,
+    protected StateInterface $state,
```
kasperg, line 36: Consider using a more generic `KeyValueStore` to handle this.

As I see it the State system in Drupal has more specific use cases than this. It also does a lot more stuff (caching, locking etc.). I cannot tell from the PR whether you considered this or not. On the other had you explicit state that you want to make the data persistent between deploys which might argue for not using the State system?

On the other hand - with caching of state enabled, which will be the case in the future anyway - then having the cached data in APCu or the like might decrease the need/complexity of a static cache.

web/modules/custom/dpl_go/src/GoSite.php
```diff
@@ -89,4 +108,50 @@ public function isGoNode(EntityInterface $node): bool {
...
+    $this->typeCoche[$cache_num][$nid] = NULL;
+
+    $node = $this->nodeStorage->load($nid);
+
+    if ($node) {
+      $this->typeCoche[$cache_num][$nid] = $this->isGoNode($node);
+    }
+
+    // Add in fresh copy of saved state to guard against race conditions.
+    $this->typeCoche[$cache_num] += $this->state->get($state_id, []);
```
kasperg, line 141: Consider if this approach should be reversed e.g.

1. Get the saved state
2. Set the value for the current node in the state

Personally I have always found the behavior of the `+=` operator hard to grok compared to e.g. `array_merge()`.

web/modules/custom/dpl_go/test/src/Unit/GoSiteTest.php
```diff
@@ -131,4 +161,76 @@ public function testIsGoNode(): void {
...
+  }
+
+  /**
+   * Test that isGoNid() recognizes non-go nodes.
+   */
+  public function testIsGoNidNonGoNode(): void {
+    $this->nodeStorage->load('124')->willReturn($this->nodeProphecy('node'));
+
+    $this->state->get('dpl_go.node_type_cache_1', [])->willReturn([]);
+    $this->state->set('dpl_go.node_type_cache_1', ['124' => FALSE])->shouldBeCalled();
```
kasperg, line 183: Consider dropping `shouldBeCalled()`.

The scope of the test is not to check if caching works. This is handled by `testIsGoNidUsesCache()`.

kasperg (approved):
I find it hard to determine if this is premature optimization and if not whether the chosen solution is in fact the right one for the job.

Look into what you think is appropriate and continue at will.

web/modules/custom/dpl_go/src/GoSite.php
xendk: Oh, things have happened to `State` while I wasn't watching. `GoSite` doesn't need to use a static cache at all, it can rely on `State` doing the heavy lifting.

web/modules/custom/dpl_go/src/GoSite.php
xendk: Well, `State` makes it unnecessary.

web/modules/custom/dpl_go/src/GoSite.php
xendk: > On the other had you explicit state that you want to make the data persistent between deploys which might argue for not using the State system?

Whut? The reason for using state is to be persistent?

Anyway...

*sigh* It shouldn't be that simple. `State` inherits from `CacheCollector` which does handy things in general, but...

It caches state in its *entirety*, which kills the idea of chunking our cache. One could argue that our "cache" isn't that big, but chugging it all along on every request that as much as look at a state key feels wrong.

So, trying to go with the Drupal flow, we can use a key value store... But the `CacheCollector` isn't much use, as it's flush on `cr`, and in this case there's no point i rebuilding the cache as node types doesn't change. But then we also loose `CacheCollectors` nice concurrency handling.

So, rather than chunking code and letting `State` do the heavy lifting, we can replace `State` with a key value store (which is annyingly non-autowirable), and keep using the `typeCache` thing.

web/modules/custom/dpl_go/test/src/Unit/GoSiteTest.php
xendk: Well, as there's a prophecy on the object, it'll complain about unexpected calls. One could make a `willReturn` instead, but I think that's more confusing.

PR #2401, opened 2025-05-15
Update contrib modules with recommended* updates.

rasben:
All of these updates are minor (except linkit, that jumps to 7.x), but none of them have updates that changes the config.

I've grouped these together, to ease the testing for DDF.

DDFNEXT-358
DDFNEXT-356
DDFNEXT-375
DDFNEXT-399
DDFNEXT-676


composer.json
```diff
@@ -104,13 +104,13 @@
         "drupal/autowire_plugin_trait": "^1.0",
         "drupal/azure_mailer": "^2.0",
         "drupal/color_field": "^3.0",
-        "drupal/config_filter": "^2.6",
+        "drupal/config_filter": "^2",
```
kasperg, line 107: Please consider not changing the requested version to something looser than what is currently installed. You are effectively saying: I will accept anything that is in the version 2.x range and we seldomly want to go backwards.

As I see it there is nothing in the current version specification that would prevent us from going to `config_filter` 2.7.0 with the current specification.

rasben:
>The title indicates security updates. I was not aware of any. Out of interest: Which of them are? This might indicate a problem in our update handling.

Screenshot 2025-05-16 at 08 27 21

The ones that are marked with red, are recommended. I understand it as a security update, although maybe that's me being daft.
Regardless, those are the ones I've prioritized

rasben:
> In general we do not send module updates to be tested by the business.

Good point, I keep forgetting this on this project.

> If you want to do something then I suggest investing some time into implementing a Cypress test which shows that the functionality provided by the module still works for us. This might be relevant for modules such as Content Lock or Flat Taxonomy.

Well, I suppose the existing cypress tests, that run through creating content etc. works well enough :)

rasben:
P.S. I skipped webform and `jsnlog` as they were more complicated, and I'd prefer to take them in a seperate PR

kasperg:
> The ones that are marked with red, are recommended. I understand it as a security update, although maybe that's me being daft.

I think that is misunderstood. Checking https://www.drupal.org/project/config_filter shows that the update to 2.7 does not address a security issue. 

> I suppose the existing cypress tests, that run through creating content etc. works well enough

My point is are no tests that check that:

- Content is in fact locked by the Content Lock module - or that our administration view shows whether content is locked or not
- Ensure that taxonomies that we want to be flat behave that way

rasben:
Isnt it a bit extreme to test if a module that does one thing, actually does the thing?
I understand that we have tests that check if the workflow for the editors still work, after things are 'welded' together, but this seems a bit too much (?)

config/sync/views.view.content.yml
```diff
@@ -277,7 +277,7 @@ display:
         options:
           perm: 'access content overview'
       cache:
-        type: tag
+        type: none
```
kasperg, line 280: I wonder why this popped up. It seems to be related to this issue: https://www.drupal.org/project/content_lock/issues/3381841

In effect this will reduce caching on the content overview page and could affect the editor experience by reducing performance. As far as I can tell our editors have not complained about this problem so I wonder if it something that we actually want to do.

Consider discussing it with the business. On the other hand it might be hard to actively remove this behavior from the module which would leave us to keep it as is.

kasperg:
> Isnt it a bit extreme to test if a module that does one thing, actually does the thing?
I understand that we have tests that check if the workflow for the editors still work, after things are 'welded' together, but this seems a bit too much (?)

I understand your argument about the effort required but I still think it has value.

Things are welded together in different ways by us e.g. showing the content lock state in the content view. Cypress tests might also help us identify non-obvious compatibility problems between related modules, Drupal Core versions, PHP versions etc.

The experience we get from working with this should enable to implement tests of other parts of the system in the future.

config/sync/views.view.content.yml
rasben: >Consider discussing it with the business. On the other hand it might be hard to actively remove this behavior from the module which would leave us to keep it as is.

Well, if I manually remove it, won't it just get overwritten on next deploy?
E.g. content_lock might set it to `none`, but after we run `config-import`, it'll get set back to `tag`

rasben: Okay, I understand what you mean now - the test fails if i do it like that, as the sites are remade from scratch 😅

Is it worth it to run it by the business then? Do we have any choice?

kasperg: > Do we have any choice?

Good question. If we want to reverse it we have to either introduce some hackery to remove the `deploy` hook for the module, use a reverse patch to remove it or the like. It will not be pretty but we could make it work.

rasben: Fixed, as we talked about:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2401/commits/cdd410b794a129a6244819bb6032ec56e6fc7f76

rasben: and, i'm not too worried about the cache tag actually being an issue, especially not with the cypress test :)

rasben: Fixed upstream: https://www.drupal.org/project/content_lock/issues/3525430

cypress/e2e/module-tests.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+      .as('firstContent');
+    cy.get('@firstContent').should('not.include.text', 'Locked by');
+  });
+
+  it('flat_taxonomy', () => {
+    cy.drupalLogin(
+      '/admin/structure/taxonomy/manage/tags?destination=/admin/structure/taxonomy/manage/tags/overview',
+    );
+    cy.get('[data-drupal-selector="edit-flat"]').check();
+    cy.get('[data-drupal-selector="edit-submit"]').click();
```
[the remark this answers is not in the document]
rasben: I did consider it, but, I'd rather not have a dependency on the actual business logic, as we're testing module logic here.

cypress/e2e/module-tests.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+    );
+    cy.get('[value="Apply to selected items"]').click({ force: true });
+
+    cy.get('.views-table tbody tr')
+      .should('include.text', 'Edit')
+      .first()
+      .as('firstContent');
+    cy.get('@firstContent').should('not.include.text', 'Locked by');
+    // We need to force it, as Cypress gets confused about the floating header.
+    cy.get('@firstContent').contains('Edit').click({ force: true });
```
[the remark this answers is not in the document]
rasben: As we talked about, I believe it's cause we here need to click on the first item in the list

cypress/e2e/module-tests.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+  it('flat_taxonomy', () => {
+    cy.drupalLogin(
+      '/admin/structure/taxonomy/manage/tags?destination=/admin/structure/taxonomy/manage/tags/overview',
+    );
+    cy.get('[data-drupal-selector="edit-flat"]').check();
+    cy.get('[data-drupal-selector="edit-submit"]').click();
+    cy.get('.tabledrag-handle[title="Move in any direction"]').should(
+      'not.exist',
+    );
+    cy.get('.tabledrag-handle[title="Change order"]').should('exist');
```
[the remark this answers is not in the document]
rasben: I have the class, to make the test clearer that it is the drag-handle we're looking at

cypress/e2e/module-tests.cy.ts
```diff
@@ -0,0 +1,62 @@
+describe('Testing contrib modules', () => {
+  it('content_lock', () => {
+    cy.drupalLogin('/admin/content');
+
+    // Making sure all content is unlocked initially.
+    cy.get('[title="Select all rows in this table"]').check();
```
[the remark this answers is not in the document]
rasben: I did try with findByTitle() but it didnt work. I just re-tried, and it still does not work here.
I'll just move forward with this ^ :)

cypress/support/commands.ts
```diff
@@ -82,12 +82,34 @@ Cypress.Commands.add('drupalLogin', (url?: string) => {
...
   cy.session({ username, password }, () => {
     cy.visit('/user/login');
+
+    // If the CookieInformation prompt is here, we want to click it, to not
+    // have it block the user information.
+    cy.get('body').then(($body) => {
+      if ($body.find('.coi-banner__accept').length > 0) {
+        cy.get('.coi-banner__accept').first().click();
+      }
+    });
```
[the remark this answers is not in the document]
rasben: I think I've spent enough time on this, and it works like this, so I'll leave it as is

PR #2431, opened 2025-05-23
Fix unit tests for BNF

rasben:
## Remove unsupported tests.

These unit tests were merged as part of a faulty pull request - they do
not work against `develop` (and never has, against this develop), and it
is not a simple quickfix to get them to work
So, instead, I remove them, so we can re-visit them in the future

## Fix unit tests for BNF node tests.


rasben:
This is the pull request i merged yesterday:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2295

7/11 tests are still valid, but the remaining 4 are getting removed here.

PR #2443, opened 2025-05-27
DDFBRA-657 - Make search profile and FB-CMS URL available through GraphQL

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-657

#### Description

This PR make sure that the GO Frontend can now fetch the following configuration from FB-CMS backend:

Search Profiles: - VIP profiler
- Default - Standardprofil
- Local - Søgeprofil
- Global - Visningsprofil

The PR also adds the `cmsUrl` to the LibraryInfo type. This field fetches the absolute CMS Base URL of FB-CMS of the library.

#### Screenshot of the result

Example GraphQL query:

```graphql
query MyQuery {
  goConfiguration {
    public {
      searchProfiles {
        default
        local
        global
      }
      libraryInfo {
        name
        cmsUrl
      }
    }
  }
}
```

Example result:
```
{
  "data": {
    "goConfiguration": {
      "public": {
        "searchProfiles": {
          "default": "next",
          "local": "next",
          "global": "next"
        },
        "libraryInfo": {
          "name": "Eksempel Biblioteket",
          "cmsUrl": "https://dapple-cms.docker"
        }
      }
    }
  }
}
```

This PR will have a sister PR in dpl-go:
<add PR link when available>

### DIFF
--- a/web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
+++ b/web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
@@ -11,6 +11,12 @@ type GoLogoutUrls {
   adgangsplatformen: String
 }
 
+type SearchProfiles {
+  defaultProfile: String
+  searchProfile: String
+  materialProfile: String
+}
+
 type UniloginConfigurationPrivate {
   clientSecret: String
   webServiceUsername: String
@@ -30,6 +36,7 @@ type GoConfigurationPublic {
   libraryInfo: GoLibraryInfo
   loginUrls: GoLoginUrls
   logoutUrls: GoLogoutUrls
+  searchProfiles: SearchProfiles
   unilogin: UniloginConfigurationPublic
 }
 
--- a/web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
+++ b/web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
@@ -13,9 +13,9 @@ type GoLogoutUrls {
 }
 
 type SearchProfiles {
-  defaultProfile: String
-  searchProfile: String
-  materialProfile: String
+  default: String
+  local: String
+  global: String
 }
 
 type UniloginConfigurationPrivate {
--- a/web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
+++ b/web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
@@ -12,9 +12,22 @@ type GoLogoutUrls {
   adgangsplatformen: String
 }
 
+"""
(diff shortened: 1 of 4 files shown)


web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
```diff
@@ -11,6 +12,12 @@ type GoLogoutUrls {
   adgangsplatformen: String
 }
 
+type SearchProfiles {
+  defaultProfile: String
+  searchProfile: String
+  materialProfile: String
```
spaceo, line 18: This is super minor, and only a consideration:
Since the object is called: SearchProfiles, I think it is redundant to call  the profiles: *Profile.
So instead: 
```
type SearchProfiles {
  default: String
  search: String
  material: String
}

File: web/modules/custom/dpl_go/src/Plugin/GraphQL/DataProducer/CmsUrlProducer.php [hunk 2]
```diff
@@ -0,0 +1,58 @@
...
+      $plugin_definition,
+      $container->get('dpl_go.go_site'),
+    );
+  }
+
+  /**
+   * Resolves the library name.
+   */
+  public function resolve(FieldContext $field_context): string {
+    $field_context->addCacheableDependency((new CacheableMetadata())->setCacheMaxAge(0));
```
spaceo, line 54: Please write a todo that this should be cacheable

web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
```diff
@@ -1,6 +1,7 @@
 
 type GoLibraryInfo {
   name: String
+  cmsUrl: String
```
spaceo, line 4: I think this should be non-nullable. Because the site always knows its url. Right?
Would be nice if have at least some properties that differ from the other optionals. 😅

spaceo (approved):
LGTM 👍 
But please look at comments

web/modules/custom/dpl_go/graphql/dpl_go.base.graphqls
Dresse: Fixed. Changed the names to match the FB-CMS (local and global)

web/modules/custom/dpl_go/src/Plugin/GraphQL/DataProducer/CmsUrlProducer.php
Dresse: Done 👍

PR #2455, opened 2025-06-02
DDFNEXT-335 - Update `vincentlanglet/twig-cs-fixer` to version `2.11.0`

kasperbirch1:
#### Link to issue

https://reload.atlassian.net/browse/DDFNEXT-335

#### Description


kasperg (changes requested):
This is comprehensive work!

I would appreciate it if the choice of version was actually the latest version. Otherwise we risk ending up with a new update-ticket immediately after solving this.

PR #2456, opened 2025-06-02
DDFBRA-653 - Update passwords for 'go_graphql' and 'bnf_graphql' users in update hook.

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-653

#### Description

This PR adds an update hook for updating the passwords for the `go_graphql` and `bnf_graphql` users based on what is stored in the ENV variables.


spaceo (commented):
The version I have reviewed does not have the `DRUPAL_REVALIDATE_SECRET` yet. But you said it is on its way, right?
Otherwise LGTM 👍 
Please look at my suggestion, if it makes sense to you.

PR #2464, opened 2025-06-03
DDFBRA-569 - Delete "Go Editor" role and move permissions to other editor roles

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-569

#### Description

The Go Editor role was always meant to be used temporary while we were developing the GO site. The role is now deprecated as we are preparing to launch GO for all libraries. 

This PR make sure to add the required GO specific permissions to the other editor roles. The PR also removes the "GO Editor" from the configuration as it is no longer needed.


spaceo (approved):
One question
otherwise...
LGTM 👍

PR #2476, opened 2025-06-05
Webmaster cypress test

xendk:
Add cypress test of webmaster module upload.

Introducing page objects.


cypress/support/e2e.ts
```diff
@@ -20,3 +20,6 @@ import './commands';
 // require('./commands')
 
 import 'cypress-plugin-api';
+
+import installLogsCollector from 'cypress-terminal-report/src/installLogsCollector';
+installLogsCollector();
```
JacobArrow, line 25: Would be nice to know why this plugin is used

cypress/e2e/webmaster-upload.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+import { AdminModulesPage } from '../pages/admin-modules';
+import { AdminModulesUninstallPage } from '../pages/admin-modules-uninstall';
+import { InstallOrUpdatePage } from '../pages/install-or-update';
+import { TestModulePage } from '../pages/test-module-page';
+
+describe('Webmaster', () => {
+  beforeEach(() => {
+    // Obviously we need to be logged in to upload a module.
+    const loginPage = new LoginPage();
+    loginPage.visit([]);
```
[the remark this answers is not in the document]
xendk: The URL defined in the PO supports placeholders, so `visit()` takes values to apply to the placeholders.

Actually, it annoyed me that it was required, so if we end up creating our own page object base class, I would override the `visit()` method to make the array optional.

cypress.config.js
```diff
@@ -8,6 +8,9 @@ module.exports = defineConfig({
       runMode: 3,
       openMode: 0,
     },
+    setupNodeEvents(on, config) {
+      require('cypress-terminal-report/src/installLogsPrinter')(on)
```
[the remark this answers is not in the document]
xendk: Yes it is. It needs both a collector and a formatter.

cypress/pages/login-page.ts
kasperg, line 1: Please consider the relevance of this page object in comparison with [the existing `drupalLogin()` command](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/6b5e1f135013a6b6006268f56b7181b5a172746b/cypress/support/commands.ts#L80-L80). 

It uses [`cy.session()`](https://docs.cypress.io/api/commands/session) to support [switching between users in the same test](https://github.com/danskernesdigitalebibliotek/dpl-cms/commit/ea442ee5580ba2bda9b7345f39cee3f254330972).

In the end I do not think both should exist. I am partial to keeping the command for the support for `cy.session()`, brevity and reducing the impact on the codebase.

cypress/pages/install-or-update.ts
```diff
@@ -0,0 +1,59 @@
...
+  uploadModule(fixture: string) {
+    this.elements.fileField().selectFile(fixture);
+    this.elements.submit().click();
+
+    // Wait for the batch job to complete and figure out if we're
+    // being offered to run database updates and do it if we are
+    // (which means this was an module update).
+    cy.get('.authorize-results')
+      // While .authorize-results signals we're done, it doesn't contain
+      // the link we need.
```
kasperg, line 24: I appreciate this level of commenting.

kasperg (approved):
:+1: I have looked through this. I think it is good.

A couple of thoughts:

1. The login handling leaves the project with two approaches to solving the same issue. We need to determine what should be the appropriate way going forward.
2. I think it would be appropriate to write an ADR for this since it is a pattern we expect to follow going forward.
3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3. 3.

cypress/pages/login-page.ts
xendk: > It uses [cy.session()](https://docs.cypress.io/api/commands/session) to support [switching between users in the same test](https://github.com/danskernesdigitalebibliotek/dpl-cms/commit/ea442ee5580ba2bda9b7345f39cee3f254330972).

Well that's easy to overlook when there's no function comment and the name implies that it does a Drupal login.

`cy.session()` doesn't quite map up to a page as such (after the first run it explicitly doesn't visit the login page), but I don't think it's entirely incompatible.

We could add a `LoginPage.ensureLogin()` (sidenote: `LoginPage` should probably be `UserLoginPage` to follow the convension) static method which uses `cy.session()` and an instance of itself within the setup method. Also makes it more obvious that we're not necessarily logging in.

cypress/fixtures/test_module/v1.0.0/test_module.tar.gz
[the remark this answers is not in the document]
xendk: I did a lot of un/tarring when creating them, but I don't think there'll be much future changes to them.

PR #2484, opened 2025-06-06
DDFSAL-161: Common material grid paragraph

rasben:
Review commit by commit, as each commit describes exactly what is happening :) 

Here's a how-to-test video:

https://github.com/user-attachments/assets/71f63f65-e2be-4f1c-a7b8-42fcdbe495fc

---

For the filters to actually work, the "MaterialGridAutomatic" component in dpl-react needs to be updated, to take `sort`, `onshelf`, `location` and `sublocation` as optional parameters - something, that I know @kasperbirch1 has also been looking partly into.
I'd like some help with that :)


web/themes/custom/novel/templates/paragraphs/paragraph--material-grid-automatic.html.twig
```diff
@@ -1 +0,0 @@
-{{ react_app_material_grid_automatic }}
```
rasben, line 1: This file (and the others below) were never needed - we just needed to set `$variables['content']['whatever']`, and Drupal would take care of the rest.

PR #2486, opened 2025-06-08
Do not use Go node service in PR environments

spaceo:
Do not use Go node service in PR environments
Except from demo and playground which needs to have the service defined in docker-compose.yml


Adamik10 (approved):
LGTM ✨

PR #2490, opened 2025-06-10
Update Drupal core 10.3.13 => 10.4.8. DDFSAL-151

rasben:
Update Drupal core 10.3.13 => 10.4.8. DDFSAL-151

Updating to the latest stable version of 10.4.x.
This also meant that several patches has been merged into core, and could be removed.

The `cms-api` package could not apply, as Composer kept getting confused, wanting to lock it to 6.0, so I locked it to 6.4 in the base composer.lock.

Apart from that, I also had to add a patch to get LegacyHook attributes to work again.
I had to add it as a patch, rather than as a link, because their patch has the wrong paths, meaning it placed the files in `core/core/lib` instead of `core/lib`.

Also includes two minor fixes that came from `PHPCBF`, after updating.

## IMPORTANT 

This update cannot be merged before DDFSAL-159 has been merged, as 10.4.x removes module upload.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -112,8 +112,8 @@
         "drupal/consumers": "^1.19",
         "drupal/content_lock": "^2.4",
         "drupal/cookieinformation": "^2.1",
-        "drupal/core-composer-scaffold": "10.3.13",
-        "drupal/core-recommended": "10.3.13",
+        "drupal/core-composer-scaffold": "10.4.8",
+        "drupal/core-recommended": "10.4.8",
         "drupal/date_range_formatter": "^4.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -200,7 +200,7 @@
         "zaporylie/composer-drupal-optimizations": "1.2.0"
     },
     "require-dev": {
-        "drupal/core-dev": "~10.1.7",
+        "drupal/core-dev": "^10.1",
```
rasben, line 203: We talked about removing this package, but, when I did, all the PHPUnits started to fall apart.

xendk (approved):

PR #2530, opened 2025-06-17
BNF: Do *NOT* set BNF-imported nodes as entity templates. DDFSAL-174

rasben:
When a node is BNF-imported, we do *NOT* want it to show up as a node template on the /node/add page. Apparantly, the entity_clone_template module assumes that no value must mean TRUE.

This commit also includes cleaning up the existing nodes that already have been imported, and has been set as templates. The client will be informed of this update.

https://reload.atlassian.net/browse/DDFSAL-174

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -334,7 +334,8 @@
                 "2708731: Allow redirect to edit form after cloning": "https://www.drupal.org/files/issues/2022-11-15/improve-workflow-of-cloned-entity-2708731-39.patch"
             },
             "drupal/entity_clone_template": {
-                "3424597: Fix error when creating entity of type without templates": "https://git.drupalcode.org/project/entity_clone_template/-/commit/ad4964dd51f7f5e814111fcb73e6e472eefccf5c.patch"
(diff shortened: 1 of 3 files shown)


xendk:
> entity_clone_template module assumes that no value must mean TRUE.

That's the most stupid thing.

What's the reason for not fixing it upstream? Seems to me that it's just the view that needs to be fixed.

xendk:
> You really want a patch solution instead?

As I see it, with a patch we avoid having to deal with the problem in the mapper and don't have to bulk update all the nodes, in essence we make it a non-issue for the DPL codebase.

> The view? but, the data is literally being set as TRUE if nothing is set.

Ah, yeah, that's the stupidest default value. That probably means we'll still need the update hook, as it can't be fixed upstream (they'll probably not like an update hook that disables all templates).

rasben:
Alright, I've created an issue and upstream MR here:

https://git.drupalcode.org/project/entity_clone_template/-/merge_requests/4/diffs?commit_id=7cd2ee6541cd25b6370b304911d49a00c46c3a0f

xendk (approved):

PR #2541, opened 2025-06-20
Strip `www` prefix from cookie_domain

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-725

#### Description

On sites whose primary domain starts with `www`, Drupal sets the cookie domain to `.www.<site>` per default. This means that the cookie isn't shared with `www.go.<site>` and the login doesn't work. So set it explicitly without the `.www` prefix.

PR set to "don't merge" as it should be applied as a hotfix (which involves creating a new branch first).

### DIFF
--- a/web/modules/custom/dpl_go/src/DplGoServiceProvider.php
+++ b/web/modules/custom/dpl_go/src/DplGoServiceProvider.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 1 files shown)


spaceo (approved):
LGTM 👍 
BUT! Please be aware that a Cypress test is failing....

xendk:
Has been merged as hotfix.

PR #2546, opened 2025-06-23
Strip `www` prefix from cookie_domain

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-727

#### Description

On sites whose primary domain starts with `www`, Drupal sets the cookie domain to `.www.<site>` per default. This means that the cookie isn't shared with `www.go.<site>` and the login doesn't work. So we set it explicitly without the `.www` prefix.

### DIFF
--- a/web/modules/custom/dpl_go/src/DplGoServiceProvider.php
+++ b/web/modules/custom/dpl_go/src/DplGoServiceProvider.php
@@ -37,16 +37,23 @@ class DplGoServiceProvider extends ServiceProviderBase {
     }
     else {
       // We can't use the service from `dpl_lagoon` for this, as obviously we
-      // can't use the container.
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_go/src/DplGoServiceProvider.php
```diff
@@ -0,0 +1,71 @@
...
+      // can't use the container while building it.
+      $route = getenv('LAGOON_ROUTE');
+
+      if (!$route) {
+        // Without a route, we can't do much.
+        return;
+      }
+
+      // Split off the scheme.
+      $parts = explode('//', $route, 2);
```
spaceo, line 49: Well done!
I think using parse_url would be more elegant here.
Then it should be possible to write:
```php
['host' => $host] = parse_url($route);

PR #2585, opened 2025-07-01
Docs update

xendk:
Guide to BNF debugging. And a few cleanups.


kasperg (approved):
:+1: Nice.

Personally I would like this to be more concrete in steps to take, log output to look for etc. but I will leave it to you to consider if there is more to do. Getting this in is an improvement in any form.

PR #2594, opened 2025-07-03
BNF: Manual-reference paragraphs. DDFSAL-189

rasben:
Review commit by commit.
This includes some re-factoring, to share logic between the old link-references and new entity-references.


rasben:
NOTICE:
I've added "[Remove comments from generated bnf.graphql file.](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2594/commits/acb21ce3d992e4976688ff8b74ba57e6516d96a1)" to this pull request, to make my life a bit easier.

However, this has it's own pull request here:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2596

I won't merge this pull request with that change, until ^ pull request has been approved somehow :)

web/modules/custom/bnf/graphql/bnf_extension.extension.graphqls
```diff
@@ -40,3 +40,15 @@ extend type NodeGoPage {
 extend type Link {
   id: String
 }
+
+extend type ParagraphNavGridManual {
```
xendk, line 44: Rather than extending each paragraph type that has a reference field, extend the underlying field type.

See
https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/f5c18fe98346501bc4459e2de05e4e61fd9970c8/web/modules/custom/bnf/graphql/bnf_extension.extension.graphqls#L30

and

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2402/commits/adba1b6b717a0d0c8a0504a8eb2913f209b3b8fa

xendk (changes requested):

web/modules/custom/bnf/graphql/bnf_extension.extension.graphqls
rasben: But, the underlying fieldtype more or less means the same thing - or am i misunderstading something?

Screenshot 2025-07-04 at 13 46 21

PR #2624, opened 2025-07-16
"Full day" events. DDFSAL-255

rasben:
Adds support for “all day” events (DDFSAL-255), including backend field, frontend display, override warning, and Cypress test.
--
Review commit-by-commit - it'll give you a good idea of what this pull request is acually about :)

### DIFF
--- a/cypress/e2e/events.cy.ts
+++ b/cypress/e2e/events.cy.ts
@@ -69,6 +69,34 @@ describe('Events', () => {
       );
   });
 
+  it('all-day event is respected', () => {
+    // Login as admin.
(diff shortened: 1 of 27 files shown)


cypress/e2e/events.cy.ts
```diff
@@ -69,6 +69,34 @@ describe('Events', () => {
       );
   });
 
+  it('all-day event is respected', () => {
```
kasperbirch1, line 72: Nice with a test 🙏

kasperbirch1 (approved):
LGTM 👍

PR #2625, opened 2025-07-18
Add option to hide events from automatic lists. DDFSAL-257

rasben:
Add a boolean field on `EventSeries`, that when checked, makes sure that the eventseries and it's instances do not show up in our various automatic lists.


kasperbirch1 (approved):
LGTM 👍

PR #2636, opened 2025-07-22
BNF: Update to Drupal 10.5. DDFSAL-260

rasben:
This also includes an update to drupal/editor_advanced_link, as the current version only works with >=Drupal 10.4

```
composer require drupal/core-composer-scaffold:^10 drupal/core-recommended:^10 drupal/core-recommended:^10 drupal/core:^10 drupal/editor_advanced_link:^2

composer update "drupal/*" -W

drush updb -y

drush cex -y
```


spaceo (approved):
LGTM 👍

PR #2637, opened 2025-07-24
Update drush to 13, & prepare for Drupal 11 DDFSAL-153

rasben:
There's a problem with PHPstan, that I cannot get my head around:

```
------ ------------------------------------------------------------------------ 
  Line   modules/custom/collation_fixer/src/Commands/CollationFixerCommands.php  
 ------ ------------------------------------------------------------------------ 
  39     Call to method success() on an unknown class Drush\Style\DrushStyle.    
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols     
  42     Call to method info() on an unknown class Drush\Style\DrushStyle.       
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols     
 ------ ------------------------------------------------------------------------ 

 ------ ---------------------------------------------------------------------- 
  Line   modules/custom/dpl_po/src/Commands/DplPoCommands.php                  
 ------ ---------------------------------------------------------------------- 
  81     Call to method error() on an unknown class Drush\Style\DrushStyle.    
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols   
  89     Call to method success() on an unknown class Drush\Style\DrushStyle.  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols   
  140    Call to method error() on an unknown class Drush\Style\DrushStyle.
 ```

It's weird, because the class does seem to show up correctly in PHPStorm.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -86,3 +86,6 @@ parameters:
     - web/modules/contrib/config_translation_po/config_translation_po.bulk.inc
     # Drupal\dpl_webmaster\Form\InstallOrUpdateModule uses this.
     - web/core/modules/update/update.authorize.inc
+
+  scanDirectories:
+    - vendor/drush/drush/src-symfony-compatibility


kasperg:
As I see it the reason why this fails is because Drush 13 introduced some complexity in their own autoloading process to manage compatibility with different Symfony versions. [Consequently `Drush\Style\DrushStyle` is not located in the directory where it normally would be according to PSR4](https://github.com/drush-ops/drush/tree/13.x/src-symfony-compatibility/v6/Style).

To address this Drush 13 itself provides a [`phpstan-bootstrap.php`](https://github.com/drush-ops/drush/blob/13.x/phpstan-bootstrap.php) file. [This allows PHPStan to load these classes](https://phpstan.org/config-reference#bootstrap), which I expect to use in their own development process.

I have tried to determine if are required to load this file ourselves or if PHPStan will autodiscover such files for all dependencies. This does not seem to be in case. As such I suggest that we reference the Drush `phpstan-bootstrap.php` file in [our own PHPStan config](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/phpstan.neon.dist).

rasben:
@kasperg Did I understand you correctly that it should be added like this?:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2637/commits/122579afff838201fe2a86dafc25e6bd475fdf5d

Because, it doesnt seem to have any affect - the error persists.

PR #2656, opened 2025-08-05
Streamline date formats. DDFSAL-106

rasben:
DDF has requested that all dates are formatted as j. F Y
- e.g.: "3. marts 2028"

### DIFF
--- a/config/sync/core.entity_view_display.eventinstance.default.default.yml
+++ b/config/sync/core.entity_view_display.eventinstance.default.default.yml
@@ -49,10 +49,10 @@ content:
       timezone_override: ''
       date_format: 'Y-m-d\TH:i:s'
       separator: '-'
-      one_day: 'd F Y'
-      one_month: 'd - {d} F Y'
-      several_months: 'd F - {d} {F} Y'
-      several_years: 'd F Y - {d} {F} {Y}'
+      one_day: 'j. F Y'
+      one_month: 'j. - {j}. F Y'
+      several_months: 'j. F - {j}. {F} Y'
+      several_years: 'j. F Y - {j}. {F} {Y}'
     third_party_settings: {  }
     weight: 1
     region: content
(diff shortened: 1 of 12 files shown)


config/sync/core.entity_view_display.eventinstance.default.default.yml
```diff
@@ -49,10 +49,10 @@ content:
       timezone_override: ''
       date_format: 'Y-m-d\TH:i:s'
       separator: '-'
-      one_day: 'd F Y'
-      one_month: 'd - {d} F Y'
-      several_months: 'd F - {d} {F} Y'
-      several_years: 'd F Y - {d} {F} {Y}'
+      one_day: 'j. F Y'
+      one_month: 'j. - {j}. F Y'
+      several_months: 'j. F - {j}. {F} Y'
```
kasperg, line 54: Please consider if there is a misplaced `.` here.

Should it be `several_months: 'j. F - {j.} {F} Y'`?

kasperg (changes requested):
Please consider if there are misplaced . throughout this PR.

Should it be 'j. F - {j.} {F} Y' instead of 'j. F - {j}. {F} Y'?

config/sync/core.entity_view_display.eventinstance.default.default.yml
rasben: Nope - it's correct.
`{j}` just means the end_date format.

I get the confusion though, but if I do `{j.}` it results in:

Screenshot 2025-08-11 at 11 11 41

kasperg: OK. I think I got confused because some locations had `{j.}` and others had `{j}.`.

In any case it looks as if we use `{j}.` consistently now. That's good.

kasperg (approved):

PR #2659, opened 2025-08-05
Ddfsal 207 drupal fbi

xendk:
#### Link to issue

This is initial work on [DDFSAL-207](https://reload.atlassian.net/browse/DDFSAL-207). It only adds the work title to the title tag, but implements all the heavy "GraphQL to get the data" stuff.

This doesn't close DDFSAL-207, and haven't been approval tested by DPL yet.

#### Description

Adds a Drush command to run the Sailor GraphQL client generator. Apart from the ergonomics of being a Drush command, it also bootstraps enough Drupal in order to be able to use Drupal services in the Sailor configuration file. This means that Sailor can generate a client for a service URL that's configured in Drupal.

Add a hook to `DplReactAppsController::externalApiBaseUrls()` so that other modules can provide service URLs. `externalApiBaseUrls()` shouldn't be mucking about in other modules config, and letting other modules provide the URLs means that `dpl_react_apps` can be decoupled from the specific services.

Move the FBI URL configuration to `dpl_fbi` and use the above-mentioned hook. It's where it belongs. Also move `FbiProfileType` from `dpl_library_agency` to `dpl_fbi` for the same reason.

The configuration of FBI profiles is still in `dpl_library_agency`. It ought to be moved to `dpl_fbi`, but whether the latter should have it's own configuration page, or hook into the one in `dpl_library_agency` is a UX matter. And before mucking with the UX we should take a hard look at the existing configuration pages from a user view-point.

Adds a GraphQL query to fetch a work title, and generate a client.

Adds a FBI service that can provide information about the FBI service, and a method to get the title of a work.

Finally add a title callback to the work page which fetches the title of the work.

#### Cache considerations

Cache was considered when planning this feature, lazy rendering was mentioned and a lot of things i can't remember.

But reconsidering, this change doesn't change much.

Varnish caches pages per URL, so there's not any change from it's viewpoint. We do seem to get `X-Varnish-Cache: MISS` on work pages, so there's something to look into, but it's also the case for the existing code, so it's out of scope for this PR.

Drupal needs to build every work page as the HTML of the React apps changes (at minimum the `data-wid` attribute containing the work id), so adding in the title isn't much extra work. And on reloads `X-Drupal-Dynamic-Cache: HIT` tells us that it is cached with a `X-Drupal-Cache-Max-Age: -1 (Permanent)'. So that's good.


web/modules/custom/dpl_fbi/src/Fbi.php
```diff
@@ -0,0 +1,110 @@
...
+
+    $urls = [];
+    // Create an URL for each profile.
+    if ($baseUrl) {
+      foreach ($this->getProfiles() as $type => $profile) {
+        // The default FBI service has its own key with no suffix.
+        $service_key = $type === FbiProfileType::DEFAULT->value ? 'fbi' : sprintf('fbi-%s', $type);
+
+        // Create a service url with the profile embedded.
+        $base_url = preg_replace('/\[profile\]/', $profile, $baseUrl);
```
kasperg, line 75: Could/should this be replaced by `getServiceUrl(string $type)`?

web/modules/custom/dpl_react_apps/src/Controller/DplReactAppsController.php
```diff
@@ -543,41 +536,31 @@ public function work(string $wid): array {
...
+    $configuredServices = $react_apps_settings->get('services') ?? [];
+    $services = \Drupal::moduleHandler()->invokeAll('dpl_react_apps_api_urls');
+    foreach ($configuredServices as $name => $conf) {
+      // @todo remove when the FBI URL configuration is moved to dpl_fbi.
+      if ($name == 'fbi') {
+        continue;
       }
+
+      $services[$name] = $conf['base_url'];
     }
```
kasperg, line 555: Please consider if this code is necessary.

As I see it FBI url configuration has now been moved to `dpl_fpi`.

drush/Commands/sailor/SailorCommands.php
```diff
@@ -0,0 +1,72 @@
...
+
+    $this->chDirToModule($module);
+    $introspect->run($input, $this->output());
+  }
+
+  /**
+   * Change to the dir of the module.
+   */
+  protected function chDirToModule(string $module): void {
+    chdir($this->moduleHandler->getModule($module)->getPath());
```
kasperg, line 69: Consider if it would be a good idea to revert to the current directory after invoking the Sailor command.

Changing the directory seems to occur globally for the running script. Unfortunately it seems as if Sailor/Symfony does not provide a way to set this on its own.

kasperg (changes requested):
:+1: Lots of changes here.

I appreciate the restructuring effort you put into using Sailor for different services and modules. I think this worked out well. Please remember how we treat the output of this on GitHub.

I also have some thought for you to consider.

I appreciate your thoughts in the PR description. I would have appreciate it even more as a part of the commit comments. They are *very* brief in this PR.

PR #2663, opened 2025-08-05
Add test of work title

xendk:
#### Description

Add a test of #2659. Doesn't work, seems that PHP doesn't use the wiremock.

Needs to be fixed.

### DIFF
--- a/cypress/e2e/withMappings/work.cy.ts
+++ b/cypress/e2e/withMappings/work.cy.ts
@@ -1,0 +1,11 @@
+import { WorkPage } from '../../pages/work-page';
+
+describe("Work page", () => {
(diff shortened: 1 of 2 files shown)


web/modules/custom/dpl_fbi/src/GraphQL/Operations/WorkInfo/Work/Titles/WorkTitles.php
```diff
@@ -0,0 +1,46 @@
+<?php declare(strict_types=1);
```
rasben: Why is this file (and all the others below) not being marked as code-generated here in Github?

PR #2634, opened 2025-07-22
Contrib updates - July 2025

rasben:
## Update all available contrib modules, for D10.4.

Full list of the updates:

```
drupal/cookieinformation       2.1.3     => 2.2.1
drupal/devel                   5.3.1     => 5.4.0
drupal/editor_advanced_link    2.2.7     => 2.2.8
drupal/graphql_compose         2.3.0     => 2.4.0
drupal/graphql_compose_preview 1.1.0     => 1.1.1
drupal/linkit                  7.0.5     => 7.0.7
drupal/upgrade_status          4.3.7     => 4.3.8

Transitive dependencies:
drupal/better_exposed_filters  6.0.6     => 7.0.5
drupal/coder                   8.3.26    => 8.3.30
drupal/core                    10.4.8    => 11.2.2
drupal/ctools                  4.0.4     => 4.1.0
drupal/entity_clone            2.0.0-beta4 => 2.1.0-beta1
drupal/graphql_compose_routes  2.3.0     => 2.4.0
drupal/jquery_ui_datepicker    2.1.0     => 2.1.1
drupal/paragraphs_features     2.1.0     => 2.2.0
drupal/schemata                1.0.0-beta3       => 1.0.0
drupal/schemata_json_schema    1.0.0-beta3       => 1.0.0
drupal/select2                 1.15.0    => 2.0.0
```

## Fix codestyle, after linters have been updated.

PR #2679, opened 2025-08-12
BNF: PHP 8.3 - Drupal Core 10.5 - Contrib modules updated DDFSAL-277

rasben:
All of this has already been reviewed and approved in these pull requests:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2629

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2634

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2636

However, I needed a way to test it all together, and having the pull requests build ontop of each other was getting harder and harder to keep updated and manage.


.lagoon.yml
```diff
@@ -178,6 +178,15 @@ environments:
           - "go.cms-playground.dpl-cms.dplplat01.dpl.reload.dk":
               annotations:
                 nginx.ingress.kubernetes.io/proxy-buffer-size: 32k
+  pr-2679:
+    routes:
+      - varnish:
+          - "core-update.dpl-cms.dplplat01.dpl.reload.dk":
```
rasben, line 181: "core-update" is a chosen name for this PR - it can be any name, but it **cannot** be pr-XXXX as lagoon will fail.

PR #2697, opened 2025-08-19
Fix `task ci:cypress:local`

kasperbirch1:
#### Description
Fixes issue where Cypress could not run locally due to errors downloading dependencies.

This pull request makes a small but important configuration update to the `docker-compose.ci.yml` file. The change sets the `NO_PROXY` environment variable to ensure that npm and yarn registry traffic bypasses any configured proxy, which helps avoid tunneling issues during CI builds.

* Configuration update:
  * Added `NO_PROXY` environment variable to the `services` section to exclude npm and yarn registries from proxying.

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -10,6 +10,8 @@ services:
     environment:
       CYPRESS_INSTALL_BINARY: 0
       YARN_CACHE_FOLDER: '/app/.cache/yarn'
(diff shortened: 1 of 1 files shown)


kasperg:
I have not been able to recreate this project and I do not think this is the correct approach to address this either.

We discussed this here https://reload.zulipchat.com/#narrow/channel/240325-DDF/topic/Yarn.2FNPM.20install.20fejl.20i.20GitHub.20Actions/near/536430029.

PR #2719, opened 2025-08-27
DDFSAL-279 - Fix facet filters lost on infinite scroll

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-279

#### Description
- Add custom patch for Facets 3.0.0 to preserve filter parameters in AJAX requests
- Fixes issue where clicking 'Vis flere' (Show More) loses active facet filters
- Patch modifies QueryString::initializeActiveFilters() to merge query parameters into request data for AJAX calls from views_infinite_scroll module
- Maintains backward compatibility and only affects AJAX requests


kasperg (approved):
:+1: I think this is good enough to be sent for testing.

I would like to talk to you more about how you ended up with this solution before we determine how/if this can be a contribution.

I would also appreciate a test that shows that this works :smile:. Perhaps this could be a part of the current Facets test at https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/1d1ea368f365a0b07580b1aafd72bec08e8d9364/cypress/e2e/facets.cy.ts ?

PR #2732, opened 2025-09-02
DDFSAL-373 - Add validation for empty CQL search in material grid automatic paragraph

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-359
https://reload.atlassian.net/browse/DDFSAL-373

#### Test
https://varnish.pr-2732.dpl-cms.dplplat01.dpl.reload.dk/

#### Dev links
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-2732

#### Description
This pull request makes a small but important change to the `CqlSearchWidget` field widget by ensuring that the CQL search filter is required in the form.

* The `cql` textarea field in `CqlSearchWidget.php` is now marked as required, which enforces user input for this field in the form.

Fixes DDFSAL-359

### DIFF
--- a/web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/CqlSearchWidget.php
+++ b/web/modules/custom/dpl_fbi/src/Plugin/Field/FieldWidget/CqlSearchWidget.php
@@ -169,6 +169,7 @@ class CqlSearchWidget extends WidgetBase {
       '#title' => $this->t('Search filters', [], ['context' => 'dpl_fbi']),
       'cql' => [
         '#type' => 'textarea',
+        '#required' => TRUE,
         '#rows' => 3,
         '#title' => $fieldStorageDefinition->getPropertyDefinition('value')?->getLabel(),
         '#default_value' => $items[$delta]->value ?? '',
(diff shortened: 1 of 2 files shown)


xendk (approved):
:+1: 

But dude, you really need to work more on the AI summaries:

> This pull request makes a small but important change to the CqlSearchWidget field widget by ensuring that the CQL search filter is required in the form.
> 
> The cql textarea field in CqlSearchWidget.php is now marked as required, which enforces user input for this field in the form.

Two sentences that basically says the same thing. The first one uses *nine words* to tell me that this is a pull request with a change (you would think that's pretty obvious). The second one tells me it's that the field is now required, and then proceeds to tell me what `#required` *does*.

22 words to describe `'#required' => TRUE,`.

Compare:

"Make CQL string required in CqlSearchWidget"

One could suggest improving that with the *why*, but I'd think it's pretty obvious that the CqlSearchWidget requires a CQL string to work, so I think that would be overkill.

Oh, and you might wanna consider squashing the history when one commit completely reverts the other.

PR #2747, opened 2025-09-08
Flip logic for which paragraphs are allowed on pages and articles.

rasben:
This allows webmaster modules to create their own paragraphs for these pages, without having to touch the config export. We want to avoid webmaster modules affecting config-export when possible, to avoid them losing access to core updates.

### DIFF
--- a/config/sync/field.field.node.article.field_paragraphs.yml
+++ b/config/sync/field.field.node.article.field_paragraphs.yml
@@ -5,33 +5,18 @@ dependencies:
   config:
     - field.storage.node.field_paragraphs
     - node.type.article
-    - paragraphs.paragraphs_type.accordion
-    - paragraphs.paragraphs_type.banner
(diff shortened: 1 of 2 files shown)


config/sync/field.field.node.article.field_paragraphs.yml
```diff
@@ -48,150 +33,135 @@ settings:
...
+      go_linkbox: go_linkbox
+      go_material_slider_automatic: go_material_slider_automatic
+      go_material_slider_manual: go_material_slider_manual
+      go_video: go_video
+      go_video_bundle_automatic: go_video_bundle_automatic
+      go_link: go_link
+      go_video_bundle_manual: go_video_bundle_manual
+      go_images: go_images
+      go_text_body: go_text_body
+    negate: 1
```
rasben, line 48: This is the magic key - the paragraphs that are chosen will be **excluded** rather than **included**.

PR #2810, opened 2025-10-07
Ddfsal 415 login limbo

xendk:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-415

#### Description

Log out users when their adgangsplatformen token has expired.

We have reports of users that ends up in a state where the React frontend errors left and right because it gets error responses from most API calls due to an expired token. They have problems getting out of this state as, as far as Drupal is concerned, they're still logged in.

This PR adds an EventSubscriber that checks the expiry of the token and simply logs them out when it happens. Not the most user-friendly of solutions, but it should hopefully limit the amount of support requests on the subject. A better, but also more involved solution would have make the React frontend able to notice the situation and ask for a fresh token, but that's much more involved.

Also adds a Drush command for making Drupal think the token of a particular user is expired, for testing.

Additionally, this cleans up a bit in `dpl_login` and adds a `User` service which just wraps the `user_logout` function from `user.module`. This decouples `OpenIDConnectRedirectController`, `DplPatronRegController` and the new `LogoutExpiredTokensSubscriber` from the user module and makes them a bit easier to test.

### DIFF
--- a/web/modules/custom/dpl_login/src/User.php
+++ b/web/modules/custom/dpl_login/src/User.php
@@ -7,7 +7,7 @@ namespace Drupal\dpl_login;
 /**
  * Service to provide access to user.module functions.
  *
- * Function calls is no fun i tests, so we use this service so other services
+ * Function calls is no fun in tests, so we use this service so other services
  * can just depend on it, and it's easily mockable in tests.
  */
 class User {


kasperbirch1 (approved):
ice work here! 👍 Let’s get it out for testing.

PR #3127, opened 2026-02-23
[Shortcut] Add branch graphql query

spaceo:
This PR adds the [branch graphql query](https://github.com/danskernesdigitalebibliotek/dpl-cms/compare/develop...feature%2Fadd-branch-graphql-query?body=&expand=1#diff-7fc6e9dd1e68c30ecdd4b8ee4e02c78d744f8b0884237b5e738abfdf639f0a07R1-R23): `getBranches`

The query:
- shows per default all the branches tied to the current agency.
- exposes the isil number, address and available context of every branch
- makes it possible to filter by: `isilId`, `availabilityContexts`, `cmsConfigured`*

`*` cmsConfigured means simply if there is a Drupal branch node representation of the branch


web/modules/custom/dpl_library_agency/src/Plugin/GraphQL/DataProducer/GetBranchesProducer.php
```diff
@@ -0,0 +1,133 @@
...
+  public function resolve(
+    ?string $isilId,
+    ?array $whitelistTypes,
+    mixed $cmsConfigured,
+    FieldContext $field_context,
+  ): array {
+    // @todo Create strategy for queries that has a TTL like this one has.
+    $six_hours = 6 * 60 * 60;
+    $field_context->addCacheableDependency((new CacheableMetadata())->setCacheMaxAge($six_hours));
+    $field_context->addCacheableDependency($this->branchSettings);
```
kasperg: Please reconsider this approach.

As I see it this the cache lifetime here is related to the [caching of the branches within the repository](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/1cfd5929ff6a07706bfa125964effe1d01a2a789/web/modules/custom/dpl_library_agency/dpl_library_agency.services.yml#L38-L38).

I see a couple of options:

- Simple/temporary: Either ensure that the cache lifetime is the same or at least that it has a comment that explains why 6 hours is a good idea.
- More complex: Make repositories and branches implement `CacheableDependencyInterface` and pass this to the field context. This will make it possible to pass cache information (both time and tags) to the field context.

web/modules/custom/dpl_library_agency/src/Plugin/GraphQL/DataProducer/GetBranchesProducer.php
```diff
@@ -0,0 +1,133 @@
...
+          $branches,
+          fn(Branch $branch) => !in_array($branch->id, $excludedBranchIds, TRUE)
+        );
+      }
+    }
+
+    if ($cmsConfigured !== NULL) {
+      $branches = array_filter(
+        $branches,
+        fn(Branch $branch) => (bool) $branch->node === (bool) $cmsConfigured
```
kasperg:
```suggestion
        fn(Branch $branch) => (bool) $branch->getNode() === (bool) $cmsConfigured
```

Even though `$branch->node` is now public I still think it is preferable to use the getter.

kasperg (changes requested):
:+1: Nice. I have some considerations for you regarding how this is structured.

My key reason for requesting changes is the todo regarding caching. I would like to see this addressed in some way.

I see the commit mentioning telling clients how long they can cache responses. I do not know if this is the case or how I feel about this.

kasperg (changes requested):
:+1: I agree with most of the

web/modules/custom/dpl_library_agency/src/Plugin/Field/FieldFormatter/AddressDawaFormatter.php
```diff
@@ -27,7 +27,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
 
     foreach ($items as $delta => $item) {
       /** @var \Drupal\address_dawa\AddressDawaItemInterface $item */
-      $elements[$delta] = $this->buildOutput($item);
+      $elements[$delta] = static::buildOutput($item);
```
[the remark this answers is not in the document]
spaceo: The whole formatter has apparently been removed fro develop! So I guess it doesn't matter anymore. I wonder what has replaced it 🤔

web/modules/custom/dpl_library_agency/src/Branch/EmptyBranchRepository.php
```diff
@@ -2,16 +2,28 @@
...
    */
   public function getBranches(): array {
     return [];
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheMaxAge(): int {
+    return Cache::PERMANENT;
```
kasperg: @spaceo What have I gotten ourselves into here? 😅

I am considering this change in regard to whether this should be permanent - or no caching at all.

Currently this implementation is only used in combination with the `FallbackBranchRepository` where it is wired as the secondary repository. `FallbackBranchRepository` will pick the secondary repository if the first one fails (throws an exception). `FallbackBranchRepository` will also return the max age for the used repository. This will lead to the max age for the response being permanent if branch retrieval fails. Not good :smile:.

I see two options:

1. Setting no cache age for `EmptyBranchRepository`
2. Changing the implementation of `FallbackBranchRepository` to use a combination of values from the primary and secondary repositories instead of relying on what was used the last time and dropping `usedRepository` entirely. E.g.
  - The minimum max age of the two. This would be from `CachableBranchRepository` if `EmptyBranchRepository` stays permanent.
  - A merge of the cache tags from the two
  - A merge of the cache contexts from the two

web/modules/custom/dpl_library_agency/src/Branch/IdBranchRepository.php
```diff
@@ -29,4 +33,11 @@ public function getBranches(): array {
     }, $this->ids);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheMaxAge(): int {
+    return 0;
```
kasperg, line 40: :+1: Agree.

web/modules/custom/dpl_library_agency/src/Branch/FbsBranchRepository.php
```diff
@@ -46,4 +49,13 @@ public function getBranches(): array {
     return $branches;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheMaxAge(): int {
+    // The FBS repository itself has no caching — it always fetches live data.
+    // Caching is handled by the wrapping CacheableBranchRepository.
+    return 0;
```
kasperg, line 58: :+1: Agree.

kasperg (changes requested):

web/modules/custom/dpl_library_agency/src/Branch/EmptyBranchRepository.php
spaceo: I went for the swift solution of bypassing caching on the `EmptyBranchRepository`

spaceo (commented):
I apparently have to submit a review before I can comment your comments! - odd I would say. But if I dont my comments have the status: "pending" and are only visiible by me...

kasperg (approved):
:+1: LGTM

PR #3131, opened 2026-02-27
Event - show address fields by default. DDF-95

rasben:
Update the event editing flow, so the address fields are always shown when relevant - but don't force the user to fill it out. We set the country to Denmark by default.


rasben:
Migrated to https://github.com/danskernesdigitalebibliotek/dpl-web/pull/250

PR #3151, opened 2026-03-03
Event API: Expose original image + tweak scaled. DDF-21

rasben:
- Updates the current "image" property, to return a 1920x?? size image. This matches the documentation, that mentions that it is what is expected.
I'm pretty sure it is a mistake that it was sent along as a locked aspect ratio until now.

- Exposes the original image as a specific property - we expose it as a seperate property, as we would prefer third parties to use the scaled version instead, as the original images can be *very* large.


kasperbirch1 (approved):
LGTM 👍

rasben:
Closing; Migrated to https://github.com/danskernesdigitalebibliotek/dpl-web/pull/251

